PR #12, opened 2018-09-19
Update sizing-general.adoc

eric-schneider:


eric-schneider:
Used for Connect video

PR #16, opened 2018-10-08
DOC-4098

simon-dew:
Derivative works under Apache License must carry a notice


RichardSmedley (commented):
As Apache is the default license for Couchbase, do we need to add it here?

RichardSmedley (approved):
To answer my own question, if it's a third party component used with CB, then yes, it needs listing here.
Moving forward, collecting license info needs some more PM input - there needs to be a comprehensive list somewhere!

PR #18, opened 2018-10-12
DOC-4178

simon-dew:
Add N1QL for Analytics changes to Analytics what's new


RichardSmedley (approved):
I think it would be better to stick to the one-sentence per line standard, but the content LGTM :+1:

jamesnocentini:
@simon-dew it might be your editor's settings that are disabling text wrap?
Only long lines are not wrapped in this change set (example https://github.com/couchbase/docs-server/pull/18/files#diff-3f873f7861077cf5c9a75468971907cfR22). The rest looks good.

PR #20, opened 2018-10-16
Change Options position

jare:
Because all the examples the ordes is: cbrestore [backup-dir] [destination] [options]


simon-dew:
Thank you for your suggestion — it's always great to receive contributions!

This change would make the `cbrestore` command synopsis inconsistent with other commands, and the current instructions work.

PR #49, opened 2018-10-30
Added 6.0 feature upgrade matrix

eric-schneider:


modules/install/pages/upgrade-strategy-for-features.adoc
```diff
@@ -0,0 +1,71 @@
...
+| None
+
+| Scorch indexing system (Full Text Search)
+| Cluster
+| Offline/Online
+| Direct
+| Yes*
+| *FTS nodes running 6.0 use the Scorch indexing system by default for new indexes.
+You can continue to use previous indexes in 6.0, but they will not use Scorch.
+| To upgrade previous indexes to use Scorch, simply recreate them using the new index type option.
```
amarantha-k: Add a link to the FTS create index page?

modules/install/pages/upgrade-strategy-for-features.adoc
```diff
@@ -0,0 +1,71 @@
...
+|===
+
+[#feature-mixed-mode]
+† _Mixed Mode in this context refers to only the nodes that are running the indicated service. So if 'No' is marked in the Mixed Mode column of a given feature, then all nodes that are running that service must be upgraded to the same version in order to enable that feature._
+
+[#upgrade-ipv6]
+== Migrating Data from an IPv4 Cluster to an IPv6 Cluster
```
amarantha-k, line 43: Version number update from 5.5 to 6.0?

amarantha-k (changes requested):
Added inline comments

modules/install/pages/upgrade-strategy-for-features.adoc
eric-schneider: Done

modules/install/pages/upgrade-strategy-for-features.adoc
eric-schneider: I assume this is regarding the 5.5 references for IPv6. It's not specific to 6.0, but right now there isn't a good home for this information, so I left it in the same place it was for 5.5. Also there is a good chance these procedures will change soon.

amarantha-k (approved):

PR #50, opened 2018-10-31
Minor update to MB-29724 description

amarantha-k:


modules/release-notes/pages/relnotes.adoc
```diff
@@ -89,15 +89,15 @@ See xref:install:install-platforms.adoc[Supported Platforms] for the complete li
 | Issue | Description
 
 | https://issues.couchbase.com/browse/MB-30637[MB-30637^]
-| *Summary: Backup and restore for the Analytics service artifacts using the cbbackupmgr utility is currently not supported. 
+| *Summary: Backup and restore for the Analytics service artifacts using the `cbbackupmgr` utility is currently not supported.
```
eric-schneider, line 92: I feel like this will get a lot of questions if we don't provide any suggestions for backup best practices.

amarantha-k: Yes, we should follow-up with the Analytics team on this one.

eric-schneider (approved):

PR #72, opened 2018-11-06
Updating to match 5.5's latest links

RichardSmedley:


simon-dew (approved):
Verified all links to the latest SDKs, as for 5.5

PR #74, opened 2018-11-09
DOC-4142: Instructions to initialize Analytics service

simon-dew:
Clarify that when using couchbase-cli to initialize a cluster, you need to use `node-init` first to set the node options, then use `cluster-init` to set the cluster options.

### DIFF
--- a/modules/install/pages/init-setup.adoc
+++ b/modules/install/pages/init-setup.adoc
@@ -107,10 +107,10 @@ For production environments, it is recommended that database files and indexes _
 You can enter more than one location to store Analytics data.
 Click btn:[+] to specify an additional location for Analytics data, or click btn:[-] to remove a location.
 
-* [.ui]*Java Runtime Path*: If desired, enter the location for an alternate Java Runtime Environment (JRE) on the current node that you want to use for the Analytics Service.
+* [.ui]*Java Runtime Path*: If desired, enter the location for an alternative Java Runtime Environment (JRE) on the current node that you want to use for the Analytics Service.
 +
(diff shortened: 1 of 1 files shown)


modules/install/pages/init-setup.adoc
```diff
@@ -250,26 +250,65 @@ Once the [.ui]*Dashboard* screen appears, the new node has been successfully add
...
 Rather than using the Couchbase Web Console, you may elect to initialize your Couchbase cluster by means of the Couchbase Command Line Interface (CLI).
 
-The following CLI syntax can be used for the initial set-up of a single-node Couchbase Server cluster.
-It allows the establishing of administrative credentials, and of port number.
-It adds all services; sets separate RAM quotas for Data, Index, Search, Eventing, and Analytics services, and sets the Index Storage Setting (the default being to support memory-optimized global indexes):
+First you must use the xref:cli:cbcli/couchbase-cli-node-init.adoc[node-init] command to initialize the first node in the cluster.
+The following CLI syntax sets the locations where data files for each service will be stored on the node, sets the location of an alternative JRE on the node, and specifies the hostname that the cluster will use for this node.
 
-[source,bash]
+[source,bash,subs=+quotes]
```
jamesnocentini, line 256: Just curious, what does the `subs=+quotes` attribute do?

jamesnocentini (approved):
Nice use of code callouts!

modules/install/pages/init-setup.adoc
simon-dew: It enables formatting within the code block (in this case, _ _ for replaceable text)

PR #77, opened 2018-11-12
Fix Missing H1

amarantha-k:


amarantha-k:
@tonyjhillman Fyi - this is to fix an error reported by our SEO crawl.

tonyjhillman (approved):
ok

PR #79, opened 2018-11-13
DOC-4364: Add information about supported JRE configurations for Analytics

simon-dew:
* New page in the Install & Upgrade / System Requirements section, listing supported JRE implementation and version for Analytics service.

* New entry for this page in navigation ToC.

This page is intended as a location to list other environmental / platform requirements when necessary.


jamesnocentini:
I usually prefer to call out requirements when a reader is about to install the said component. People usually don’t go out of their way (at least I don’t) to check system requirements unless it’s clearly stated on the page that relates to the action they wish to perform (here, installation). What do you think about moving that content as a pre-requisites section on the installation page?

simon-dew:
Not sure if I understand @jamiltz — I agree that people won't go out of their way to check the system requirements. But there is no separate installation procedure for Analytics now, it's a service like Query or Indexing, and is installed when you install Couchbase Server.

As such, I think the requirements for Analytics should be listed in the Install & Upgrade section, along with the rest of the system requirements. I agree I need a link from the System Requirements page.

simon-dew:
Don't merge just yet, I'll make that change.

jamesnocentini:
Oh I see it’s a requirement for analytics only. Yes a link to that new page from the analytics intro page or server installation page would be good.

jamesnocentini:
Great! Other feedback from a first time reader: it would be nice to have a link to https://docs.couchbase.com/server/6.0/install/install-intro.html in the intro paragraph of 
https://docs.couchbase.com/server/6.0/analytics/run-query.html and 
https://docs.couchbase.com/server/6.0/analytics/primer-beer.html. That way readers know where to go to get an instance running before they continue reading those pages.

eric-schneider (approved):

simon-dew:
Thanks for the feedback — I've realised this page needs one important piece of information for clarity.

PR #107, opened 2018-11-15
DOC-4411

RichardSmedley:
Link fix 4.5

### DIFF
--- a/modules/developer-guide/pages/sub-doc-api.adoc
+++ b/modules/developer-guide/pages/sub-doc-api.adoc
@@ -222,7 +222,7 @@ When using key-value APIs, updates to a single field requires CAS to maintain co
 In case of highly contended documents, if a CAS mismatch occurs the operation needs to be restarted even though the modified field remains the same.
 Sub-document APIs do not require the use of CAS when updating single fields.
 However, you can still use the CAS protection for the document if your application requires it.
-For more information on CAS, see xref:2.3@java-sdk::common/concurrent-mutations-cluster.adoc[Concurrent Document Mutations].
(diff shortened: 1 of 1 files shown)


modules/developer-guide/pages/sub-doc-api.adoc
```diff
@@ -222,7 +222,7 @@ When using key-value APIs, updates to a single field requires CAS to maintain co
 In case of highly contended documents, if a CAS mismatch occurs the operation needs to be restarted even though the modified field remains the same.
 Sub-document APIs do not require the use of CAS when updating single fields.
 However, you can still use the CAS protection for the document if your application requires it.
-For more information on CAS, see xref:2.3@java-sdk::common/concurrent-mutations-cluster.adoc[Concurrent Document Mutations].
+For more information on CAS, see xref:2.3@java-sdk::concurrent-mutations-cluster.adoc[Concurrent Document Mutations].
```
eric-schneider, line 225: I think it should just be one colon: and not two:: 
https://docs.couchbase.com/home/contribute/cross-references.html#component-coordinate

eric-schneider (changes requested):

modules/developer-guide/pages/sub-doc-api.adoc
RichardSmedley: >  If a component name is present, and the name of the module is ROOT, the module name position can be left empty in some cases. That’s why you see a component name followed by two colons (::) sometimes.

https://docs.antora.org/antora/1.1/asciidoc/page-to-page-xref/ 

(I tested it with two, and have used them whenever I link component::page.adoc

eric-schneider (approved):

PR #126, opened 2018-11-19
DOC-4239: Modified partitioning paragraph

tonyjhillman:


deepkaran (changes requested):
In the first line, please change "increases aggregate query performance" to "increase query performance". Rest looks good.

tonyjhillman:
I think it has to be 'Index Partitioning increases', rather than 'increase', since 'Partitioning' is singular. I could change it to 'Partitioned indexes increase', if that is better.

deepkaran:
Sorry that was a typo. I wanted to say "Index Partitioning increases aggregate query performance" -> "Index Partitioning increases query performance"

tonyjhillman:
Modified and ready to merge for 6.0.

deepkaran (approved):

amarantha-k (approved):

PR #174, opened 2018-12-04
adding filterExpression to example

mgroves:
It doesn't seem like filterExpression is explicitly documented, but it does work, so I added it to the example. (See this discussion on StackOverflow: https://stackoverflow.com/questions/53591638/rest-api-for-enabling-advanced-filtering-in-couchbase-xdcr/53615931)


tonyjhillman (approved):
ok

PR #218, opened 2018-12-12
DOC-4253: max-parallelism parameter example

simon-dew:


modules/settings/pages/query-settings.adoc
```diff
@@ -567,19 +585,20 @@ These require the values to be given in `batch_named_args`, which contains a lis
 
 NOTE: The keys in each object need to match the named parameters in the `statement`.
 
-| `client_context_id`
+a| [[client_context_id]]
+`client_context_id`
 | string
 |
 2+| [Optional] A piece of data supplied by the client that is echoed in the response, if present.
```
jamesnocentini:
```suggestion
2+a| [Optional] A piece of data supplied by the client that is echoed in the response, if present.
```

jamesnocentini: Before the above suggestion.

image

After

image

Not sure why the `[Note]` is not visible.

modules/settings/pages/query-settings.adoc
```diff
@@ -631,44 +655,47 @@ cbq> \set -creds travel-sample user:pword, beer-sample user:pword;
...
 2+a|
 [Optional] For later, multiple executions, a query can be prepared, which results in five properties, of which one is called encoded_plan.
 This can then be used to execute the query.
 
 *Example*: Prepare the query result of the most expensive hotel.
-
- $ curl -v http://localhost:8093/query/service \
- -d 'statement=PREPARE pricy_hotel FROM SELECT name, max(price) FROM `travel-sample` WHERE type="hotel";
+[source,console]
+$ curl -v http://localhost:8093/query/service  -d 'statement=PREPARE pricy_hotel FROM SELECT name, max(price) FROM `travel-sample` WHERE type="hotel";
```
jamesnocentini:
```suggestion
$ curl -v http://localhost:8093/query/service  -d 'statement=PREPARE pricy_hotel FROM SELECT name, max(price) FROM `travel-sample` WHERE type="hotel";'
```

jamesnocentini (changes requested):

modules/settings/pages/query-settings.adoc
simon-dew: Good catch. I think the NOTE isn't shown because the admonition block markup is wrong. I'll add that as a new edit.

jamesnocentini (approved):

PR #222, opened 2018-12-14
DOC-4570: Removed last clause from html filter description

tonyjhillman:


tonyjhillman:
If this seems okay I'll make the change to the other versions, as appropriate.

amarantha-k (approved):

PR #223, opened 2018-12-14
DOC-4572: Modified wording of xdcr intro

tonyjhillman:


RichardSmedley (approved):
Removal looks fine to me - understanding that more work is being done on the topic. :+1:

PR #233, opened 2019-01-11
DOC-4464: Document Common Table Expression

simon-dew:
UPDATE: Draft build of the documentation has been removed. Please see updated documentation here:

* [WITH clause](https://docs-staging.couchbase.com/server/6.5/n1ql/n1ql-language-reference/with.html)
* [SELECT Syntax › WITH Clause](https://docs-staging.couchbase.com/server/6.5/n1ql/n1ql-language-reference/select-syntax.html#with-clause)
* [Identifiers › Aliases](https://docs-staging.couchbase.com/server/6.5/n1ql/n1ql-language-reference/identifiers.html#identifier-alias)


marcogrecopriolo:
I think we could improve on 

"The major issue with the LET clause is that the assignment comes after the FROM clause, and therefore can only be used in subselects, forcing the user to add an extra SELECT layer to be able to mimic the WITH functionality. The WITH syntax is more natural and more easily readable."

Although it is all fundamentally true, maybe rather than highlighting the problem, we could highlight the difference?

"The major difference with the LET clause is that the assignment comes before the SELECT clause vs after the FROM clause, allowing an earlier definition of expressions."

Or something of the sort...

sitaramv:
There is another syntax FROM... WHERE ... SELECT  ..., WITH also possible in that syntax before FROM

ANSI JOIN syntax is Missing in JOIN.  Are u pick up right one to update. Or you handling thise separately.

Alias ... Instead of Any where you should say with in the scope.
             In given query block.  WHERE can use aliases defined in parent query, WITH, FROM, LET, But not from GROUPING, LETTING and SELECT projection. Check scope in https://docs.couchbase.com/server/6.0/n1ql/n1ql-language-reference/subqueries.html

sitaramv:
Addition to marco comments.
WITH is evaluated once per query block
LET evaluated every object produced by source ( FROM/JOIN clause)

simon-dew:
Hi @sitaramv 

> There is another syntax FROM... WHERE ... SELECT ..., WITH also possible in that syntax before FROM

I believe this is covered by existing railroad diagrams.

> ANSI JOIN syntax is Missing in JOIN. Are u pick up right one to update. Or you handling thise separately.

ANSI JOIN will be covered by DOC-4462.

> Alias ... Instead of Any where you should say with in the scope.
> In given query block. WHERE can use aliases defined in parent query, WITH, FROM, LET, But not from GROUPING, LETTING and SELECT projection. Check scope in https://docs.couchbase.com/server/6.0/n1ql/n1ql-language-reference/subqueries.html

Thanks!

sitaramv:
I believe this is covered by existing railroad diagrams.

Hum. I am not sure in that case is the with-select place is right one. Let me talk to marco

marcogrecopriolo:
Checked the railroads - they are correct both select-from and from-select appear after the with clause

simon-dew:
> "The major difference with the LET clause is that the assignment comes before the SELECT clause vs after the FROM clause, allowing an earlier definition of expressions."
> 
> Or something of the sort...

Done

simon-dew:
> WITH is evaluated once per query block
> LET evaluated every object produced by source ( FROM/JOIN clause)

Done

simon-dew:
> Alias ... Instead of Any where you should say with in the scope.
> In given query block. WHERE can use aliases defined in parent query, WITH, FROM, LET, But not from GROUPING, LETTING and SELECT projection. Check scope in https://docs.couchbase.com/server/6.0/n1ql/n1ql-language-reference/subqueries.html

Done

PR #249, opened 2019-01-15
DOC-4610: Document CONCAT2()

simon-dew:
Please see draft documentation here:

* [String Functions › CONCAT2()](https://simon-dew.github.io/docs-site/DOC-4610/server/6.5/n1ql/n1ql-language-reference/stringfun.html#fn-str-concat2)
* [String Functions › CONCAT()](https://simon-dew.github.io/docs-site/DOC-4610/server/6.5/n1ql/n1ql-language-reference/stringfun.html#fn-str-concat)


sitaramv:
Add this and you are good

If any argument or array element is MISSING, Returns MISSING.
If any argument or array element is non-string, Returns NULL.

simon-dew:
> If any argument or array element is MISSING, Returns MISSING.
> If any argument or array element is non-string, Returns NULL.

Done

PR #307, opened 2019-01-29
DOC-4468: Document chained LET

simon-dew:
UPDATE: Draft documentation build removed. Please see updated changes [here](https://docs-staging.couchbase.com/server/6.5/n1ql/n1ql-language-reference/let.html).

(Added description and example of chained LET.)


amarantha-k (approved):

simon-dew:
@marcogrecopriolo

> Looks good – the only thing that I would say is that the same concept as chained LETs applies to chained WITHs, in short, if you have a WITH clause with multiple common table expressions, a subsequent CTE can reference a prior CTE, much like it happens with LETs.
We should mention that in the WITH clause too.

PR #330, opened 2019-02-07
Improve non-root steps

jimwwalker:
I followed the non-root instructions and found them to be a little confusing, my issue was mainly that it wasn't clear that when I perform the cpio (step 1) the directory I am in is the directory where the install will reside, reading the steps as they are lead me to think that I could cpio in the download directory and then direct the install to anywhere (the reloc.sh step)

It appeared that the reloc.sh script would move the extracted files to anywhere I want, but it doesn't, there seemed to be no choice other than "reloc.sh pwd", any other path and it failed.

So i've added what I think are some changes to help clarify that the download directory is the final installation directory as well.


eric-schneider (approved):
Thanks Jim!

Based on what you found, I think these instructions make things a lot more clear.

PR #336, opened 2019-02-13
DOC-4462: Document ANSI JOIN support for expression and subquery term

simon-dew:
UPDATE: Draft documentation has been removed. Please see updated documentation here:

* [JOIN clause › ANSI JOIN Right-Hand Side](https://docs-staging.couchbase.com/server/6.5/n1ql/n1ql-language-reference/join.html#ansi-join-rhs)
* [JOIN clause › ANSI JOIN clause › Examples](https://docs-staging.couchbase.com/server/6.5/n1ql/n1ql-language-reference/join.html#examples)
* [JOIN clause › ANSI JOIN Hints](https://docs-staging.couchbase.com/server/6.5/n1ql/n1ql-language-reference/join.html#ansi-join-hints)
* [SELECT Syntax › JOIN Clause › ANSI JOIN](https://docs-staging.couchbase.com/server/6.5/n1ql/n1ql-language-reference/select-syntax.html#ansi-join-clause)


simon-dew:
This PR builds on the refactoring done in [PR 331](https://github.com/couchbase/docs-server/pull/331)

simon-dew:
- [x] select-syntax.adoc, index-nest-clause syntax, `JOIN` should be `NEST`
- [x] select-syntax.adoc, join-clause syntax, missing `join-clause ::=`

miaobingjie:
- [x] AS alias - should mention that when right-hand side is a subquery or an generic expression, then alias is required
- [x] Limitations - the 3rd bullet is specific for when right-hand side is a keyspace, and nested-loop join is used. Can say something like: when right-hand side of ANSI JOIN is a keyspace reference, for nested-loop join consideration an appropriate secondary index must exist on the right-hand side keyspace; for hash join consideration a primary index can be used
- [x] USE HASH hint - this section refers to keyspace since previously we only support keyspace reference as right-hand side for ANSI JOIN. Now that subquery and expression terms are supported, we should reword to use right-hand side expression instead of keyspace

PR #360, opened 2019-03-01
DOC-4461: Document ANSI MERGE support

simon-dew:
UPDATE: Draft documentation has been removed. Please see completed documentation here:

- [MERGE](https://docs-staging.couchbase.com/server/6.5/n1ql/n1ql-language-reference/merge.html)


miaobingjie:
The information looks ok in general. My only comment is that there appears to be quite a bit of duplicated information between ANSI MERGE and Lookup MERGE. For example, merge-source and merge target are exactly the same between the two. merge-hint is different. Then in merge actions the update action and the delete action are also exactly the same, insert action is different. I think you did it this way on purpose. I don't know whether it's possible that in Lookup MERGE section refer back to ANSI MERGE section and say something like this is the same as in ANSI MERGE section, instead of presenting the same information again.

PR #361, opened 2019-03-04
DOC-4857: ANSI OUTER JOIN to ANSI INNER JOIN transformation

simon-dew:
UPDATE: Draft documentation has now been removed. Please see updated documentation here:

* [JOIN Clause › ANSI JOIN Clause › Syntax › Join Type](https://docs-staging.couchbase.com/server/6.5/n1ql/n1ql-language-reference/join.html#ansi-join-type)

The only change is the new NOTE section.

### DIFF
--- a/modules/n1ql/pages/n1ql-language-reference/join.adoc
+++ b/modules/n1ql/pages/n1ql-language-reference/join.adoc
@@ -118,6 +118,16 @@ WHERE r.sourceairport = "SFO"
 ----
 |===
 
+[NOTE]
+--
+In Couchbase Server 6.5 and later, if you create either of the following:
+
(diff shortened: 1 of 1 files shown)


miaobingjie:
It is somewhat confusing in that the WHERE clause in a query is not "associated" with any join defined in the query. A query can have multiple joins but only a single WHERE clause.
Also technically we consider filters in the WHERE clause as well as ON clauses of any subsequent INNER JOINs for purpose of determining whether a filter rejects null or missing result for right-hand side (of a left outer join). I would put it somewhat like:
a LEFT OUTER JOIN where all the NULL or MISSING results from the right-hand side can be filtered out by a predicate in the WHERE clause or the ON clause of a subsequent INNER JOIN.
similar for RIGHT OUTER JOIN.

PR #396, opened 2019-03-18
DOC-4454 Document Couchbase Server process list

eric-schneider:
Provide a list of processes that Couchbase runs. Currently Linux only.

- [Couchbase Server Processes (review site)](https://eric-schneider.github.io/DOC-4454/server/6.0/install/server-processes.html)


jamesnocentini (approved):

tonyjhillman (approved):
I think this is a very valuable page, and I shall link to it in various places. 

Possibly, the list of processes that only run on the node on which the service is deployed might itself be expressed as a column in the table - otherwise, it might be good to phrase the sentence you have to emphasise that the list is the full list (not just, "including"..if you know what I mean).

PR #412, opened 2019-03-22
DOC-4472: Document UNNEST scan improvements

simon-dew:
@RichardSmedley — we missed this inconsistency when applying edits to the Array Indexing document.


RichardSmedley (approved):
LGTM
Well spotted :+1:

PR #413, opened 2019-03-25
DOC-4958 Remove top-level tools page

eric-schneider:
Currently, the following page has no content:
https://docs.couchbase.com/server/6.0/tools/tools-ref.html

I looked back, and this originated around 4.1/4.5: https://github.com/couchbase/docs-cb4/commit/cd55abba0725f1449a0ce1f99b87f9844e9760f3

I haven't found any references in the docs to this topic, so I think we should delete it and turn the xref in the nav into a normal nav listing.


amarantha-k (changes requested):
Instead of removing the top-level page, could you convert it to a landing page that list the available tools, maybe grouping them by function such as query, backup-restore, monitoring, troubleshooting, etc

amarantha-k (approved):
Summarizing discussion over Slack: 
Plan to revisit this page for the suggested updates at a later point. 
Agree to removing the page (with no content) from the nav for now.

PR #416, opened 2019-03-26
DOC-4961: Remove Admin API Explorer until it is updated

simon-dew:
The Admin API Explorer is in beta and has not been updated since Couchbase Server 5.0. Removing it from the documentation until we've had a chance to update it.


RichardSmedley (approved):
LGTM :+1:

PR #419, opened 2019-03-26
DOC-4454 Correct Index process

eric-schneider:


amarantha-k (approved):
Index Service instead of Indexer Service?

eric-schneider:
Chaitra approved

PR #424, opened 2019-03-29
DOC-5002: Modified description of XDCR Compression Type in Advanced S…

tonyjhillman:
…ettings table


Chaitra19 (commented):
No, no. Can't say auto is best compression technology. It's just a configuration parameter. Don't publish this, we will go over it in a call

amarantha-k:
@Chaitra19 Could you please take a look at the updates and approve if no further changes are required? Thanks!

Chaitra19 (commented):
There is a change in this behavior, we do not do Snappy from 6.5. Let me confirm if we are backporting that to older versions as well. 
Don't wait on this for beta. But I'll confirm with team and revert.

amarantha-k:
@Chaitra19 Please update the PR if you have a confirmation.

PR #427, opened 2019-04-02
DOC-4475: additional statistics for N1QL auditing

simon-dew:
I needed to document the Admin REST API so that I can document the new statistics in Mad Hatter.

Please see draft build [here](https://simon-dew.github.io/docs-site/DOC-4475/server/6.0/n1ql/n1ql-rest-api/admin.html).

This new page was generated using Swagger2Markup.


simon-dew:
The YAML spec from which I generated the new Admin REST API page can be found in [cb-swagger #3](https://github.com/couchbaselabs/cb-swagger/pull/3)

simon-dew:
- [x] Needs to mention authentication: basic or none
- [x] Need to clarify consumes `application/x-www-form-urlencoded` by default, can accept `application/json` if `Content-Type` header specified
- [x] Correct `application/xml` for **/debug/vars** endpoint

RichardSmedley (approved):
LGTM
(afaict - don't have time to test out the REST interface today, but other copy all fine.)

jamesnocentini:
I checked out the branch and built it. The output looks very clean, everything linkable. Great job!!

jamesnocentini (approved):

jamesnocentini:
@simon-dew The swagger spec could (not necessarily on this PR) also be checked in the docs-server repo. Perhaps in the `assets/attachments` folder? (that's where it's kept for the sync-gateway component https://github.com/couchbase/docs-sync-gateway/tree/release/2.1/modules/ROOT/assets/attachments).

eric-schneider:
It's these types of pages where I wish we had H3s in the right-hand nav.

simon-dew:
> The swagger spec could (not necessarily on this PR) also be checked in the docs-server repo. Perhaps in the `assets/attachments` folder?

Good idea, especially as we have less control over what gets accepted into the cb-swagger repo.

simon-dew:
> Perhaps in the `assets/attachments` folder?

On second thoughts @jamiltz -- this would mean that the source YAML file would be included in the build (in the `_attachments` directory), and I don't know if we want to do that.

jamesnocentini:
It’s a good point. If we publish it on docs.couchbase.com users might think it’s officially supported. Swagger specs can be used for many things like generating http clients (swagger codegen)..and it would be harder to tell users not to do that (or that we can’t support issues with those clients) if the spec is on docs.couchbase.com. That’s the main drawback I see.

PR #432, opened 2019-04-04
DOC-4709: No Example of EXECUTE Statement

simon-dew:
Draft build documents available here:

* [EXECUTE](https://simon-dew.github.io/docs-site/DOC-4709/server/6.0/n1ql/n1ql-language-reference/execute.html)
* [PREPARE](https://simon-dew.github.io/docs-site/DOC-4709/server/6.0/n1ql/n1ql-language-reference/prepare.html)


modules/n1ql/pages/n1ql-language-reference/execute.adoc
```diff
@@ -1,23 +1,71 @@
 = EXECUTE
 :page-topic-type: concept
+:imagesdir: ../../assets/images
```
jamesnocentini, line 3: Is there any specific reason to add this document attribute? AFAIK, Antora automatically adds it when building.

jamesnocentini: I've tried locally, seems to work without the explicit declaration of `imagesdir`. Convenient feature to save keystrokes.

jamesnocentini (approved):

modules/n1ql/pages/n1ql-language-reference/execute.adoc
simon-dew: Good point. I only put these attributes in so that I can see the pictures when I preview the document in my text editor or [in GitHub](https://github.com/couchbase/docs-server/blob/0bb84dc877b06b86688b706213cb6f1d912af962/modules/n1ql/pages/n1ql-language-reference/execute.adoc). It's not so important in this doc, but it becomes more useful when there are a lot of images. Dan confirmed that the attribute is stripped out when Antora builds the web help, so it doesn't do any harm.

jamesnocentini: I had always assumed images could not display on github. Good to know it works that way.

simon-dew: I'm investigating the `_attributes.adoc` file, which looks like the "correct" way of handling common page attributes. It's mentioned in the [Antora documentation](https://docs.antora.org/antora/2.0/modules/#module-overview), but I can't find any other documentation about it.

PR #452, opened 2019-04-11
DOC-3842 maxconn 65,000

RichardSmedley:


simon-dew (commented):
Where does the figure 65,000 come from? I can't see that in the linked issue.

RichardSmedley:
@simon-dew I had to grep NS Server for it.
http://src.couchbase.org/source/xref/mad-hatter/ns_server/src/ns_config_default.erl#166

simon-dew (approved):

PR #460, opened 2019-04-16
DOC-5058: Admin API page gives AsciiDoctor WARNINGs at build time

simon-dew:
The auto-generated Admin API document contains the path parameters `{name}`, `{cluster}`, `{node}`, and `{request}`. These are interpreted by Antora as AsciiDoc attributes. Since none of these attributes is defined, Antora gave warning messages at build time.

I cannot change the format of the path parameters, as they are necessary for the validity of the Swagger spec from which the Admin API document is generated.

In this fix I have added dummy attribute definitions to the Paths section of the document. These define the AsciiDoc attributes `{name}`, `{cluster}`, `{node}`, and `{request}` as escaped versions of the corresponding path parameter names. The output is identical, but Antora no longer generates warnings at build time.

I have also updated to the latest snapshot of Swagger2Markup, which means the table definitions have also changed in the Admin API document.

### DIFF
--- a/modules/n1ql/pages/n1ql-rest-api/admin.adoc
+++ b/modules/n1ql/pages/n1ql-rest-api/admin.adoc
@@ -35,6 +35,14 @@ __Version__ : 6.0
 [[_paths]]
 == Paths
 
+:name: \{name}
+:cluster: \{cluster}
+:node: \{node}
+:request: \{request}
+
+// Dummy attributes for path parameters
+
+
 [[_admin_clusters_get]]
 === GET /admin/clusters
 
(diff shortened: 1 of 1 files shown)


simon-dew:
The source Swagger spec and the tools to generate this document are [here](https://github.com/simon-dew/swagger2markup-gradle-project-template/tree/release/6.0).

The dummy attribute definitions come from this file [here](https://github.com/simon-dew/swagger2markup-gradle-project-template/blob/release/6.0/src/docs/asciidoc/extensions/paths/document-begin-text.adoc).

RichardSmedley (approved):
Nice hack - builds for me without errors :+1:

simon-dew:
Hack is the right word 🙄

eric-schneider:
@simon-dew Why did all the column widths have `a` appended to them? Was this to allow AsciiDoc markup in those columns, like when you add `a|` to a row?

simon-dew:
> Why did all the column widths have `a` appended to them?

Yes. ([From the AsciiDoctor manual](https://asciidoctor.org/docs/user-manual/#cols-format).) Why they changed it I can't say...

PR #464, opened 2019-04-17
DOC-5058: Admin API page gives AsciiDoctor WARNINGs at build time

simon-dew:
Further to #460: found another one.


eric-schneider (approved):
If https://github.com/couchbase/docs-server/pull/460 works fine, then this looks good to me ;-)

PR #466, opened 2019-04-18
DOC-4907 update meta package number

eric-schneider:


eric-schneider:
This is just a simple update to the meta package version found in older versions of the docs to help avoid obvious errors due to platform incompatibility.
The actual reformulated and tested docs are here: https://docs.couchbase.com/server/6.0/install/ubuntu-debian-install.html

PR #472, opened 2019-04-19
Server 5.5.4 release notes draft for review

amarantha-k:


arod1987:
@evgenymakarenko reviewed query and signed off

arod1987:
@ajaybhullar will review GSI and give his comments

ajaybhullar:
| https://issues.couchbase.com/browse/MB-33664[MB-33664^]
| If you set the default num_replica using setting, it skipped validation during index creation when the node list is presented in the WITH clause.


this one should probably say

It was possible to create an index with less replicas than the default num_replicas if you used nodes_list in the WITH clause with less nodes than the default number of replicas

| https://issues.couchbase.com/browse/MB-32720[MB-32720^]
| Fixed an issue so that the index service does not get stuck during warmup.

If you could elaborate on this one, saying that the index service got stuck because of core dumps or something to that effect that would be great. 

These are the only two under index that I feel should be modified

ritamcouchbase:
Ritam reviewed the ns_server, kv_engine and security.

arod1987:
XDCR, tools and UI look good to me

arod1987 (changes requested):
| https://issues.couchbase.com/browse/MB-33664[MB-33664^]
| If you set the default num_replica using setting, it skipped validation during index creation when the node list is presented in the WITH clause.

this one should probably say

It was possible to create an index with less replicas than the default num_replicas if you used nodes_list in the WITH clause with less nodes than the default number of replicas

| https://issues.couchbase.com/browse/MB-32720[MB-32720^]
| Fixed an issue so that the index service does not get stuck during warmup.

If you could elaborate on this one, saying that the index service got stuck because of core dumps or something to that effect that would be great.

These are the only two under index that I feel should be modified

amarantha-k:
Thanks everyone! Feedback has been incorporated.

PR #478, opened 2019-04-26
DOC-4938 fix nav entry for cbrestore

eric-schneider:
From Dan: "the equals sign in the linked text (i.e., rehash=1) is interpreted as an attribute assignment (key: rehash, value: 1). Once the attribute is taken away, no text remains. When that happens, Antora falls back to using the page ID as the linked text."


simon-dew (approved):
LGTM. I haven't tested the build.

eric-schneider:
Works on local build.

PR #505, opened 2019-05-20
Reconcile mad-hatter branches

simon-dew:
Convince git that `couchbase:mad-hatter` and `simon-dew:mad-hatter` are the same.


jamesnocentini:
@simon-dew out of interest, how would this PR change anything? GitHub shows 0 lines changed.

PR #530, opened 2019-06-04
Doc 5297

tonyjhillman:


modules/manage/pages/manage-security/configure-ldap.adoc
```diff
@@ -4,12 +4,12 @@
...
-== Understanding LDAP Authentication.
+== Understanding LDAP Authentication and Authorization
 
 For authentication purposes, Couchbase Server can be configured to use LDAP.
-This allows the user, when they attempt to access Couchbase Server by presenting their credentials, to be authenticated by LDAP.
-Group information for the authenticated user is fetched from the LDAP server.
-Each group is mapped to a Couchbase-Server group, whose assigned roles are granted to the user.
+This allows users, when they attempt to access Couchbase Server by presenting their credentials, to be authenticated by LDAP.
+Optionally, LDAP _group_ information for the authenticated user can be fetched from the LDAP server.
+Each LDAP group can be mapped to a Couchbase-Server group, whose assigned roles are thereby granted to the user.
```
timofey-barmin: I suggest we add the following, since it's important feature (please rephrase if needed):
In this case if there is no extra roles needs to be assigned to the user directly there is no need to create the user in Couchbase Server. The user will authenticate on LDAP server "transparently" and get roles via "ldap groups to couchbase groups" mapping.

modules/manage/pages/manage-security/configure-ldap.adoc
```diff
@@ -4,12 +4,12 @@
...
 
 [#understanding-ldap-authentication]
-== Understanding LDAP Authentication.
+== Understanding LDAP Authentication and Authorization
 
 For authentication purposes, Couchbase Server can be configured to use LDAP.
-This allows the user, when they attempt to access Couchbase Server by presenting their credentials, to be authenticated by LDAP.
-Group information for the authenticated user is fetched from the LDAP server.
-Each group is mapped to a Couchbase-Server group, whose assigned roles are granted to the user.
+This allows users, when they attempt to access Couchbase Server by presenting their credentials, to be authenticated by LDAP.
```
timofey-barmin, line 10: I think we need to add 2 more things here:
1. Add more details on how authentication is done. First of username specified by the user is mapped to LDAP distinguished name (which can be done either by using a template or by performing an LDAP search). Then Couchbase-Server tries to authenticate the user using the DN from first step and password that is provided by the user; 
2. How users get roles after authentication. Something like this (please rephrase): After authentication users are granted with roles that were assigned to corresponding "external" users in Couchbase-Server.

modules/manage/pages/manage-security/configure-ldap.adoc
```diff
@@ -58,8 +58,8 @@ Note the options at the upper right:
 image::manage-security/ldapAndGroupsTabs.png[,200,align=left]
 
 The *LDAP* tab allows an LDAP server to be specified for external authentication.
-The *Groups* tab allows definition of _Couchbase-Server groups_, which are to be associated with _LDAP groups_, and which can be assigned Couchbase-Server _roles_.
-The *ADD USER* tab allows individual users to be added, and for their authentication to be specified as *external*; allowing their LDAP group-membership to be mapped to one or more Couchbase-Server groups and their assigned roles.
+The *Groups* tab allows definition of _Couchbase-Server user-groups_, which are to be associated with _LDAP groups_, and which can be assigned Couchbase-Server _roles_.
+The *ADD USER* tab allows individual users to be added, and for their authentication to be specified as *external*; allowing their LDAP group-membership to be mapped to one or more Couchbase-Server user-groups and their assigned roles.
```
timofey-barmin: Phrase "allowing their LDAP group-membership to be mapped to one or more Couchbase-Server user-groups and their assigned roles." looks a bit incorrect to me.
"external" users don't allow LDAP group-membership to be mapped to anything. The only purpose of the "external" users is to assign roles to them. If we assign roles to "ldap" groups (by mapping them to CB groups) we might not need "external" users at all. We might want to have both, though, in sense that user can get some roles from ldap groups and some directly assigned to corresponding "external" user. Let me know if it sounds confusing.

modules/manage/pages/manage-security/configure-ldap.adoc
```diff
@@ -69,52 +69,58 @@ To configure an LDAP server with the UI, left-click on the *LDAP* tab, at the up
...
 The dialog presents two main panels.
 The left panel features options whereby LDAP can be configured.
 The right panel features help text, for *LDAP Host Configuration*, *User Authentication*, and *Group Authorization*.
 
 Initially, the left panel displays the following fields:
 
-* *LDAP Hosts*. A comma-separated list of host-names, each of which is an LDAP server on which external authentication is to occur.
+* *LDAP Host(s)*. A comma-separated list of host-names, each of which is an LDAP server on which external authentication can occur.
+The first accessible server in the list is the one that is used.
+For example, if the list is `server1, server2`, provided that `server1` is accessible, it is used.)
```
timofey-barmin: Typo: parentheses at the end is needless

modules/manage/pages/manage-security/configure-ldap.adoc
```diff
@@ -69,52 +69,58 @@ To configure an LDAP server with the UI, left-click on the *LDAP* tab, at the up
...
 [#certificate-text-field]
 image::manage-security/certificateTextField.png[,400,align=left]
 +
-The text of the appropriate certificate should be copied and pasted into the *Certificate Text* field.
+The text of the appropriate certificate should be copied and pasted, in PEM format, into the *Certificate Text* field.
 
 * *Contact LDAP host anonymously*.
-This checkbox, if checked, causes Couchbase Server to attempt to contact the LDAP host anonymously.
+Checking this checkbox causes Couchbase Server to attempt to contact the LDAP host anonymously.
 However, the attempt succeeds only if supported by the LDAP configuration.
```
timofey-barmin: I would say "if supported by the LDAP server" to not confuse it with the LDAP configuration that is described in this document

modules/manage/pages/manage-security/configure-ldap.adoc
```diff
@@ -123,10 +129,10 @@ image::manage-security/configureLDAPhalfComplete.png[,520,align=left]
 
 
 Optionally, the *Check Network Settings* button can now be left-clicked on.
-The tests whether the specified LDAP hosts are accessible across the network.
-If one or more servers is not accessible, an error is displayed on the dialog.
+This tests whether the specified LDAP hosts are accessible across the network.
+If none of the servers is accessible, an error is displayed on the dialog.
 
-The dialog provides the following additional fields:
+The dialog provides the following additional controls, for editing and deleting the group:
```
timofey-barmin: Not sure what you mean. We are not editing and deleting groups here.

modules/manage/pages/manage-security/configure-ldap.adoc
```diff
@@ -135,33 +141,107 @@ This expands the dialog vertically, as follows:
 [#configure-ldap-dialog-enable-ldap-user-auth-field]
 image::manage-security/configureLdapEnableLdapUserAuthField.png[,400,align=left]
 +
-This provides three options whereby usernames can be mapped on the LDAP server.
+This provides three options whereby usernames can be mapped to LDAP distinguished names (https://ldap.com/ldap-dns-and-rdns/[DN^]).
 The default option is *Template*.
-An appropriate template should be entered into the *Template* editable text field.
+An appropriate mapping should be entered into the *Template* editable text field: this avoids the need to request the LDAP server to provide the username-to-DN mapping.
+The required format is indicated by the placeholder text, within the field.
```
timofey-barmin: Format is described in RFC 4514 (https://tools.ietf.org/html/rfc4514).

modules/manage/pages/manage-security/configure-ldap.adoc
```diff
@@ -135,33 +141,107 @@ This expands the dialog vertically, as follows:
...
+The `%u` is replaced by the username used at login.
++
+For example, based on the placeholder, if the user logs in as `exampleUser`, the user is mapped to the DN `cn=exampleUser,dc=example,dc=com`.
```
timofey-barmin: in placeholder the different template is shown. Correct example: if template is "cn=%u,dc=example,dc=com" and user logs in as `exampleUser`, the user is mapped to the DN `cn=exampleUser,dc=example,dc=com`

modules/manage/pages/manage-security/configure-ldap.adoc
```diff
@@ -135,33 +141,107 @@ This expands the dialog vertically, as follows:
...
++
+Each of the other options, *LDAP query* and *Custom*, likewise provides a vertical expansion of the dialog, to accommodate entry of an appropriate mapping-procedure.
+Selecting *LDAP query* displays the following:
++
+image::manage-security/ldapQuery.png[,400,align=left]
++
+Couchbase server performs the specified LDAP https://ldap.com/the-ldap-search-operation/[search operation] to get the user's DN from LDAP server.
+*Base* is the _search base_ https://ldap.com/the-ldap-search-operation/[DN].
+*Filter* is the https://ldap.com/the-ldap-search-operation/[search filter].
+Search scope is always `one`; and `%u` is replaced with the username that was used by the current administrator to log into Couchbase Web Console.
```
timofey-barmin:
```suggestion
Search scope is always `one`; and `%u` is replaced by the username used at login.
```

modules/manage/pages/manage-security/configure-ldap.adoc
```diff
@@ -135,33 +141,107 @@ This expands the dialog vertically, as follows:
...
++
+This option allows a custom rule-set to be provided.
+The rules are used for mapping usernames to LDAP DNs, and must be expressed using the following JSON format:
++
+image::manage-security/jsonSearchSyntax.png[,400,align=left]
++
+These rules, expressed sequentially, are applied one by one, until one of the regular expressions matches.
+When a match is achieved, the specified query or template is used to retrieve the user's DN.
++
+If no match is found, the assumes the username employed by the administrator at Couchbase-Server login to be the DN (in other words, the administrator logged in with a username such as `cn=exampleUser,dc=example,dc=com`.
```
timofey-barmin:
```suggestion
If no match is found, the Couchbase Server assumes the username employed by the administrator at Couchbase-Server login to be the DN (in other words, the administrator logged in with a username such as `cn=exampleUser,dc=example,dc=com`.
```

It sounds to me like you are talking about the administrator who is configuring LDAP here. If so, it's not correct. Who actually we are talking about here are the future users that will use ldap to log in the system. Their usernames will be treated as DN's. Please let me know if it doesn't make any sense.

modules/manage/pages/manage-security/configure-ldap.adoc
```diff
@@ -135,33 +141,107 @@ This expands the dialog vertically, as follows:
...
+The LDAP groups of which a user is a member can be searched for by means of either the *User's attributes* or an *LDAP Query*, each of which is provided as a radio-button option.
+Selection of each reveals a corresponding set of fields, in which information can be added:
+
+** *User's attributes*.
+Selected by default, this instructs Couchbase Server to assume that each LDAP user-record contains an attribute featuring the list of groups of which the user is a member.
+Couchbase Server therefore performs the following LDAP search: `<userDN>?<attribute>?one`.
++
+The value of the specified `attribute` is treated as a list of groups.
+For example, if `attribute` is set to `memberOf`,
+Couchbase Server perform the following search for the specified user's groups:
```
timofey-barmin:
```suggestion
Couchbase Server performs the following search for the specified user's groups:
```

modules/manage/pages/manage-security/configure-ldap.adoc
```diff
@@ -207,10 +289,12 @@ image::manage-security/configureLDAPdialogSaveButton.png[,260,align=left]
 
 Alternatively, left-click on *Cancel* to abandon the configuration procedure.
 
-[#create-couchbase-server-group]
-=== Create a Couchbase-Server Group
+[#map-ldap-groups-to-couchbase-server-roles]
+=== Map LDAP Groups to Couchbase-Server Roles
 
-To create a Couchbase-Server _group_, and map it to an _LDAP group_, left-click on the *Groups* tab, at the upper right of the *Users* panel, on the *Security* screen:
+To map an LDAP group to a Couchbase-Server role, you must first create a Couchbase-Server user-group; then associate this user-group with an LDAP group; and finally, assign roles.
```
timofey-barmin:
```suggestion
To map an LDAP group to Couchbase-Server roles, you must first create a Couchbase-Server user-group; then associate this user-group with an LDAP group; and finally, assign roles.
```

modules/manage/pages/manage-security/configure-ldap.adoc
```diff
@@ -229,39 +313,78 @@ image::manage-security/addGroupTab.png[,90,align=left]
...
+image::manage-security/editGroupAdmins.png[,420,align=middle]
+
+Within this dialog, the description, mapping, or roles for the group can be edited.
+For details of selecting roles within the *Roles* panel, see xref:manage:manage-security/manage-users-and-roles.adoc[Manage Users and Roles].
+
+
+[#adding-an-externally-authenticated-user]
+=== Adding an Externally Authenticated User
+
+Couchbase-Server users can be specified as _externally authenticated_, when they are added to the system.
```
timofey-barmin, line 439: I would also add here that authentication does not require creation of "external" users and the whole point of doing that is to assign roles to ldap users. If this is not needed (users get roles via ldap groups), it's not required to create "external" users in Couchbase Server. 

There are three ways to assign roles to an ldap user:
1) Directly, in add/edit user dialog (admin needs to create "external" user);
2) Add manually to Couchbase user-group (admin needs to create "external" user and add it to user-group) (Tony, please note that ldap groups are not needed for this option);
3) Using ldap groups, no need to create "external" user in this case. LDAP groups are retrieved from LDAP server. Then they are mapped to Couchbase user-groups, so there is no need in creation of "external" user in Couchbase Server.

modules/manage/pages/manage-security/configure-ldap.adoc
```diff
@@ -229,39 +313,78 @@ image::manage-security/addGroupTab.png[,90,align=left]
...
+If an LDAP group-mapping has been defined, externally authenticated users can also be recognized as members of one or more corresponding Couchbase-Server user-groups, and thereby assigned the appropriate privileges.
```
timofey-barmin:
```suggestion
If an LDAP group-mapping has been defined, externally authenticated users can also be recognized as members of one or more corresponding Couchbase-Server user-groups, and thereby assigned the appropriate privileges.
```
Yes, but as I explained above it works even without creation of "external" users. Probably need to rephrase or remove.

modules/manage/pages/manage-security/configure-ldap.adoc
```diff
@@ -311,53 +434,70 @@ For example:
...
 This call references the LDAP server at `172.23.125.25`, on port `389`, enabling authorization and authentication for user-credentials to be passed from Couchbase Server.
 
 For more information, see xref:rest-api:rest-configure-ldap.adoc[Configure LDAP].
 
-[#reference-ldap-groups-with-the-rest-api]
-=== Reference LDAP Groups with the REST API
+[#map-groups-with-the-rest-api]
+=== Map Groups with the REST API
+
+Use the `PUT /settings/rbac/groups/admins` method and URI, as follows:
```
timofey-barmin:
```suggestion
Use the `PUT /settings/rbac/groups/<GroupName>` method and URI, as follows:
```

modules/manage/pages/manage-security/manage-users-and-roles.adoc
```diff
@@ -131,28 +115,425 @@ on [.ui]*travel-sample*, the view is as follows:
...
+The fields for the user's names remain, and can be used.
+
+The view provided by the dialog's right-hand panel can be toggled, by means of the *Roles* and *Groups* tabs, which appear above the panel:
+
+image::manage-security/rolesAndGroupsToggle.png[,140,align=middle]
+
+The *Roles* option can be used for externally authenticated users, just as it is used for locally authenticated users.
+The only difference is that no password need be specified for externally authenticated users.
+
+The *Groups* option can only be used when _Native LDAP_ has been established as the means of external authentication, and _LDAP group-support_ has been enabled.
```
timofey-barmin: This is not correct actually. Groups can be used for both local and external users. These are Couchbase groups (not ldap groups). Any couchbase user (doesn't matter local or external) can be explicitly added to Couchbase group.

If native LDAP is configured it becomes possible to use LDAP groups in order to assign roles to users without adding them to Couchbase groups explicitly. In this case users also become members of Couchbase groups but it happens implicitly (the mapping sits in LDAP server in this case). Please let me know if it doesn't make any sense to you.

modules/manage/pages/manage-security/manage-users-and-roles.adoc
```diff
@@ -131,28 +115,425 @@ on [.ui]*travel-sample*, the view is as follows:
...
+
+[#add-an-externally-authenticated-user-by-group]
+==== Add an Externally Authenticated User with Group-Mapping
+
+If the externally authenticated user _is_ to be associated with an LDAP-group mapping, do not assign roles to the user individually.
+Instead, toggle the right-hand panel, by left-clicking on the *Groups* tab:
+
+image::manage-security/groupsTab.png[,140,align=middle]
+
+This displays the Couchbase-Server user-groups that have been mapped to LDAP groups:
```
timofey-barmin: Non necessarily mapped. Couchbase groups can be used without ldap groups. Admin just needs to create a group and add a used to this group (what ldap groups give us is that this is no need to add user to groups manually, this information is taken from LDAP server in this case).

modules/manage/pages/manage-security/manage-users-and-roles.adoc
```diff
@@ -131,28 +115,425 @@ on [.ui]*travel-sample*, the view is as follows:
...
+=== Get User Information with the CLI
+
+To list the cluster's current users, enter the following.
+Note that the command is piped to the `jq` program, to optimize output-readability.
+
+----
+/opt/couchbase/bin/couchbase-cli user-manage --cluster http://10.143.192.101 \
+--username Administrator \
+--password password \
+--list | jq
```
timofey-barmin, line 252: not sure we should use jq in documentation examples. Use it only if it is described somewhere what this is and why it is used. I suggest we remove it.
```suggestion
--list
```

modules/manage/pages/manage-security/manage-users-and-roles.adoc
```diff
@@ -131,28 +115,425 @@ on [.ui]*travel-sample*, the view is as follows:
...
+    ],
+    "external_groups": [],
+    "name": "David Brown"
+  }
+]
+----
+
+The entries include information on the `name` and `roles` of the user, and on the local and external `groups` to which the user belongs.
+
+Information on currently defined _groups_ can similarly be returned:
```
timofey-barmin, line 335: I guess the example doesn't match the description here. List of defined groups can be retrieved by
```
./install/bin/couchbase-cli user-manage --cluster http://10.143.192.101 \
--username Administrator \
--password password \
--list-groups
```

timofey-barmin (changes requested):
Please address the comments I left

modules/manage/pages/manage-security/configure-ldap.adoc
```diff
@@ -188,96 +212,119 @@ For example, a rule-set might be expressed as follows:
...
 Couchbase Server maps the specified username to an LDAP DN, and performs authentication on the LDAP server.
 Notifications confirming success or failure duly appear on the dialog.
 
 [#enable-ldap-group-authorization]
-* *Enable LDAP group authorization & sync*.
+[#group-authorization-enablement]
+=== Group Authorization Enablement
+
+Fields for enabling _LDAP Group Support_ are made visible by means of the *Enable LDAP group authorization & sync* switch.
+Enablement means that Couchbase Server retrieves from the *LDAP Hosts(s)* all _LDAP group membership_ information for users that the *LDAP Host(s)* authenticate.
```
timofey-barmin: Couchbase Server doesn't retrieve all LDAP group membership info. We retrieve group membership info on per user basis only when authorization for this particular is being done (I think it can be called "on-demand").

modules/manage/pages/manage-security/manage-users-and-roles.adoc
```diff
@@ -35,110 +35,178 @@ image::manage-security/usersAndGroupsTab.png[,170,align=left]
...
+To create a `ClusterAdmin` group, each of whose members is granted the *Cluster Admin* role, enter the following:
+
+[#add-new-group-dialog-no-mapping]
+image::manage-security/addNewGroupDialogNoMapping.png[,420,align=left]
+
+As this indicates, no LDAP group mapping is required to define a group intended for the support only of users who are defined on Couchbase Server as _local_ or _external_.
+For an explanation of LDAP group mappings, see xref:manage:manage-security/configure-ldap.adoc#map-ldap-groups-to-couchbase-server-roles[Map LDAP Groups to Couchbase Server Roles].
+
+Save the group by left-clicking on the *Save* button.
+Note that the group will not be displayed by Couchbase Web Console until at least one user has been created and assigned to it.
```
timofey-barmin: Doesn't seem to be correct. I think that's a UI bug. Groups should be visible even if there is no users created and groups are empty.

timofey-barmin (changes requested):

modules/manage/pages/manage-security/configure-ldap.adoc
tonyjhillman: I have modified this throughout.

modules/manage/pages/manage-security/configure-ldap.adoc
tonyjhillman: I've modified throughout.

modules/manage/pages/manage-security/configure-ldap.adoc
tonyjhillman: I think I understand this now, and have modified throughout.

modules/manage/pages/manage-security/configure-ldap.adoc
tonyjhillman: Thanks.

modules/manage/pages/manage-security/configure-ldap.adoc
tonyjhillman: Thanks, fixed.

modules/manage/pages/manage-security/configure-ldap.adoc
tonyjhillman: Typo, fixed. Thanks.

modules/manage/pages/manage-security/configure-ldap.adoc
tonyjhillman: I have added the link.

modules/manage/pages/manage-security/configure-ldap.adoc
tonyjhillman: I have modified as described.

modules/manage/pages/manage-security/configure-ldap.adoc
tonyjhillman: Changed.

modules/manage/pages/manage-security/configure-ldap.adoc
tonyjhillman: I think I understand. I have changed to:

If no match is found, Couchbase Server assumes that the DN is the username presented to Couchbase Server by the user who is attempting to gain access (in other words, assumes that a username such as `cn=exampleUser,dc=example,dc=com` has been presented).

modules/manage/pages/manage-security/configure-ldap.adoc
tonyjhillman: Thanks, fixed.

modules/manage/pages/manage-security/configure-ldap.adoc
tonyjhillman: Fixed.

modules/manage/pages/manage-security/configure-ldap.adoc
tonyjhillman: I haven't added it here explicitly, but have modified the passage, and I've added it elsewhere, particularly in the overview. See what you think.

modules/manage/pages/manage-security/configure-ldap.adoc
tonyjhillman: Thanks, fixed.

modules/manage/pages/manage-security/manage-users-and-roles.adoc
tonyjhillman: Thanks, I understand this now. I have changed this page accordingly, I hope in the right ways.

modules/manage/pages/manage-security/manage-users-and-roles.adoc
tonyjhillman: I understand, and have changed this.

modules/manage/pages/manage-security/manage-users-and-roles.adoc
tonyjhillman: I've actually used it in quite a few places in the docs generally, in that it's the only way to get some of the output readable. But anyway, I've changed this reference here to do what I normally do, which is link to the external page describing it.

modules/manage/pages/manage-security/manage-users-and-roles.adoc
tonyjhillman: Thanks, I've fixed this.

modules/manage/pages/manage-security/configure-ldap.adoc
tonyjhillman: I have changed it to:

Enablement means that when a user has been authenticated by the *LDAP Hosts(s)*, Couchbase Server retrieves the _LDAP group membership_ information for that user, in order to authorize them.

modules/manage/pages/manage-security/manage-users-and-roles.adoc
tonyjhillman: Okay, I have removed this statement.

modules/manage/pages/manage-security/configure-ldap.adoc
```diff
@@ -188,96 +216,120 @@ For example, a rule-set might be expressed as follows:
...
-+
+
 image::manage-security/ldapQueryDetail.png[,320,align=left]
 
-** The *Traverse nested groups* checkbox, when checked, allows nested groups to be traversed by the search.
+[#traverse-nested-groups]
+==== Traverse Nested Groups
+
+The *Traverse nested groups* checkbox, when checked, allows nested groups to be traversed by the search.
+Couchbase Server uses the same search to find groups of groups recursively (with each group's DN being substituted for `%D`).
```
timofey-barmin: Not sure if it is clear here that we are talking about the search that is described above ('User's Attributes' or 'LDAP Query'). Maybe we need to rephrase. If you are sure that it's clear enough, let's keep it.

modules/manage/pages/manage-security/configure-ldap.adoc
```diff
@@ -375,16 +416,20 @@ This brings up the *Edit Group Admins* dialog:
 image::manage-security/editGroupAdmins.png[,420,align=middle]
 
 Within this dialog, the description, mapping, or roles for the group can be edited.
-For details of selecting roles within the *Roles* panel, see xref:manage:manage-security/manage-users-and-roles.adoc[Manage Users and Roles].
+For details of selecting roles within the *Roles* panel, see xref:manage:manage-security/manage-users-and-roles.adoc[Manage Users, Groups, and Roles].
 
+Note that if the *Delete Group* button is left-clicked on, the group is deleted.
+This means that all mappings between LDAP groups and the roles that were assigned to this group are also deleted.
+However, if users registered on Couchbase Server as either _local_ or xref:manage:manage-security/configure-ldap.adoc#adding-an-externally-authenticated-user[external] were allocated roles by means of this group, the deletion of this group causes those users to be individually assigned the roles they previously held through membership of the group; and consequently, the users retain their privileges.
```
timofey-barmin: As we discussed in slack, this sentence should be modified or removed

modules/manage/pages/manage-security/manage-users-and-roles.adoc
```diff
@@ -35,110 +35,178 @@ image::manage-security/usersAndGroupsTab.png[,170,align=left]
...
+image::manage-security/accessGroupsButton.png[,150,align=left]
+
+The *Groups* view is now displayed:
+
+image::manage-security/groupSecurityRow.png[,800,align=left]
+
+As with the *Users* view, left-clicking on a group's row displays controls that include *Delete* and *Edit* options.
+Left-click on the *Edit* button to display the *Edit Group <group-name>* dialog, which is similar to the *Add New Group*
+dialog examined above, and allows all the group's attributes, except the name, to be modified.
+Note that if a group still containing one or more users is deleted, those users are directly assigned the roles they previously held through membership of the now-deleted group, and thereby retain their privileges.
```
timofey-barmin: As discussed in slack, this needs to be either modified or removed

timofey-barmin (changes requested):
Thank you for this work Anthony. Please address the three minor comments that I added and I think we can merge.

modules/manage/pages/manage-security/configure-ldap.adoc
tonyjhillman: I've removed.

modules/manage/pages/manage-security/manage-users-and-roles.adoc
tonyjhillman: It now just says: 
Note that if the Delete Group button is left-clicked on, the group is deleted. This means that all mappings between LDAP groups and the roles that were assigned to this group are also deleted.

modules/manage/pages/manage-security/configure-ldap.adoc
tonyjhillman: So, what I've done is include this information for the first subject, User's Attributes; then create pointers back to that information, from the second subject, LDAP Query. Is there any difference in the way things are applied? If so, I should write a separate section for each.

modules/manage/pages/manage-security/manage-users-and-roles.adoc
```diff
@@ -587,6 +587,19 @@ http://10.143.192.101:8091/settings/rbac/users
 If successful, the command provides as its output a document containing an entry for each of the current users.
 This output is identical in form to that shown above, in xref:manage:manage-security/manage-users-and-roles.adoc#get-user-information-with-the-cli[Get User Information with the CLI].
 
+[#get-group-information-with-the-rest-api]
+=== Get Group Information with the REST API
+
+To list the cluster's current users, use the `GET /settings/rbac/groups` method and URI as follows:
```
timofey-barmin: users -> groups

modules/manage/pages/manage-security/manage-users-and-roles.adoc
```diff
@@ -587,6 +587,19 @@ http://10.143.192.101:8091/settings/rbac/users
...
+
+----
+curl -v -X GET -u Administrator:password \
+http://10.143.192.101:8091/settings/rbac/groups
+----
+
+If successful, the command provides as its output a document containing an entry for each of the current users.
```
timofey-barmin: users->groups

timofey-barmin (approved):
Please fix the typos (see comments) and merge

modules/manage/pages/manage-security/manage-users-and-roles.adoc
tonyjhillman: Thanks. Fixed.

modules/manage/pages/manage-security/manage-users-and-roles.adoc
tonyjhillman: Fixed.

tonyjhillman: Thanks very much!

PR #551, opened 2019-06-26
DOC-5424: Added new roles and privileges to Data Backup

tonyjhillman:


simon-dew (approved):
LGTM. I have reservations about the shortened column names, but that is part of the broader issue around wide tables, so happy to approve this as-is for now.

PR #554, opened 2019-06-28
DOC-5293: Bitwise functions (port to Query)

simon-dew:

* [Bitwise Functions](https://github.com/couchbase/docs-server/blob/b8f9362ab48b9d2901811390a844d9b931a18472/modules/n1ql/pages/n1ql-language-reference/bitwisefun.adoc) — Updated Query document with corrections from Analytics

Documentation issue: [DOC-5293](https://issues.couchbase.com/browse/DOC-5293)

### DIFF
--- a/modules/ROOT/nav.adoc
+++ b/modules/ROOT/nav.adoc
@@ -218,7 +218,7 @@
   *** xref:n1ql:n1ql-language-reference/functions.adoc[Functions Overview]
   *** xref:n1ql:n1ql-language-reference/aggregatefun.adoc[Aggregate Functions]
   *** xref:n1ql:n1ql-language-reference/arrayfun.adoc[Array Functions]
-  *** xref:n1ql:n1ql-language-reference/bitwisefun.adoc[Bitwise Functions in N1QL]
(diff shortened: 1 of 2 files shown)


simon-dew:
@htowaileb comments:

- [x] Link for `expressions` ...  should work properly once merged. No harm confirming that though if possible.

- [x] BitClear: Remove the statement above the `note` stating that `zero or negative has no impact`, it has an impact, and it returns `null`, it’s written below it in a `Note box`. (So just deleting the former statement will do)

- [x] BitNot: Arguments -> `that contains the target bits to reverse` instead of `that contains the target bit or bits to reverse`, that’s appropriate for functions that have `value/array of positions`, *BitNOT* does not have that.

- [x] BitTest: Minor: There is no need to color the result of FALSE and TRUE in the queries with the red color, I see no harm in keeping it though.

- [x] BitXor:Return Value: representing the bitwise XOR between `the input integers` instead of `the two input integers`

simon-dew:
> There is no need to color the result of FALSE and TRUE in the queries with the red color

The red is applied by GitHub to mark a possible error in the JSON. Changed TRUE and FALSE to lower case to fix this.

htowaileb (approved):
Looks good.

PR #558, opened 2019-07-04
Update install-ports.adoc

johnsonvv:
Added broader explanation on SSL/TLS.


amarantha-k:
SSL/TLS is a standard concept and there's plenty of information available on the internet. Hence we'd prefer to not include any external links.

PR #563, opened 2019-07-09
Release notes draft for Server 5.5.5

amarantha-k:


modules/release-notes/pages/relnotes.adoc
```diff
@@ -5,14 +5,109 @@ This release specifically highlights agility, performance, and manageability and
...
+=== Fixed Issues
+
+==== Cross Datacenter Replication (XDCR)
+
+[#table_fixedissues_xdcr_555,cols="25,66"]
+|===
+| Issue | Description
+
+| https://issues.couchbase.com/browse/MB-34297[MB-34297^]
+| The topology change detector, upon detecting that a target bucket no longer exists, would delete the replication spec right away. However, in case the target node was being failed over adn had stopped serving the bucket, this could cause the replication spec to be incorrectly deleted. This has been fixed and the topology change detector does not delete replication spec when the target bucket cannot be found.
```
mihirkamdarqe: spell mistake - adn instead of and

nelio2k: > the topology change detector does not delete replication spec when the target bucket cannot be found

XDCR no longer deletes replication specs incorrectly during transient rebalancing states.

_Explanation: If the target bucket is removed, XDCR still will remove the replication spec using a non-topology-change-detector module. This removal mechanism is much more fail-safe._

nelio2k (changes requested):

mihirkamdarqe (approved):

PR #566, opened 2019-07-11
Updated the old blog link present on the page

Shweta0489:
Updated the old blog link "http://blog.couchbase.com/2016/january/moving-sql-business-logic-to-the-application-layer" to "https://blog.couchbase.com/moving-sql-business-logic-to-the-application-layer/".


RichardSmedley (approved):
LGTM :+1:

PR #600, opened 2019-07-18
DOC-4998 Consolidate to one init cluster setup

eric-schneider:
Cluster provisioning instructions were being duplicated, so the main point of this change is to remove the stale version from the Install module, and point existing references to the version in the Manage module.

While performing this work, some cleanup was performed on the consolidated version, including:
* New screenshots with higher resolution, alt text, and in some cases, additional content that was missing
* Removed references to Mac filesystem paths since they were no longer present in screenshots
* Minor AsciiDoc and style consistency cleanup

@tonyjhillman This was a last-minute decision to remove the duplicate content in the Install section. I know that you have been working in this area quite recently, so if you have local changes that I'm stepping on, let me know and we can sort it out. Otherwise, I think what I've done here is made some pretty benign, yet very valuable updates.


tonyjhillman (approved):
BTW, I think this has been edited in the meantime, so you may encounter some clashes.

PR #646, opened 2019-07-31
DOC-5583: Fix query timings curl

JFlath:


modules/manage/pages/monitor/monitoring-n1ql-query.adoc
```diff
@@ -153,7 +153,7 @@ To set query settings using the REST API, specify the parameters in the request
 
 [source,console]
 ----
-$ curl -v http://localhost:8093/query/service -d "profile='timings'&statement=SELECT * FROM default LIMIT 1";
+$ curl -v http://localhost:8093/query/service -d "profile=timings&statement=SELECT * FROM default LIMIT 1";
```
simon-dew: semi-colon should be within the quotes (`;"` not `";`) otherwise it's interpreted as a bash command separator

simon-dew (changes requested):
Thanks @JFlath , just two minor corrections.

modules/manage/pages/monitor/monitoring-n1ql-query.adoc
simon-dew: Also the outer double-quotes (`"`) should be single quotes (`'`)

simon-dew: So overall, I think it should be: 

```
curl -v http://localhost:8093/query/service -d 'profile=timings&statement=SELECT * FROM default LIMIT 1;'
```

JFlath: Heh, yeah, as it happens I don't believe the `;` is needed as part of the statement itself, so if it gets interpreted as a bash separator it's benign. Should be cleaned up though.

Any particular reason for changing the quote type? Just a style thing?

simon-dew: Happy for you to remove the `;` altogether if you want!

When I tested that statement, if I used double quotes `"` I got a N1QL syntax error at LIMIT. If I used single quotes `'` it worked. I haven't got to the bottom of that one yet.

simon-dew (approved):

PR #647, opened 2019-07-31
DOC-4861: TLS minimum version and additional security updates

tonyjhillman:


tonyjhillman:
This PR contains various updates, across the security pages.
The main ones are the two new pages, manage-tls.adoc, and rest-security-setting.adoc.

modules/manage/pages/manage-security/manage-tls.adoc
```diff
@@ -0,0 +1,160 @@
...
+
+[#tls-and-cipher-suites]
+== TLS and Cipher-Suites
+
+_Transport Layer Security_ (TLS) is a protocol that provides security over a computer network.
+Couchbase Server supports the configuration of TLS, to secure communications between cluster-nodes, clusters, and external clients: this includes selection of an appropriate TLS version.
+
+TLS supports multiple methods for exchanging keys, encrypting data, and authenticating message-integrity.
+Appropriate methods, supported by both participants in an intended networked communication, can be specified through selection of an appropriate _cipher-suite_, from defined lists of ones acceptable.
+By default, Couchbase Server uses the list of cipher-suites provided by its underlying operating system; however, Couchbase Server also allows a _customized_ list to be defined and used instead.
```
timofey-barmin: I think that's not exactly true. Please read the "2.2 Cipher Suites" paragraph in https://docs.google.com/document/d/1kQPo3q1Mu0vrBLoVAfpgkttsosyNuQpSARkTuayT8rw/edit
(please note that this doc says that we use openssl 1.0.1, but actually we already switched to 1.1.(something))
I don't think that we should provide detailed description for each service, but it's worth saying that we use "HIGH" ciphers or something like that. Please choose the right wording.

modules/manage/pages/manage-security/manage-tls.adoc
```diff
@@ -0,0 +1,160 @@
...
+
+When a TLS connection is established between a client application and Couchbase Server, using the secure port 18091, a _handshake_ occurs, as defined by the https://en.wikipedia.org/wiki/Transport_Layer_Security#TLS_handshake[TLS Handshake Protocol].
```
timofey-barmin: 18091 is only one of the ssl ports we use, so maybe it is worth saying: "for example, using the secure port 18091"

modules/manage/pages/manage-security/manage-tls.adoc
```diff
@@ -0,0 +1,160 @@
...
+-p password \
+--set \
+--tls-honor-cipher-order 1 \
+--cipher-suites TLS_RSA_WITH_AES_128_CBC_SHA,TLS_RSA_WITH_AES_256_CBC_SHA
+----
+
+The `set` parameter establishes use of the command to make settings.
+The `tls-honor-cipher-order` parameter-value is specified as `1`, meaning that the server's preference-order for cipher-suites will be used, rather than the client's.
+(Note, however, that this setting is also the default; and so will remain established even if this parameter is not specified.)
+The `cipher-suites` parameter takes a value that is a list of cipher-suites to be used for the cluster.
```
timofey-barmin, line 100: 1) It is worth adding that command accepts IETF RFC names for cipher suites (there two naming standards: IETF RFC and OpenSSL).  
2) Please also add that empty list [] means that ciphers are not set (default value). It means that default set of ciphers will be used.

modules/rest-api/pages/rest-setting-security.adoc
```diff
@@ -0,0 +1,173 @@
...
+
+Allows settings to be established for the use of encryption.
+This includes the encryption _level_, for node-to-node encryption within the cluster;
+UI disablement over insecure connections; TLS minimum version; and a list of server-accepted cipher suites.
+
+[#curl-syntax]
+== Curl Syntax
+
+----
+curl -X GET -d hostname=<host>:<port>
```
timofey-barmin: This request seems to be wrong. I think you meant:
```
curl -u <administrator>:<password> \
  http://<host>:<port>/settings/security
```

modules/rest-api/pages/rest-setting-security.adoc
```diff
@@ -0,0 +1,173 @@
...
+  -d user=<administrator>
+  -d password=<password>
+  -u <administrator>:<password>
+
+curl -X POST -d hostname=<host>:<port>
```
timofey-barmin: This request seems to be wrong. You don't need to pass hostname, user, password. On other hand it should as follows:
```
curl -X POST -u <administrator>:<password> \
  -d disableUIOverHttp=<true|false> \
  -d disableUIOverHttps=<true|false> \
  -d clusterEncryptionLevel=<all|control> \
  -d tlsMinVersion=<tlsv1|tlsv1.1|tslv1.2> \
  -d honorCipherOrder=<true|false> \
  -d cipherSuites=<list-of-accepted-cipher-suites> \
  http://<host>:<port>/settings/security

```

modules/rest-api/pages/rest-setting-security.adoc
```diff
@@ -0,0 +1,173 @@
...
+* `disableUIOverHttp`.
+Whether access to Couchbase Web Console should be disabled over http.
+Default is `false`, meaning that the console can be accessed over http.
+
+* `disableUIOverHttps`.
+Whether access to Couchbase Web Console should be disabled over https.
+Default is `false`, meaning that the console can be accessed over https.
+
+* `clusterEncryptionLevel`.
+Controls the level of encryption imposed on inter-node communications within the cluster.
```
timofey-barmin, line 49: It only works when node-to-node encryption is turned on. If it doesn't there is no encryption neither for control traffic, nor for data traffic

modules/rest-api/pages/rest-setting-security.adoc
```diff
@@ -0,0 +1,173 @@
...
+Can be either `control` (meaning that only server-management information is passed in encrypted form) or `all` (meaning that all information, including data handled by services, is passed in encrypted form).
+
+* `tlsMinVersion`.
+Specifies the minimum TLS version accepted by the cluster.
+Can be `tlsv1`, `tlsv1.1`, or `tlsv1.1`.
+The default is `tlsv1`.
+This parameter can only be set when encryption between cluster-nodes has been explicitly enabled: see xref:manage:manage-nodes/apply-node-to-node-encryption.adoc[Apply Node-to-Node Encryption], for details.
```
timofey-barmin: It actually can be set when n2n encryption is disabled. This sentence should be moved up to 'clusterEncryptionLevel' description

modules/rest-api/pages/rest-setting-security.adoc
```diff
@@ -0,0 +1,173 @@
...
+
+* `honorCipherOrder`.
+Specifies whether the cluster uses its own cipher-suite preference, rather than the client's.
+The cluster's list of accepted cipher-suites can be defined with the `cipherSuites` flag (see below).
+The default value of `honorCipherOrder` is `true`: a setting of `false` is _not_ recommended, since insecure.
+
+* `cipherSuites`.
+Specifies a list of the cipher suites to be used by the cluster, in order of preference.
```
timofey-barmin, line 64: Please also add that empty list [] means that ciphers are not set (default value). It means that default set of ciphers will be used.

timofey-barmin (changes requested):
Please address my review comments

modules/manage/pages/manage-security/manage-tls.adoc
tonyjhillman: OK, I have said "HIGH".

modules/manage/pages/manage-security/manage-tls.adoc
tonyjhillman: I've made the comment about naming standards at the introduction to the section (I'm assuming it applies both the REST and CLI). I've added the empty list for both REST and CLI as [] and "" respectively.

modules/rest-api/pages/rest-setting-security.adoc
tonyjhillman: Thanks.

modules/rest-api/pages/rest-setting-security.adoc
tonyjhillman: Thanks.

modules/rest-api/pages/rest-setting-security.adoc
tonyjhillman: Have added.

modules/rest-api/pages/rest-setting-security.adoc
tonyjhillman: Have removed this remark from this paragraph.

modules/rest-api/pages/rest-setting-security.adoc
tonyjhillman: Done.

modules/manage/pages/manage-security/manage-tls.adoc
timofey-barmin: Sorry for the confusion. I meant that we only support IETF RFC names. We don't accept OpenSSL names.

tonyjhillman: ok

timofey-barmin (approved):

tonyjhillman:
Thanks very much.

PR #659, opened 2019-08-06
DOC-5157: Document system keyspaces

simon-dew:

* [Monitor Queries › system:functions](https://simon-dew.github.io/docs-site/DOC-4941/server/6.5/manage/monitor/monitoring-n1ql-query.html#sys-functions)
* [Monitor Queries › system:functions_cache](https://simon-dew.github.io/docs-site/DOC-4941/server/6.5/manage/monitor/monitoring-n1ql-query.html#sys-functions-cache)

Documentation issue: [DOC-5157](https://issues.couchbase.com/browse/DOC-5157)

### DIFF
--- a/modules/manage/pages/monitor/monitoring-n1ql-query.adoc
+++ b/modules/manage/pages/monitor/monitoring-n1ql-query.adoc
@@ -1946,7 +1946,7 @@ For more examples, take a look at the blog: https://blog.couchbase.com/optimize-
 
 include::ROOT:partial$developer-preview.adoc[tag=admonition]
 
-This catalog maintains a list of all user-defined functions.
+This catalog maintains a list of all user-defined functions across all nodes.
 To see the list of all user-defined functions, use:
 
 [source,n1ql]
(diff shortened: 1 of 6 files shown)


marcogrecopriolo:
2 things
- each n1ql node has its own cache, so you are likely to see the same function loaded in multiple nodes.
In other words, system:functions lists all the functions that are known (this is global), system:functions_cache lists the functions that each node has used.
On that front, we may want to document the "node" field too.

- the functions cache caches also functions that do not exists. It does sound strange, but the general idea is that if by some coding mistake, the same non existent function is being called over and over again, this would generate a bottleneck on the functions metadata storage.
By caching what we know does not exist, we don't need to go there.
A non existent cache entry looks like this:

cbq> execute function notFound();
{
    "requestID": "42e3e328-f5fc-4aef-88aa-9bf32b407e9f",
    "signature": null,
    "results": [
    ],
    "errors": [
        {
            "code": 10101,
            "msg": "Function not found notFound"
        }
    ],
    "status": "errors",
    "metrics": {
        "elapsedTime": "9.716885ms",
        "executionTime": "9.568876ms",
        "resultCount": 0,
        "resultSize": 0,
        "errorCount": 1
    }
}
cbq> select * from system:functions_cache;
{
    "requestID": "c4f9fbe5-0584-4de9-b9df-5a44d602db36",
    "signature": {
        "*": "*"
    },
    "results": [
    {
        "functions_cache": {
            "avgServiceTime": "22.79µs",
            "lastUse": "2019-08-06 16:59:51.630317391 +0100 BST m=+17.966707065",
            "maxServiceTime": "22.79µs",
            "minServiceTime": "0s",
            "name": "notFound",
            "namespace": "default",
            "node": "127.0.0.1:8091",
            "type": "global",
            "undefined_function": true,
            "uses": 1
        }
    }
    ],
    "status": "success",
    "metrics": {
        "elapsedTime": "13.519924ms",
        "executionTime": "13.358927ms",
        "resultCount": 1,
        "resultSize": 446
    }
}

PR #675, opened 2019-08-07
DOC-2496: Extended settings while creating an index from UI

abhinavdangeti:


modules/fts/pages/fts-creating-indexes.adoc
```diff
@@ -517,6 +517,26 @@ Note that the type of an index is saved in its JSON definition, which can be pre
...
 
+[#index-partitions]
+== Index Partitions
+
+The *Index Partitions* interface provides a section to enter the number of partitions the index is to be split into:
+
+[#fts_index_partitions_interface]
+image::fts-index-partitions-interface.png[,300,align=left]
+
+The default option for this setting is 6. Note that this number will represent the number of active partitions for an index, and will be distributed across all the nodes in the cluster where the search service is running.
```
amarantha-k: Suggest rewording to:

Note that this number represents the number of active partitions for an index, and the active partitions are distributed across all the nodes in the cluster where the search service is running.

PR #712, opened 2019-08-19
Fix backtick rendering

JFlath:
Would appreciate if someone with the skills to actually preview this properly could double-check I haven't made things worse... ;)

### DIFF
--- a/modules/n1ql/pages/n1ql-language-reference/reservedwords.adoc
+++ b/modules/n1ql/pages/n1ql-language-reference/reservedwords.adoc
@@ -11,7 +11,7 @@ Some of the keywords are not currently implemented but are reserved for future u
 == Using Reserved Words as Identifiers
 
 N1QL allows escaped identifiers to overlap with keywords.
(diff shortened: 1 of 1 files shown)


modules/n1ql/pages/n1ql-language-reference/reservedwords.adoc
```diff
@@ -11,7 +11,7 @@ Some of the keywords are not currently implemented but are reserved for future u
 == Using Reserved Words as Identifiers
 
 N1QL allows escaped identifiers to overlap with keywords.
-To use a reserved word as an identifier, you must escape it by enclosing the reserved word inside backticks (``).
+To use a reserved word as an identifier, you must escape it by enclosing the reserved word inside backticks (```).
```
simon-dew, line 14: That works, but looking at the original text I guess there are supposed to be two backticks?
I always use a predefined character attribute when I want an actual backtick to appear: {backtick}
So you could try `{backtick}{backtick}`

simon-dew (changes requested):
Thanks for the PR James, I only just noticed it. I guess no-one saw it because you haven't selected any reviewers.

PR #729, opened 2019-09-05
DOC-5716: Index advisor

simon-dew:

* [Query Workbench › Index Advisor](https://simon-dew.github.io/docs-site/DOC-5716/server/6.5/tools/query-workbench.html#index-advisor)

Documentation issue: [DOC-5716](https://issues.couchbase.com/browse/DOC-5716)

The `ADVISE` statement and the `ADVISOR` function will be documented in separate pull requests.


chang27:
Hi Simon,
I think this document explains more about UI design of Index Advisor, please consider to add UI team to review.

- [x] I am not sure how UI design is for Index Advisor for the part: "an ADVISE command is **automatically** run in the background to generate index advice for that query. ", for the design in the backend, we need execute N1QL query: advise + query statement. I wonder whether it will automatically run "advise" on one query as it will add extra workload to server engine.

- [x] "Covered Index" -> "Covering Index". Covering Index is an optional part in the output, given that not all the queries qualify for it.

- [x] Actually "Index recommendation" is the main and key part of the output, that is based on the predicates of the query. If we can not come up with any index candidate here, we say that we can not recommend index.

- [x] Please have a review on the output from "https://blog.couchbase.com/index-advisor-for-n1ql-query-statement/",  we have another field called "recommending_rule" inside "index recommendation" which is not included in the UI design, that is to explain how we come up with the recommended indexes.

Please let me know if there is any question, thanks!
Chang

PR #767, opened 2019-09-26
DOC-5714: Cost-Based Optimizer

simon-dew:

* [Cost-Based Optimizer](https://simon-dew.github.io/docs-site/DOC-5714/server/6.5/n1ql/n1ql-language-reference/cost-based-optimizer.html)
* [UPDATE STATISTICS](https://simon-dew.github.io/docs-site/DOC-5714/server/6.5/n1ql/n1ql-language-reference/updatestatistics.html)

Documentation issue: [DOC-5714](https://issues.couchbase.com/browse/DOC-5714)


simon-dew:
Note for reviewers. The Cost-Based Optimizer document intentionally does not go into all the details on how selectivity and cardinality are estimated, or how confidence is used to set the sample size. It aims to give a general overview, with emphasis on the distribution statistics so that users can understand the options for the `UPDATE STATISTICS` statement.

simon-dew:
(Bingjjie's comments)

**Update statistics**

- [x] Syntax diagram

  In the syntax diagram, we should not use “index-key” to describe fields for update statistics. Although it’s common to collect statistics on index fields, it is not limited to index fields. A query can have predicates on non-indexed fields and statistics can be collected on those fields as well. For example, you can have:
  ``` 
  Create index ix1 on default(c1, c2)
  Update statistics for default(c1, c2, c3)
  Select * from default where c1 = 10 and c2 > 5 and c3 between 20 and 40
   ```
  I think the original document used something like `<index-expression>`, which means any expression that is supported as an index key can also be used in update statistics statement.
 
- [x] Index key

  Remove the 3rd bullet referring to PAIRS() function, I’ve not tested this one.

- [x] Usage

  This section may need a bit rewording, you are assuming all update statistics fields are index fields

**Cost-based optimizer**
 
- [x] Summary (very beginning), takes into account the cost of memory, CPU, and disk usage when ….

  Also need to include “network” or “network transport”
 
- [x] Operations, 3rd point

  The description of “cardinality” is incorrect, it’s not a measure of the uniqueness of the data values, it’s the number of documents remaining after all applicable predicates are applied

PR #802, opened 2019-10-12
Fix links to index join clause

manishbansal8843:


modules/n1ql/pages/n1ql-language-reference/selectintro.adoc
```diff
@@ -380,7 +380,7 @@ See xref:n1ql-language-reference/subqueries.adoc[Subqueries] for more details.
 [source,N1QL]
 ----
 SELECT *
-FROM   (SELECT t.airportname
+FROM   (SELECT s1.airportname
```
simon-dew: Reject this change. In the original example, the inmost subquery selects a single airport in the United States and gives it the alias `t`. The next subquery selects the airport name from `t` and gives it the alias `s1`. The outermost query selects everything from `s1` and gives it the alias `s2`. If we make the change given in this line of the pull request, the query returns an empty object, and the results in lines 391-401 are incorrect.

Revert this line to the original as follows:

```suggestion
FROM   (SELECT t.airportname
```

modules/n1ql/pages/n1ql-language-reference/selectintro.adoc
```diff
@@ -330,7 +330,7 @@ In this example:
 
 This example illustrates a special kind of JOIN where the documents on the right side of JOIN contain the foreign key reference to the documents on the left side.
 Such JOINs are referred to as index JOIN.
-For details, see xref:n1ql-language-reference/from.adoc#index-join-clause[Index JOIN Clause].
+For details, see xref:n1ql-language-reference/join.adoc#index-join-clause[Index JOIN Clause].
```
simon-dew, line 333: LGTM 👍

simon-dew (changes requested):
Change on line 383 needs to be reverted

amarantha-k (approved):

simon-dew (approved):
LGTM 👍

PR #817, opened 2019-10-22
DOC-4401: Remove references to config options

b33f:
See MB-23166


eric-schneider (approved):
lgtm - though, we'll need to circle back to [DOC-2153](https://issues.couchbase.com/browse/DOC-2153) to ensure the full doc story is completed.

PR #847, opened 2019-11-05
Doc 5948

tonyjhillman:


modules/learn/pages/data/transactions.adoc
```diff
@@ -0,0 +1,56 @@
+= Distributed ACID Transactions
+
+[abstract]
+_Distributed ACID Transactions_ guarantee that multiple documents can be updated atomically.
+
+[#transactions-and-couchbase-server]
+== Transactions and Couchbase Server
+
+_Transactions_ are operations that ensure that when multiple discrete data-elements need to be modified such that only the successful modification of all justifies the modification of any, _either_ all the modifications do occur successfully; _or_ none of them occurs.
```
programmatix, line 9: I haven't really seen 'data-elements' before.. is there a reason not to use the simpler 'documents'?

programmatix: NIT: Both 'occur' and 'occurs' are used, implying modifications is both plural and singular.

modules/learn/pages/data/transactions.adoc
```diff
@@ -0,0 +1,56 @@
+= Distributed ACID Transactions
+
+[abstract]
+_Distributed ACID Transactions_ guarantee that multiple documents can be updated atomically.
+
+[#transactions-and-couchbase-server]
+== Transactions and Couchbase Server
+
```
programmatix, line 8: I feel we've lost the 'sales pitch' of the current docs here; we're straight into dry technical details without selling the benefits of our transactions.

modules/learn/pages/data/transactions.adoc
```diff
@@ -0,0 +1,56 @@
...
+_Transactions_ are operations that ensure that when multiple discrete data-elements need to be modified such that only the successful modification of all justifies the modification of any, _either_ all the modifications do occur successfully; _or_ none of them occurs.
+Transactions are implemented by Couchbase Server as _Distributed ACID Transactions_.
+
+A Distributed ACID Transaction can be performed on any number of documents.
+For each document, the transaction makes a copy of the document-data, and saves the copy temporarily in the xref:learn:data/extended-attributes.adoc[Extended Attributes] area of the document's meta data.
+Updates and modifications are then made to the key-value pairs contained by this meta data copy: consequently, the changes that occur during the transaction prior to commitment &#8212; these constituting _uncommitted_ (or _dirty_) data &#8212; are non-readable by any operation other than this transaction itself.
```
programmatix, line 14: This makes it sound like there's separate operations to copy the data into xattrs, _then_ apply modifications.  That sounds slow.  Instead, the app gets a document inside a transaction.  They make all modifications to that in-memory of the version of the document.  Then they call ctx.replace(), which puts that modified copy into the xattrs.

modules/learn/pages/data/transactions.adoc
```diff
@@ -0,0 +1,56 @@
...
+Only following _commit_ of the documents modified by the transaction &#8212; when the modified data-copy is removed from the meta data area, and is written over the main body of the document's data &#8212; do the corresponding changes become readable by other transactions and operations.
+This isolation-level is known as https://jepsen.io/consistency/models/read-committed[Read Committed].
+
+Note that this use of a document's extended attributes area results in an approximate doubling of its pre-transaction size, while the transaction is underway.
+In consequence, transactions can only be used on documents whose maximum size is 10 MB (which is half of the maximum-permitted document-size, 20 MB).
+
+As well as making use of individual documents' extended attributes, Distributed ACID Transactions also create additional documents in each bucket that they access.
+These include _Active Transaction Records_ (up to 1024 of which can exist per bucket, and whose names are prefixed with `atr-`), and _Transaction Client Records_ (one of which exists per bucket, duly named `txn-client-record`).
```
programmatix, line 22: Naming is `_txn:atr-` and `_txn:client-record` now

modules/learn/pages/data/transactions.adoc
```diff
@@ -0,0 +1,56 @@
...
+These documents are automatically maintained by Couchbase Server, and should not be modified by any application.
+They have no impact on use of either the _Query Service_ or the _Search Service_.
```
programmatix, line 24: Is this sentence needed?  As a reader I think it would have me wondering "... so it _does_ affect views, eventing and analytics?"

modules/learn/pages/data/transactions.adoc
```diff
@@ -0,0 +1,56 @@
...
+Note that transaction records are not written to nodes that do not contain data to be updated: consequently, only nodes that _do_ contain data to be updated are affected by a transaction.
```
programmatix, line 25: I had to reread this 3 times, maybe we can find a simpler way of putting it.

modules/learn/pages/data/transactions.adoc
```diff
@@ -0,0 +1,56 @@
...
+
+No two Distributed ACID Transactions can access the same document at the same time.
+If such an attempt occurs, only one of the transactions can proceed; while the other must await commitment of the data, before itself being able to modify it.
```
programmatix, line 28: Well 'access' as in reads is fine, it's two concurrent writes that fails.
The language here makes it sound like this is a limitation or problem I think?  Detecting write-write conflicts is a good thing :) We could instead here discuss how such conflicts are handled transparently and automatically for the app, through retrying the transaction (this is a benefit of ours over other some competitor implementations).

modules/learn/pages/data/transactions.adoc
```diff
@@ -0,0 +1,56 @@
...
+Distributed ACID Transactions perform commits by means of _durable writes_.
+Each commit provides _eventual consistency_, which is consistent with the _single-document write_ model of durable writes; as described in xref:learn:data/durability.adoc[Durability].
+
+[#dependencies-and-limitations]
+=== Dependencies and Limitations
+
+_Distributed ACID Transactions_ constitute an evolving technology.
+Currently, the following dependencies and limitations apply:
+
+* Transactions require the Enterprise Edition 6.5 of Couchbase Server.
```
programmatix, line 39: No, it's Community Edition too.  6.5 is correct.

modules/learn/pages/data/transactions.adoc
```diff
@@ -0,0 +1,56 @@
...
+They also require the Couchbase xref:3.0@java-sdk::start-using-sdk.adoc[Java SDK], version 3.0 or after.
+Additionally, it is strongly recommended that _Network Time Protocol_ (NTP) be used to synchronize time across all cluster-nodes: this ensures that abandoned transactions are cleaned up in a timely manner.
+See xref:learn:install/synchronize-clocks-using-ntp.adoc[Clock Sync with NTP].
+
+* Only documents whose initial size is 10 MB or less can be included in a transaction.
+
+* Commits are performed by means of _durable writes_, which are single-document only.
+Consequently, _eventual consistency_ is achieved as the writes occur.
```
programmatix, line 47: Well no, one of the big things about durable writes is that they're not eventually consistent: the operation completely blocks until the desired consistency is achieved.

modules/learn/pages/data/transactions.adoc
```diff
@@ -0,0 +1,56 @@
...
+Therefore, to ensure atomicity, non-transactional updates (such as might be performed by means of the _Data Service_ or _Query Service_) should not be made to any document involved in a transaction while the transaction is itself in progress.
```
programmatix, line 48: I divide atomicity into two types (I'm going to run this by Shivani to see if she agrees):

_Write_ atomicity is making sure that either all the docs are written, or none are (rollback).  This is the 'A' in ACID, and we have this.

_Read_ atomicity is the impression (if not the reality) that a commit happens in one atomic step.  As a distributed database, we cannot do this (without destroying performance, at least).  The Jepsen paper was the first to really call out this atomic commit as not being a guaranteed thing that all databases offer, and it describes the Monotonic Atomic View (MAV) isolation level to cover it.  So read atomicity is part of the isolation level, rather than the Acidic 'A'.  We have this when two transactions interact, but not with N1QL, regular KV reads, etc.  They don't do the MAV logic.

Given that, hopefully it's clear that the issue with non-transactional updates doesn't affect either read or write atomicity.  The problem is that those updates can be lost, as the transaction will go ahead and overwrite them (once it's reached the commit point, it has to complete the commit).

modules/learn/pages/data/transactions.adoc
```diff
@@ -0,0 +1,56 @@
...
+
+* The number of writes required in a transaction is greater than the number required for non-transactional updates: writes are required both to _stage_ and to _commit_; and still further writes must be made in order to update transaction records.
```
programmatix, line 50: Sounds a touch negative.. maybe 'still further' -> 'additional', and 'must be made' -> 'are required'?

modules/learn/pages/data/transactions.adoc
```diff
@@ -0,0 +1,56 @@
...
+Each commit provides _eventual consistency_, which is consistent with the _single-document write_ model of durable writes; as described in xref:learn:data/durability.adoc[Durability].
+
+[#dependencies-and-limitations]
+=== Dependencies and Limitations
+
+_Distributed ACID Transactions_ constitute an evolving technology.
+Currently, the following dependencies and limitations apply:
+
+* Transactions require the Enterprise Edition 6.5 of Couchbase Server.
+They also require the Couchbase xref:3.0@java-sdk::start-using-sdk.adoc[Java SDK], version 3.0 or after.
```
programmatix, line 40: This is a big limitation, can we give an impression that other langs are on the horizon.

modules/learn/pages/data/validity.adoc
```diff
@@ -0,0 +1,70 @@
+= Validity Guarantees (ACID)
+
+[abstract]
+Couchbase Server validates data by means of _Atomicity_, _Consistency_, _Isolation_, and _Durability_.
```
programmatix, line 4: Is 'validates' the right term?  That brings JSON schema validation to mind, for me.

modules/learn/pages/data/validity.adoc
```diff
@@ -0,0 +1,70 @@
...
+
+The meanings of these properties, and the ways they are achieved by Couchbase Server, are described below.
+Note that various references are made to https://jepsen.io[Jepsen] &#8212; a project dedicated to improving the safety of distributed databases and other systems, with reference to specific https://jepsen.io/consistency[Consistency Models].
+
+[#atomicity-definition-and-implementation]
+=== Atomicity
+
+_Atomicity_ guarantees that when multiple data-elements need to be modified such that only the successful modification of all justifies the modification of any, _either_ all the modifications do occur successfully; _or_ none of them occurs.
+
+Atomicity is innately supported by the xref:learn:data/document-data-model.adoc[Couchbase Data Model]; which is naturally expressive of hierarchical data, and therefore typically only requires a single document for data that, under a relational model, would require multiple tables; and would therefore maintain data-elements discretely.
```
programmatix, line 20: So is this doc kinda explaining how ACID relates to both single doc operations, and multi-doc transactions?  I'm a touch confused by why this content is moved out of the transactions docs tbh - is it just to split a too-large doc?

modules/learn/pages/data/validity.adoc
```diff
@@ -0,0 +1,70 @@
...
+The _Database Change Protocol_ (DCP) keeps active and replica data-elements both consistent and up-to-date with _eventual consistency_.
+
+Additionally, Couchbase Server provides xref:learn:clusters-and-availability/xdcr-overview.adoc[Cross Data Center Replication (XDCR)]; which uses eventual consistency to replicate data-modifications across different clusters, potentially located in different data centers.
+
+The Couchbase xref:learn:services-and-indexes/services/index-service.adoc[Index Service] provides high-performance data-access; and as such, maintains fast-access copies of specified data.
+Such copies are updated with eventual consistency, across the cluster.
+A xref:n1ql:n1ql-rest-api/index.adoc#table_xmr_grl_lt[Read Your Own Writes] mechanism is available, to allow queries to read modified data from an index at the earliest point at which consistency is achieved.
+
+When data-elements must be updated consistently across multiple documents, xref:learn:data/transactions.adoc[Distributed ACID Transactions] can be used.
```
programmatix, line 40: I'd never seen a word reused for so many different purposes, until I came to work on a database and discovered "consistency"...

There's a lot of confusing elements in this section, I feel, as it's trying to define multiple sorts of consistency.  I don't think it's your fault, I think it's that we haven't defined super-well yet what consistency means for our transactions - in RMDBS, it traditionally referred to preserving things like foreign key constraints: things that don't apply to us.  Let me run this by Shivani.   

I think we should keep consistency strictly related to transactions here though.  This section is defining consistency for XDCR and indexing.

"When data-elements must be updated consistently across multiple documents," -> this is atomicity rather than consistency.

modules/learn/pages/data/validity.adoc
```diff
@@ -0,0 +1,70 @@
...
+
+[#isolation-definition-and-implementation]
+=== Isolation
+
+_Isolation_ ensures that no transaction's data-modifications are visible to any other transaction prior to commitment: a _transaction_ being an operation that performs an atomic update, involving multiple discrete data-elements.
```
programmatix, line 45: Weellll, _Read Committed_ Isolation means this..  but there are much higher isolation levels possible.  Serializable, the highest, means that you can create a theoretical history where every transaction happened in serial.  At its simplest isolation level really means, what's the behaviour of concurrent transactions.

modules/learn/pages/data/validity.adoc
```diff
@@ -0,0 +1,70 @@
...
+[#durability-definition-and-implementation]
+=== Durability
+
+_Durability_ guarantees that data written into a database is fully conserved; and is not lost even in the event of significant system-failure.
+
+Couchbase Server implements durability by means of _durable writes_ (sometimes referred to as _synchronous writes_).
+These may be appropriate when saving data whose loss could have a considerable, negative impact.
+For example, data corresponding to a financial transaction.
+A durable write must be completed on a majority of those cluster-nodes that bear the active and replica copies of the data, before the application is informed that the data has been committed.
+A successful commit may be considered to have occurred when the written data is either in the memory or on the disk of each of the nodes that constitute the majority.
```
programmatix, line 60: Maybe avoid the term commit here, given this is talking about single-docs and commit is an overloaded term for transactions.

programmatix (changes requested):

modules/learn/pages/data/transactions.adoc
tonyjhillman: Right. Well, it has been a strong principle that in this section of the docs, we don't do sales or marketing. We've worked hard to keep it to the technical details. That's the function of the section. And that's also the issue with addressing this content here.

modules/learn/pages/data/transactions.adoc
tonyjhillman: I'm attempting to define 'transactions' at the highest level (and all the terms involved with acid), and then go down to the couchbase specifics. So is there something better that is not couchbase-specific, referring to pieces of data in different locations?

modules/learn/pages/data/transactions.adoc
tonyjhillman: How about: 'For each document, the transaction makes changes in memory; then puts the modified copy of the document into the Extended Attributes area of the document's meta data. Consequently, the changes that occur during the transaction prior to commitment - these constituting 'uncommitted' or 'dirty' data - are non-readable by any operation other than this transaction itself.'

modules/learn/pages/data/transactions.adoc
tonyjhillman: 'None' is actually singular, though people often use plural.

modules/learn/pages/data/transactions.adoc
tonyjhillman: ok

modules/learn/pages/data/transactions.adoc
tonyjhillman: I based this on a sentence in the original, which said it doesn't affect search or n1ql. I can certainly remove if it's not required.

modules/learn/pages/data/transactions.adoc
tonyjhillman: I can just say: Only nodes that contain data to be updated are affected by a transaction. Does that work?

modules/learn/pages/data/transactions.adoc
tonyjhillman: How about: 'Multiple Distributed ACID Transactions can read the same document at the same time. If two transactions simultaneously attempt to write to a single document, the conflict is handled transparently; by allowing one to proceed, and retrying the other.' Something like that?

modules/learn/pages/data/transactions.adoc
tonyjhillman: ok

modules/learn/pages/data/transactions.adoc
tonyjhillman: Ok. I was referring to the transactions that use them. I can remove. But if I say that commits are performed by means of durable writes, which are single-document only, is that a problem?

modules/learn/pages/data/transactions.adoc
tonyjhillman: So, what do we say here? Do you want to change this? If I remove 'to ensure atomicity', is the rest of the sentence good?

modules/learn/pages/data/transactions.adoc
tonyjhillman: How about: 'The number of writes required in a transaction is greater than the number required for non-transactional updates: writes are required both to stage and to commit; and additional are required in order to update transaction records.'

modules/learn/pages/data/validity.adoc
tonyjhillman: Well, I based it on Wikipedia, which says that ACID is 'a set of properties of database transactions intended to guarantee validity even in the event of errors, power failures, etc'. We can change to something better, but I think we need to explain what binds these four properties together.

modules/learn/pages/data/validity.adoc
tonyjhillman: Well, I think if we do decide to define ACID, we need to do it for Couchbase Server and all its technologies. I think it will be perceived as odd if we describe some aspects, but not others. 

At any rate, the acronym includes Durability of course, and that is a separate page. So my sense is that the page logically needs to go on top, then transactions and durability beneath it.

tonyjhillman: Note also that atomicity is something we already ascribe to the standard single-document model, which is pre-transaction.

modules/learn/pages/data/validity.adoc
tonyjhillman: GSI, FTS indexes, and XDCR are each referred to in the original doc as representative of 'eventual consistency', so I put them here. I assumed that on that basis, DCP was a valid contender as well. 

Other than referring to indexes and RYOW, the original doc says: "Synchronous Replication, available from Couchbase Data Platform 6.5 onwards, ensures documents are synchronously replicated to a majority of replicas."  So that sounds like a duplication of durability placed underneath consistency, and I guess doesn't really specify transactions. So perhaps I'm not sure what sure should go here, transaction-wise.

modules/learn/pages/data/validity.adoc
tonyjhillman: So, is the behaviour 'levelled' in terms of how protected are one transaction's changes from those of another? How, I wonder, do we say this.

modules/learn/pages/data/validity.adoc
tonyjhillman: Yes. This is the language we've derived over time for the Durability page. I just copied it. I can probably just omit the last sentence for this summary.

PR #872, opened 2019-11-11
DOC-5812: change number of replicas with ALTER INDEX 

simon-dew:

* [ALTER INDEX](https://simon-dew.github.io/docs-site/DOC-5812/server/6.5/n1ql/n1ql-language-reference/alterindex.html) — described `replica_count` and `drop_replica` actions; new examples

Also minor changes to the following pages:

* [CREATE INDEX › Usage › Index Replicas](https://simon-dew.github.io/docs-site/DOC-5812/server/6.5/n1ql/n1ql-language-reference/createindex.html#index-replicas) — added ways to find the replica ID
* [Index Partitioning › Partition Placement](https://simon-dew.github.io/docs-site/DOC-5812/server/6.5/n1ql/n1ql-language-reference/index-partitioning.html#partition-placement) — information on changing the nodes for partition placement, moved here from ALTER INDEX 

Documentation issue [DOC-5812](https://issues.couchbase.com/browse/DOC-5812)


simon-dew:
Needs review from member of Index team

modules/n1ql/pages/n1ql-language-reference/alterindex.adoc
```diff
@@ -195,30 +240,45 @@ WITH {"action": "move", "nodes":
...
 
+[source,n1ql]
+----
 CREATE INDEX country_idx ON `travel-sample`(country, city)
-WHERE type="route" USING GSI
-WITH {"nodes": ["192.168.10.10:8091", "192.168.10.11:8091"]};
+WITH {"nodes": ["192.168.10.10:8091", "192.168.10.11:8091", "192.168.10.12:8091"]}
+
+ALTER INDEX `travel-sample`.country_idx
+WITH {"action": "replica_count", "num_replica": 1, "nodes": ["192.168.10.10:8091", "192.168.10.11:8091", "192.168.10.12:8091"]}
```
amithk: When decreasing the number of replicas, "nodes" parameter in "WITH" clause is unused. It is used only in case of Increasing number of replicas. So, to avoid confusion, we can remove this "nodes" parameter form this example, from the ALTER INDEX statement.

modules/n1ql/pages/n1ql-language-reference/alterindex.adoc
```diff
@@ -195,30 +240,45 @@ WITH {"action": "move", "nodes":
 ----
 ====
 
-.Removing an extra replica
+.Increasing the number of replicas
```
amithk, line 247: There are two possibilities. 
(1) If they want, user can specify the "nodes" parameter in the "WITH" clause in ALTER INDEX statement.
(2) If they don't want, user can choose not to specify any "nodes".

Both of these are valid scenarios for the user. 

When "nodes" is specified, the new replica will be placed on the nodes from the list specified by the user.
When "nodes" is not specified, the replica can be placed on any of the available indexer nodes in the cluster.

Note: In both of these cases, no single indexer node will host 2 replicas of the same index (or same partition of the same index). Also, replicas will get distributed across the available server groups.

I suggest we should have two examples for "Increasing the number of replicas", one without "nodes" parameter and one with "nodes" parameter.

simon-dew:
- [x] Partitioned indexes can have WHERE and USING clauses.

modules/n1ql/pages/n1ql-language-reference/alterindex.adoc
simon-dew: Done

modules/n1ql/pages/n1ql-language-reference/alterindex.adoc
simon-dew: Done

simon-dew:
[Additional comments](https://issues.couchbase.com/browse/DOC-5812?focusedCommentId=369751&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-369751)

PR #882, opened 2019-11-18
DOC-4241: GSI REST API for stats needs to be documented

simon-dew:

* [Index Statistics API](https://simon-dew.github.io/docs-site/DOC-4241/server/6.5/rest-api/rest-index-stats.html)

Documentation issue: [DOC-4241](https://issues.couchbase.com/browse/DOC-4241)


simon-dew:
See [cb-swagger #20](https://github.com/couchbaselabs/cb-swagger/pull/20) for OpenAPI spec

jamesnocentini (approved):

simon-dew:
Needs review by member of GSI team

akhilmd:
@simon-dew,
- In the examples, instead of the index `over1000`, is it possible use a sample index built as a part of loading travel-sample bucket?


- Reword definition for `items_count`
    - > Current total indexed document count
     
    Use 'item' instead of 'document' as the `items_count` stat may not map to number of documents indexed for array indexes.
    - > , refreshed on restart.
    
    This may not be worth mentioning as this stat is accurate even after restart. 

- Array index specific stats, `avg_array_length` and `docid_count` are not defined.
    - These are exposed only for array indexes.
    - docid_count: Number of documents that are currently indexed.
    - avg_array_length: Average number of items in each array.

simon-dew:
@akhilmd 

> In the examples, instead of the index `over1000`, is it possible use a sample index built as a part of loading travel-sample bucket?

Do any of the sample indexes have partitions?
I wanted to show the output for an index with partitions on a particular node.

simon-dew:
- [X] Add `avg_array_length`
- [X] Add `docid_count`
- [X] Update `data_size`
- [X] Update `items_count`

prathibha-cb:
@simon-dew :
1. When I run the same curl command as in the Example 1, I see 18 stats under index but in your example, I see only 10 stats. For the first example, it would be good to have all stats in the output as user will be able to get a grasp for all stats we provide. 
Actually, for both the below examples, we can output which is all-inclusive stats to it:
Example 1: 
curl -X GET -u Administrator:password "http://localhost:9102/api/v1/stats?skipEmpty=true&pretty=true"
Example 2:
curl -X GET -u Administrator:password "http://localhost:9102/api/v1/stats/travel-sample/over1000?partition=true&skipEmpty=true&pretty=true"

2. indexer_state - Schema can be updated for the sake of completeness to:
enum (Active, Pause, Warmup)

3. memory_used : Amount of memory used by the indexing service in the node (bytes)

simon-dew:
- [x] add a note to clarify that you can’t specify a partition in the path

PR #885, opened 2019-11-19
DOC-5812: change number of replicas with ALTER INDEX

simon-dew:
Additional changes following further review of #872


deepkaran (commented):
I think we can remove the references to "node list" from here to avoid any confusion. Indexer uses "node list" only on create index. 

"The indexing service remembers the number of partitions and replicas. Given sufficient capacity, the dropped replica is rebuilt after the next rebalance -- although it may be placed on a different index node, depending on the resource usage statistics of the available nodes."

simon-dew:
> remove the references to "node list" from here to avoid any confusion

Done

PR #889, opened 2019-11-21
DOC-4757: Explicitly state that Analytics upgrade from 5.x to 6.x is not supported

simon-dew:
- Add warning that datasets must be recreated

Documentation issue: [DOC-4757](https://issues.couchbase.com/browse/DOC-4757)


westmann (commented):
Should we add that the datasets must be removed before the upgrade and recreated after the upgrade?

simon-dew:
> Should we add that the datasets must be removed before the upgrade and recreated after the upgrade?

Good point. Updated

westmann (approved):

jamesnocentini (approved):

PR #899, opened 2019-11-25
Doc 6000

tonyjhillman:


modules/learn/pages/data/durability.adoc
```diff
@@ -81,7 +81,7 @@ Accordingly, it will be written to disk on those nodes.
 * _Timeout_.
 The number of milliseconds within which the durability requirements must be met.
 The SDK specifies a default of 2500.
-If the number is exceeded, the entire synchronous write is aborted.
+If this number is exceeded, the entire synchronous write is aborted.
```
owend74: I believe this is incorrect the synchronous write cannot be aborted (SDK does not have the ability to do this).  I think that it just that a timeout is returned to the end-user.  they don't know whether the write ultimately succeeds.
i.e. is essentially an ambiguous response.

owend74 (changes requested):

modules/learn/pages/data/durability.adoc
```diff
@@ -81,6 +81,7 @@ Accordingly, it will be written to disk on those nodes.
 * _Timeout_.
 The number of milliseconds within which the durability requirements must be met.
 The SDK specifies a default of 2500.
+If the number is exceeded, the entire synchronous write is aborted.
```
owend74: See comment on other commit - don't believe the synchronous write is aborted.  Might want to confirm with Graham.

owend74 (changes requested):

PR #918, opened 2019-12-02
DOC-5987: prepared statements calls USING syntax does not work

simon-dew:

* [EXECUTE › Parameters](https://github.com/couchbase/docs-server/blob/bf6217277388220bcfb33bb71463a7964fbcb43d/modules/n1ql/pages/n1ql-language-reference/execute.adoc#parameters)
* [N1QL Error Codes › 5xxx Codes (exec)](https://github.com/couchbase/docs-server/blob/bf6217277388220bcfb33bb71463a7964fbcb43d/modules/n1ql/pages/n1ql-language-reference/n1ql-error-codes.adoc#5xxx-codes-exec)

Docs issue: [DOC-5987](https://issues.couchbase.com/browse/DOC-5987)

### DIFF
--- a/modules/n1ql/pages/n1ql-language-reference/execute.adoc
+++ b/modules/n1ql/pages/n1ql-language-reference/execute.adoc
@@ -29,7 +29,7 @@ TIP: If the name of the prepared statement contains hyphens, wrap the entire nam
 
 === USING Clause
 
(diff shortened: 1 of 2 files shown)


sitaramv:
Not only rest API, or cbq shell. Even clients can't set this.  Say only one (USING clause or named/positional parameters and not both) is allowed

PR #935, opened 2019-12-06
terminology improvement

simon-dew:
https://tools.ietf.org/id/draft-knodel-terminology-00.html#master-slave


RichardSmedley (approved):
LGTM
Thanks Simon :+1:

PR #954, opened 2019-12-30
Corrected Line 5

Mohit-Nain:
"This helps ensure" does not read perfectly, the change has been proposed to overcome this.


amarantha-k:
With the verb help, the word "to" is optional. "helps ensure" and "helps to ensure" are both correct.

PR #956, opened 2020-01-02
Line 8 update

Mohit-Nain:
Since we are taking about multiple operations , hence data mutation term should be used in place of "a data mutation"


Mohit-Nain:
@RichardSmedley can you please have a look at this open PR

RichardSmedley (approved):
:+1:

PR #966, opened 2020-01-07
DOC-4917: Fix formatting in query documentation

simon-dew:
Port #965 to mad-hatter


simon-dew:
- [x] Remove FTS examples from CURL() function, replaced with SEARCH()

PR #1010, opened 2020-02-03
DOC-5993a: Reference page on data type conversion for xdcr

tonyjhillman:


modules/xdcr-reference/pages/xdcr-filtering-data-type-conversion.adoc
```diff
@@ -0,0 +1,479 @@
...
+[cols="1,4"]
+|===
+| *Precedence* | *Data Type*
+| 0 | Invalid
+| 1 | Missing
+| 2 | Null
+| 3 | Boolean
+| 4 | Numeric
+| 5 | String
+| 6 | Time
```
nelio2k, line 32: (Optional) Wording is not exact so can consider ways to link to existing documentation link for clarification - https://docs.couchbase.com/server/current/xdcr-reference/xdcr-filtering-expressions.html#handling-dates
Not sure what is the optimal way to do it here...

modules/xdcr-reference/pages/xdcr-filtering-data-type-conversion.adoc
```diff
@@ -0,0 +1,479 @@
...
+
+* _Regular expression_ matches.
+Checking for a regular-expression _match_ between different data types returns `false`.
+Therefore, if `var1` and `var2` represent different data types, the expression `REGEXP_CONTAINS(var1, var2)` returns `false`.
++
+Checking for a regular-expression _non-match_ between different data types returns `true`.
+Therefore, if `var1` and `var2` represent different data types, the expression `NOT REGEXP_CONTAINS(var1, var2)` returns `true`.
+
+* _Magnitude comparison_ between different data types is resolved in accordance with the respective positions of the data types in the list of xref:xdcr-reference:xdcr-filtering-data-type-conversion.adoc#supported-data-types[Supported Data Types]).
+This list presents the supported data types in order of _precedence_: note that _precedence_ and _magnitude_ are inversely proportional; and that therefore, the data type at the bottom of the list (Binary) has the least precedence and greatest magnitude; while the data type at the top (Invalid) has the greatest precedence and least magnitude.
```
nelio2k, line 119: I like the way you describe it here. It's very clear and easy to understand the way it is phrased.

modules/xdcr-reference/pages/xdcr-filtering-data-type-conversion.adoc
```diff
@@ -0,0 +1,479 @@
...
+| `Time`
+| image:xdcr-reference:checkmark-orange-sm.png[] Valid if `String` can be parsed as a parameter to the `DATE` function.
+| Standard comparision is performed if valid; otherwise, collation comparison is performed.
+
+| 25
+| `Array`
+| `Array`
+| image:xdcr-reference:dash-black-inline.png[] No conversion need be performed.
+| Comparison is performed along the following lines: if the length of `Array_1` is greater than the length of `Array_2`, then `Array_1 > Array_2` returns `true`.
+If the arrays are of equal length, corresponding objects are compared in sequence, until one is found to be the greater, and `true` is returned &#8212; _i.e._ `Array_1[object_1] > Array_2[object_1]`, `Array_1[object_2] > Array_2[object_2]`, and so forth.
```
nelio2k: > until one is found to be the greater, and `true` is returned

The above reads as if "greater" and "true" are the operations being used internally, which may not be generic enough.

The idea to convey here should be objects within the arrays are compared in sequence against the expression set forth by the user if they are of the same length. What if the user had the expression `arr1 != arr2 == true` and the arrays are of the same length and only the last object of both arrays are different? The algorithm here is to compare objects by object until it figures out arr1 is > or < than arr2 due to the last element within the array, and then apply the operation "!=" to the result, which is true.

modules/xdcr-reference/pages/xdcr-filtering-data-type-conversion.adoc
```diff
@@ -0,0 +1,479 @@
...
+
+| 26
+| `Object`
+| `Object`
+| image:xdcr-reference:dash-black-inline.png[] No conversion need be performed.
+| Comparison is performed along the following lines: if the length of `Object_1` is greater than the length of `Object_2`, then `Object_1 > Object_2` returns `true`.
+If the objects are of equal length, corresponding bytes of the objects' data are compared in sequence, until one is found to be the greater, and `true` is returned &#8212; _i.e._ `Object_1[byte_1] > Object_2[byte_1]`, `Object_1[byte_2] > Object_2[byte_2]`, and so forth.
```
nelio2k: Same here as above.

modules/xdcr-reference/pages/xdcr-filtering-data-type-conversion.adoc
```diff
@@ -0,0 +1,479 @@
...
+
+| Any numeric values without precision delimiter or mantissa.
+| `Int`
+| `variable > 1234`
+
+`variable < -2345`
+
+| Any number representing a valid `golang` float, optionally with precision delimiter and/or mantissa.
+| `Float`
+| `variable >= 1.2343+25`
```
nelio2k: > 1.2343+25

Should be 1.2343**e**+25
https://stackoverflow.com/questions/19554972/json-standard-floating-point-numbers/19554986#19554986

modules/xdcr-reference/pages/xdcr-filtering-data-type-conversion.adoc
tonyjhillman: Would the following be better? 

Comparison is performed along the following lines: the specified expression is first applied to the arrays themselves, based on the respective array-lengths.
For example, if the expression is `arg_1 > arg_2`, this gets applied as `Array_1 > Array_2`; and if the length of `Array_1` is indeed greater than the length of `Array_2`, the requirement is considered to have been met; and the comparison-process ends.

In cases where the requirement is _not_ met based on array-lengths, due to the array-lengths being _equal_, the comparison-process continues as follows: in sequence, pairs of correspondingly positioned objects from the arrays are compared, until the specified requirement is met, or is finally determined _not_ to have been met.
For example, if the expression is `arg_1 > arg_2`, this gets applied as `Array_1[object_1] > Array_2[object_1]`, `Array_1[object_2] > Array_2[object_2]`, and so forth.

Question: So, if the condition not met because Array_1 < Array_2, does that mean the result is false, and comparison ends at that point?

nelio2k: > For example, if the expression is arg_1 > arg_2, this gets applied as Array_1[object_1] > Array_2[object_1], Array_1[object_2] > Array_2[object_2], and so forth.

Arrays are indexed by positive integers and not a specific object so it may be more technically correct to say "Array_1[0] > Array_2[0]". 
I understand the concept being conveyed but it looked kinda odd from a coding standpoint.

> Question: So, if the condition not met because Array_1 < Array_2, does that mean the result is false, and comparison ends at that point?

I'm assuming you're asking 

> if the condition is not met because the *length* of Array_1 is smaller than the length of Array_2 when the expression given is 'Array_1 > Array_2'

Then yes, the result returned is "false" and the object-by-object comparison is not executed as it isn't needed.

modules/xdcr-reference/pages/xdcr-filtering-data-type-conversion.adoc
tonyjhillman: I can point to that document for further info.

modules/xdcr-reference/pages/xdcr-filtering-data-type-conversion.adoc
tonyjhillman: Great. Thanks.

modules/xdcr-reference/pages/xdcr-filtering-data-type-conversion.adoc
tonyjhillman: Ok. Thanks. Will change to [0] etc.

modules/xdcr-reference/pages/xdcr-filtering-data-type-conversion.adoc
tonyjhillman: See the solution I've attempted. I've broken the explanations out as notes beneath the table.

modules/xdcr-reference/pages/xdcr-filtering-data-type-conversion.adoc
tonyjhillman: Ok. Done.

nelio2k (approved):

amarantha-k (approved):

PR #1019, opened 2020-02-07
Corrected typo: 762 was written as 62. fixed that

gurmeetsinghdke:
762 was written as 62. So corrected that type.


tonyjhillman (approved):
thanks!

PR #1062, opened 2020-02-21
Doc 6196 6092 cbstats refresh

jamesneimanconsulting:
Covers this request from @amarantha-k: 
DOC-6196: cbstats examples need to include authentication.
DOC-6092: Need to make it clear that cbstats is Data Service only.
Additionally, all the cbstats pages (all pages under cbstats) need to be refreshed, with fresh output being generated, and outdated options being found and removed.


tonyjhillman (changes requested):
Thanks for handling. I've added some info to DOC-6092 with some requests for further changes, mainly on the intro page: some of the information is badly out of date, so we should at least remove.

tonyjhillman:
I've handled this under https://github.com/couchbase/docs-server/pull/1108

PR #1063, opened 2020-02-24
DOC-4558: Added docvalues to child field GUI

tonyjhillman:


modules/fts/pages/fts-creating-indexes.adoc
```diff
@@ -188,6 +188,10 @@ When unchecked, term vectors are not included.
 Term vectors are the locations of terms in a particular field.
 Certain kinds of functionality (such as highlighting, and phrase search) require term vectors.
 Inclusion of term vectors results in larger indexes and correspondingly slower index build-times.
+* [.ui]*docvalues*: When checked, the _value_ for each instance of the field is itself included in the index.
+This optimises (but it not essential for) xref:fts:fts-response-object-schema.adoc[Handling Response Objects] and the use of xref:fts:fts-response-object-schema.adoc#search-facets[Search Facets].
```
sreekanth-cb: Enabling docValues is essential for the working of facets query on those fields .
Also, if the user wishes to perform sorting of search results based on field values, then those fields need to be indexed with docValues enabled.

sreekanth-cb (changes requested):

modules/fts/pages/fts-creating-indexes.adoc
tonyjhillman: Have modified. So, I've said that if this is not checked, Facets and value-sorting are not supported. Is that accurate? If so, maybe we need pointers back to this page from those that discuss Facets and value-sorting.

sreekanth-cb (approved):
This looks better, not sure whether the existing documentation about facets or sorting of fields mentions about doc values for adding that extra links.

amarantha-k (approved):

PR #1068, opened 2020-02-25
Doc 4840

tonyjhillman:


eric-schneider (approved):
lgtm - though, I've been informed that the underlying source code in [kv_engine](https://github.com/couchbase/kv_engine) refers to this number as `revSeqno`. So I guess we just can't win 😆. But I think we're probably doing the right thing by going with what's in the UI, for consistency's sake.

tonyjhillman:
Great, thanks!

PR #1077, opened 2020-02-25
Doc 6174

jamesneimanconsulting:
@amarantha-k These updates regarding the REST API are submitted per the JIRA ticket description.


tonyjhillman (commented):
Some more work, I think, is required here. I believe the endpoint should be /sampleBuckets/install - but I'm having a hard time getting it to work.

tonyjhillman:
Okay, so there are two endpoints to be documented here I think. The first is the one that retrieves the available sample buckets:

$ curl -X GET -u Administrator:password http://10.143.194.101:8091/sampleBuckets/
[{"name":"beer-sample","installed":true,"quotaNeeded":104857600},{"name":"gamesim-sample","installed":false,"quotaNeeded":104857600},{"name":"travel-sample","installed":true,"quotaNeeded":104857600}]

The second is the one that installs sample buckets from a specified list:

curl -X POST -u Administrator:password http://localhost:8091/sampleBuckets/install -d '["travel-sample", "beer-sample"]'

tonyjhillman:
I've handled this under https://github.com/couchbase/docs-server/pull/1100

PR #1085, opened 2020-02-26
Eventing port name is wrong

jon-strabala:
#1 Note "debugPort" is wrong it will not work the correct name is "eventing_debug_port"

#2 The example is wrong it must be terminated by a period for example if we had the following in static_config

{eventing_debug_port, 9444}
{net_kernel_verbosity, 10}

The couchbase server WILL NOT start but if we add a period after the closing brace

{eventing_debug_port, 9444}.
{net_kernel_verbosity, 10}.

The couchbase server starts

#3 make the syntax clear e.g. add "(enclosed in braces and terminated by a period)"

#4 warn that this will flush all data and restart the server from scratch, e.g. "WARNING: Changing port mapping should be done on the initial node/cluster setup as the required reset and reconfiguration will also purge any data on the node. "

### DIFF
--- a/modules/install/pages/install-ports.adoc
+++ b/modules/install/pages/install-ports.adoc
@@ -1,5 +1,5 @@
 = Couchbase Server Ports
-:fn-eventing-debug-port: pass:n[footnote:fn-eventing-debug-port[pass:n[The Eventing Service Debugger port (9140, `debugPort`) is an internal port and is not supported for external access outside of the cluster. You should only use this port in your development environments.\]]]
+:fn-eventing-debug-port: pass:n[footnote:fn-eventing-debug-port[pass:n[The Eventing Service Debugger port `eventing_debug_port` (9140) is an internal port and is not supported for external access outside of the cluster. You should only use this port in your development environments.\]]]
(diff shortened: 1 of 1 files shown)


eric-schneider:
Thanks for creating this PR @jon-strabala. Can you confirm which releases, if any, that these changes should be backported to? Or are they only applicable to 6.5?

jon-strabala:
Eric,

1) I tested the eventing port name and the "failure case" on a CC 7.0.0 source build.

Since you asked I just did the following:

2) I tested the eventing port name and the "failure case" on a CC 6.5.1 source build.

But since I run debian 9 I can't easily run the 6.0.X or prior releases since they are not built on http://nas.service.couchbase.com/builds/releases

Best


Jon Strabala


________________________________

jon-strabala:
Eric,

3) I tested the eventing port name and the "failure case" on Enterprise Edition 6.0.3 build 2895

    couchbase-server-enterprise_6.0.3-debian8_amd64.deb

4) I tested the "failure case" on e.g.missing period on
    couchbase-server-enterprise_5.1.0-debian8_amd64.deb

- Jon
________________________________

amarantha-k (approved):

PR #1105, opened 2020-03-04
Update N1QL+FTS search example to use one of the demonstration indexes

abhinavdangeti:


abhinavdangeti:
Ping @amarantha-k and @simon-dew for review here.

modules/fts/pages/fts-demonstration-indexes.adoc
```diff
@@ -112,14 +113,15 @@ This page also provides a definition for the `geoIndex` definition, used in the
...
     "numReplicas": 0
   },
   "uuid": ""
 }
 ----
 
-[#travel-sample-index-hotel-description]
-== Index Definition: travel-sample-index-hotel-description
+[#travel-sample-index-hotel-country-description]
+== Index Definition: travel-sample-index-hotel-country-description
```
simon-dew: These changes will have an effect on the SDK docs, for example [Searching from the SDK](https://docs.couchbase.com/java-sdk/2.7/full-text-searching-with-sdk.html). Are there separate PRs to update all the SDK docs?

If none of the provided demonstration indexes are good for the new SEARCH() function example, it would be better to add an entirely new demonstration index for the SEARCH() function example, and leave travel-sample-index-hotel-description unchanged.

modules/n1ql/pages/n1ql-language-reference/searchfun.adoc
```diff
@@ -310,33 +310,33 @@ This query returns 5 results, and the results are ordered, as specified by the s
...
   {
-    "id": "airline_2657"
+    "id": "hotel_37597"
   },
 ...
 ]
 ----
 
-If the full text search index being queried has its default mapping disabled and has a custom type mapping defined ("airline" in the above example), the query needs to specify the type explicitly.
+If the full text search index being queried has its default mapping disabled and has a custom type mapping defined ("hotel" in the above example), the query needs to specify the type explicitly. For a sample index definition for the above query, refer to xref:fts-demonstration-indexes.adoc#travel-sample-index-hotel-description-country[Travel sample index definition over documents of type: "hotel"].
```
simon-dew:
```suggestion
If the full text search index being queried has its default mapping disabled and has a custom type mapping defined ("hotel" in the above example), the query needs to specify the type explicitly.
For a sample index definition for the above query, refer to xref:fts-demonstration-indexes.adoc#travel-sample-index-hotel-country-description[Travel sample index definition over documents of type: "hotel"].
```

simon-dew: New line for new sentence, and section ID did not match the section ID in `fts-demonstration-indexes.adoc`.

modules/n1ql/pages/n1ql-language-reference/searchfun.adoc
```diff
@@ -310,33 +310,33 @@ This query returns 5 results, and the results are ordered, as specified by the s
 As an alternative, you could limit the number of results and order them using the N1QL xref:n1ql-language-reference/limit.adoc[LIMIT] and xref:n1ql-language-reference/orderby.adoc[ORDER BY] clauses.
 ====
 
-.Search against a full text search index that carries a custom type mapping
+.Search against a full text search index that carries a custom type mapping on documents with "type" field: "hotel"
```
simon-dew: Don't need to extend the title, it's long enough as it is and the description below explains the details of the custom type mapping.

simon-dew (changes requested):
Thanks for doing this change @abhinavdangeti , it's great to make the examples reproducible for the users. I'm concerned that the suggested changes in this PR will have side effects on the SDK documentation, and I think that needs to be addressed. See individual comments.

modules/fts/pages/fts-demonstration-indexes.adoc
abhinavdangeti: Thanks for pointing that it out, I'll update the SDK documentation, there's a custom analzyer defined in that index def which isn't even picked up, so it's useless which i've dropped.

modules/n1ql/pages/n1ql-language-reference/searchfun.adoc
abhinavdangeti: Done.

modules/fts/pages/fts-demonstration-indexes.adoc
abhinavdangeti: On second thought, let me not update the sdk docs, but adapt the example i've used here.

modules/n1ql/pages/n1ql-language-reference/searchfun.adoc
abhinavdangeti: Done.

abhinavdangeti:
@simon-dew ping for re-review.

modules/n1ql/pages/n1ql-language-reference/searchfun.adoc
```diff
@@ -316,27 +316,30 @@ As an alternative, you could limit the number of results and order them using th
...
-...
+  {
+    "id": "hotel_20422"
+  }
 ]
 ----
 
-If the full text search index being queried has its default mapping disabled and has a custom type mapping defined ("airline" in the above example), the query needs to specify the type explicitly.
+If the full text search index being queried has its default mapping disabled and has a custom type mapping defined ("hotel" in the above example), the query needs to specify the type explicitly.
+For a sample index definition for the above query, refer to xref:fts-demonstration-indexes.adoc#travel-sample-index-hotel-description[Travel sample index definition over documents of type: "hotel"].
```
simon-dew:
```suggestion
If the full text search index being queried has its default mapping disabled and has a custom type mapping defined, the query needs to specify the type explicitly.
The above query uses the demonstration index xref:fts:fts-demonstration-indexes.adoc#travel-sample-index-hotel-description[travel-sample-index-hotel-description], which has the custom type mapping "hotel".
```

simon-dew: Correct xref as it's refering to a different module. Also suggested some minor wording changes.

simon-dew (changes requested):
That's brilliant, thanks. Just need to correct an xref so it refers to the right module. I've also made some minor wording suggestions to make it clear why the query contains `t1.type = "hotel"`.

The updates for the demonstration indexes are fine — in fact I found that Couchbase Server 6.5 imports them as Scorch indexes anyway. I'll let @RichardSmedley know that the definitions of the demonstration indexes have changed and he might need to check the code samples for SDK, but it shouldn't be a problem, I hope.

amarantha-k:
@simon-dew @abhinavdangeti Just checking if we're expecting any more changes on this PR.

abhinavdangeti:
Looks all set to me!

simon-dew (approved):

simon-dew:
Sorry for the delay! Pls go ahead and merge

PR #1106, opened 2020-03-05
Typo fix

arshita-kakkar:
Updated "temporary named" to "temporarily named"


simon-dew (commented):
I think this was correct as it was — it's a "[temporary [named result set]]", not a "[[temporarily named] result set]". Perhaps change to something like "ad hoc named result set" or "non-permanent named result set" to avoid the ambiguity.

simon-dew (approved):
Sorry for the delay. LGTM 👍

PR #1107, opened 2020-03-05
DOC-6262: Added limitations section

tonyjhillman:


tonyjhillman:
So, I've tried to simplify and reduce this a little, since I find the context into which we are placing it already provides a lot. I've also tried to clarify this issue that apparently related to default versus named collections - though I'm not absolutely sure I'm interpreting it accurately. Anyway, see what you think.

tonyjhillman:
One issue I've not tried to tackle is that of our 'recommendation' that a new bucket be created. I'm not sure exactly why it's a recommendation, and I'm thinking that if we indeed make it, we probably should be very explicit as to why.

kaminijagtiani (changes requested):
IMO We should explicity add:
If a user wants to try out KV collections feature,customer should turn DP mode on, create a new bucket and then create Scope and collections and use only KV on this bucket for 6.5 server.
Support of N1QL and indexing for collections will be added in the next major release

amarantha-k:
@tonyjhillman Could you add an explicit note based on the info provided by Kamini? Thanks.

tonyjhillman:
Current version reviewed and approved by Jim Walker.

amarantha-k (approved):

PR #1110, opened 2020-03-06
DOC-6401for6.0: Modified Data Service addition text

tonyjhillman:


modules/manage/pages/manage-nodes/create-cluster.adoc
```diff
@@ -88,7 +88,7 @@ Your email-address will be added to the Couchbase community mailing-list, so tha
 
 You now have two options for proceeding.
 If you left-click on the [.ui]*Finish With Defaults* button, cluster-initialization is performed with default settings, provided by Couchbase; the Couchbase Web Console [.ui]*Dashboard* appears, and your configuration is complete.
-However, if you wish to customize those settings, left-click on the btn:[Configure Disk, Memory, Services] button, and proceed as follows.
+However, if you wish to customize those settings, left-click on the *Configure Disk, Memory, Services* button, and proceed as follows.
```
amarantha-k: We discussed the use of btn: macro in the Style & Tone meeting and decided that the output needs to be updated. I would suggest reverting this change so the page is consistent and we don't have to go back and revert this change later.

amarantha-k (approved):

PR #1118, opened 2020-03-09
DOC-6409: Updated checkpoint and config commands for cbstats for 6.5

tonyjhillman:


modules/cli/pages/cbstats/cbstats-config.adoc
```diff
@@ -114,11 +117,27 @@ ep_item_freq_decayer_chunk_duration:                   20
 ep_item_freq_decayer_percent:                          50
 ep_item_num_based_new_chk:                             true
 ep_keep_closed_chks:                                   false
+ep_magma_commit_point_every_batch:                     false
```
amarantha-k, line 121: Assuming this was run on a EE setup, could we add a note as CE users won't see "ep_magma_*"?

amarantha-k (approved):
Approving with request to add a clarifying note for "ep_magma*"

PR #1121, opened 2020-03-10
DOC-5601: Update N1FTY document with indexUUID option

simon-dew:

* [Search Functions › SEARCH() › Arguments](https://github.com/couchbase/docs-server/pull/1121/files#diff-ca957de7fb542ab484147334ab10f381)

Docs issue: [DOC-5601](https://issues.couchbase.com/browse/DOC-5601)

### DIFF
--- a/modules/n1ql/pages/n1ql-language-reference/searchfun.adoc
+++ b/modules/n1ql/pages/n1ql-language-reference/searchfun.adoc
@@ -143,9 +143,8 @@ For more information, refer to xref:fts:fts-creating-indexes.adoc[Creating Index
 This provides an index hint to the Query engine.
 If the full text index cannot be identified, an error occurs.
 
-You can use the `indexUUID` field as an alternative to the `index` field.
-You can also specify both the `indexUUID` field and the `index` field together.
-If you do this, the `indexUUID` field and the `index` field must both identify the same full text index.
+You can use the `indexUUID` field alongside the `index` field to help identify a full text index.
+The `indexUUID` field and the `index` field must both identify the same full text index.
 If they identify different full text indexes, or if either of them does not identify a full text index, an error occurs.
 
 You can find the UUID of a full text index by viewing the index definition.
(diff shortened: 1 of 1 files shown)


simon-dew:
DOC-5602 will be implemented by a separate pull request.

modules/n1ql/pages/n1ql-language-reference/searchfun.adoc
```diff
@@ -136,6 +136,21 @@ It overrides the default mapping and is used during index selection and filterin
...
 For more information, refer to xref:fts:fts-creating-indexes.adoc[Creating Indexes].
 
+| `indexUUID`
+[Optional]
+| string
+| A string, containing the UUID of a full text index in the keyspace.
+This provides an index hint to the Query engine.
+If the full text index cannot be identified, an error occurs.
+
+You can use the `indexUUID` field as an alternative to the `index` field.
```
abhinavdangeti: This isn't accurate. `indexUUID` isn't an alternative to the `index` field, but can be passed in alongside the `index` field to further assert that you're executing the query on the very index. It is however optional.

abhinavdangeti (changes requested):

modules/n1ql/pages/n1ql-language-reference/searchfun.adoc
simon-dew: Sorry @abhinavdangeti I don't quite understand this comment, given that you can use the `indexUUID` field without the `index` field. For example:

```
SELECT META(t1).id
FROM `travel-sample` AS t1
WHERE SEARCH(t1.country, "+United +States", {"indexUUID": "472b0cad1f126fe9"});
```

abhinavdangeti: This is a covering index situation. The verification phase doesn't even run in this case.

In case of a non-covering index situation (for example query for something from the object body), the verification phase will pick up the index mapping of an FTS index - if the index name is provided. To make this further accurate, if an indexUUID is also provided, the UUID of the fetched index mapping is also verified.

simon-dew: Ah, understood. Thanks.

abhinavdangeti (approved):

RichardSmedley (approved):
LGTM

PR #1130, opened 2020-03-13
Add best practices concerning residency ratio

jon-strabala:
Discuss that metadata bucket be 100% memory resident

Explain ETMPFAIL failures on high velocity Eventing is due to lack of resources and how to solve

### DIFF
--- a/modules/eventing/pages/troubleshooting-best-practices.adoc
+++ b/modules/eventing/pages/troubleshooting-best-practices.adoc
@@ -158,4 +158,4 @@ There are two solutions:
 
 . The first solution is to increase the Memory Quota of the bucket in question (thus increasing the resident ration).
 
-. The second solution is adding more Data nodes, faster disk IO, more meemory to eliminate the resource bottleneck.
+. The second solution is adding more Data nodes, faster disk IO, more memory to eliminate the resource bottleneck.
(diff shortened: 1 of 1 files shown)


modules/eventing/pages/troubleshooting-best-practices.adoc
```diff
@@ -131,3 +131,31 @@ Ensure that you configure the script timeout value after carefully evaluating th
...
+"stack":"Error\n    at OnUpdate (MyEventingFunction.js:177:25)"}
+```
+
+The above error indicates that the system is under provisioned for the load.  Under the hood Eventing will try to access to the data store five (5) times with a 200ms pause between attempts. If all of the attempts fail, the handler, in this case _MyEventingFunction_, throws an *ETMPFAIL* message. This is important to understand as trapping the above exception and retrying the same operation inside your handler will only exacerbate the issue and make things worse.  Of course your handler can take other actions such as creating a notification.
+
+There are two solutions: 
+
+. The first solution is to increase the Memory Quota of the bucket in question (thus increasing the resident ration).
+
+. The second solution is to reduce vBuckets per node by adding more Data nodes.
```
ingenthr: Unless there is something eventing specific, number of vbuckets per node doesn't matter so much.  Faster disk IO, more memory, whether done through more nodes or modification of node resources is what matters.

jon-strabala: Thank Matt - corrected.

amarantha-k (approved):

PR #1165, opened 2020-03-27
Update deployment-considerations-lt-3nodes.adoc

BKaneAtWork:
one small change in the Two-Node Deployments section: correcting "has less than two nodes" to "has less than three nodes"


RichardSmedley (approved):
:+1:

PR #1177, opened 2020-04-01
DOC-6333for6.0: Added $ character to regular expressions for xdcr

tonyjhillman:


modules/learn/pages/clusters-and-availability/xdcr-filtering.adoc
```diff
@@ -50,14 +50,17 @@ You can enclose a range of characters in square brackets, to match against all o
 | Matches any alphanumeric character.
 
 | `^`
-| Matches beginning of input.
-If the multi-line flag is set to `true`, also matches immediately after a line-break character.
-For example, `/^A/` does not match the '[.code]``A``' in "[.code]``an A``", but does match the '[.code]``A``' in "[.code]``An E``".
+| Matches the beginning of input.
+For example, `/^a/` matches `abb` and `ab`; but does _not_ match `ba`, `bab`, or `bba`.
+If the multi-line flag is set to `true`, `^` matches the start of _any line_ within the input string, instead of the beginning of the entire string.
```
JFlath, line 55: I realise that it's not the focus of the PR, but is this correct?  Not sure a multi-line flag applies in this context... perhaps something that was carried over from docs on the regex lib?

modules/learn/pages/clusters-and-availability/xdcr-filtering.adoc
```diff
@@ -50,14 +50,17 @@ You can enclose a range of characters in square brackets, to match against all o
 | Matches any alphanumeric character.
 
 | `^`
-| Matches beginning of input.
-If the multi-line flag is set to `true`, also matches immediately after a line-break character.
-For example, `/^A/` does not match the '[.code]``A``' in "[.code]``an A``", but does match the '[.code]``A``' in "[.code]``An E``".
+| Matches the beginning of input.
+For example, `/^a/` matches `abb` and `ab`; but does _not_ match `ba`, `bab`, or `bba`.
```
JFlath, line 54: Not the focus of the PR, but surrounding the regex string with slashes seems odd here. It's a convention for "this is a regex", but *not* something you do when you enter it in the product, so might not make sense to add them here?

JFlath: (I do really like the change of examples here btw!)

modules/learn/pages/clusters-and-availability/xdcr-filtering.adoc
```diff
@@ -50,14 +50,17 @@ You can enclose a range of characters in square brackets, to match against all o
...
+If the multi-line flag is set to `true`, `^` matches the start of _any line_ within the input string, instead of the beginning of the entire string.
 
-The '[.code]``^``' has a different meaning when it appears as the first character in a character set pattern.
-See https://developer.mozilla.org/en/docs/Web/JavaScript/Guide/Regular_Expressions#special-negated-character-set[complemented character sets^] for details and an example.
+Used within square brackets, as the first character, `^` implies _negation_: therefore, `[^0-9]` matches against any character that is not a digit.
+
+| `$`
+| Matches the end of input.
+For example, `/$a/` matches `ba` and `bba`; but does _not_ match `abb`, `bab`, or `ab`.
```
JFlath, line 61: I haven't explicitly tested, but the general pattern is that the `$` represents the end of the line itself, rather than being a marker for "this is a string at the end of the line". E.g., I think this should be:

```
For example, `a$` matches `ba` and `bba`; but does _not_ match `abb`, `bab`, or `ab`.
```

See above for slashes around regexes.

modules/learn/pages/clusters-and-availability/xdcr-filtering.adoc
```diff
@@ -50,14 +50,17 @@ You can enclose a range of characters in square brackets, to match against all o
...
+If the multi-line flag is set to `true`, `$` matches the end of _any line_ within the input string, instead of the end of the entire string.
```
JFlath, line 62: See above for multi-line.

JFlath (changes requested):

modules/learn/pages/clusters-and-availability/xdcr-filtering.adoc
tonyjhillman: Yes, could well be. I have removed.

modules/learn/pages/clusters-and-availability/xdcr-filtering.adoc
tonyjhillman: Okay, I have removed the slashes.

modules/learn/pages/clusters-and-availability/xdcr-filtering.adoc
tonyjhillman: Thanks. I have changed accordingly.

modules/learn/pages/clusters-and-availability/xdcr-filtering.adoc
tonyjhillman: Have removed.

PR #1208, opened 2020-04-13
DOC-6469 Add known issue and update upgrade doc

amarantha-k:


modules/install/pages/upgrade.adoc
```diff
@@ -40,7 +40,7 @@ That means, if you're more than one major version behind, you'll need to upgrade
 | 5.x
 | Enterprise Edition
 | Offline or Online
-| Direct upgrade to 6.5
+| 5.x -> 6.0.4+ -> 6.5
```
eric-schneider: Based on the release note text, I would think that 5.5.1-5.5.6 would work.

```suggestion
| 5.5.0 -> 6.0.4+ -> 6.5

5.5.1+ -> 6.5
```

eric-schneider (approved):
lgtm

malarky:
From the issues in Jira, it looks like 5.5.0 and 5.5.1 are the same - it's release *prior* to 5.5.0 (e.g. all 5.0.x/5.1.x) that are problematic.

saigon:
Only cluster prior version 5.5.0 (like 5.0.x or 5.1.x) which could not directly upgrade to 6.5.0.
Cluster with version 5.5.x is ok to upgrade directly to 6.5.0

amarantha-k:
Thanks @malarky and @saigon for the review. Updated the text to incorporate your feedback.

PR #1215, opened 2020-04-14
MB-37643: Added TTL issue to known issues

owend74:
Also stated that fixed in 6.0.4 and removed "Couchbase Server versions 5.5.x and 6.0.0 - 6.0.3 are affected by this issue." from the 6.0.4 release notes.


RichardSmedley (commented):
Thanks @owendCB 
I think a different approach is going to be taken with _Known Issues_ in patch releases...
I'll let @amarantha-k comment.

amarantha-k (approved):

PR #1220, opened 2020-04-14
DOC-6601 asp.net core quick start tutorial, images, sample project

mgroves:
Here's my first draft of the ASP.NET Core quickstart.


modules/getting-started/pages/quickstart-dotnet27-aspnetcore31-visualstudio-first-query-cb65.adoc
```diff
@@ -0,0 +1,154 @@
...
+
+[source,JavaScript]
+----
+{
+	"Date" : "2020-04-14",
+	"TemperatureC" : 25,
+	"Summary" : "Today's weather is partly cloudy and warm."
+}
+----
+
```
blbryant, line 102: An sample curl statement would be good here, along with a sample response.

modules/getting-started/pages/quickstart-dotnet27-aspnetcore31-visualstudio-first-query-cb65.adoc
```diff
@@ -0,0 +1,154 @@
...
+image::bucketOneDocument.png[Template select in Visual Studio]
+
+== Step 4: Query the document
+
+Create a second controller action called `GetData`. This will be for an API endpoint to retrieve a document given a key.
+
+For this tutorial, we'll write a N1QL (SQL for JSON) query to retrieve the document:
+
+[source,SQL]
+----
```
blbryant, line 139: We want this to be a secondary lookup, not a primary key lookup.  SELECT * FROM default WHERE <attribute> ...

modules/getting-started/pages/quickstart-dotnet27-aspnetcore31-visualstudio-first-query-cb65.adoc
mgroves: I hesitate to use cURL because Powershell cURL is (annoyingly) not the same. However, Postman can import cURL statements, so I made a note of that and added the cURL sample.

modules/getting-started/pages/quickstart-dotnet27-aspnetcore31-visualstudio-first-query-cb65.adoc
mgroves: This occurred to me after talking with @deniswsrosa this morning. Not sure what I was thinking 😜, but I've changed it to use temperatureC instead of the META().id

PR #1222, opened 2020-04-15
DOC-6613: Added info on service availability during rack outages

tonyjhillman:


okuzmin (approved):
Looking good, thank you!

amarantha-k (approved):

PR #1231, opened 2020-04-16
DOC-6602 tutorial for .net console app with linq2couchbase

mgroves:


httpJunkie:
Do we need to add the `.gitignore` file?

httpJunkie (approved):
Looks good, I did add one comment. Do we really need to upload the `.gitignore` file?

It doesn't bother me if you think it is needed. I understand having the `csproj` and `Program.cs` for reference.

mgroves:
One of the complaints I've heard is that samples "aren't complete enough". So I wanted to make sure a complete working sample is available (namespaces, csproj, etc), even if I don't show everything within the tutorial itself (there's a link at the bottom of my tutorial to the project).

PR #1235, opened 2020-04-17
Small description correction

jon-strabala:
Text was for "Deploy" "From now" the description of a "Resume" is slightly different


amarantha-k (approved):
Lgtm

PR #1238, opened 2020-04-20
DOC-6395: New non-root install and upgrade

tonyjhillman:


mcpaddy (commented):
On the whole it looks good. I was wondering if we need to explain about the ulimit and what they should be set to in a non-root environment This was raised in DOC-2651 so maybe it should be in a different commit. I do think it will be good to clean it up as we are trying to make the non-root install process as simple as possible.

mcpaddy (approved):

amarantha-k (approved):

PR #1249, opened 2020-04-22
DOC-6644for6.0: Fixed formatting

tonyjhillman:


amarantha-k (approved):
Not a blocker for this PR but noticed the use of "left-click*". Unless there is a specific reason, we probably should update the occurrences to "click" at the next possible opportunity.

PR #1252, opened 2020-04-23
Release notes for v6.5.1

amarantha-k:


tonyjhillman (changes requested):
The areas within which geospatial search queries return matches can now be specified as polygons, in addition to circles and rectangles. Each polygon is expressed as a series of longitude-latitude coordinates, each determining the location of one corner of the polygon.

tonyjhillman (approved):

PR #1264, opened 2020-04-27
Fixing link to node SDK

mgroves:
The link is broken on the site. I *think* it's just a matter of an extra colon?

Also, why does this link just to node? Would it be okay to change this to something like: 

**Execute N1QL queries programmatically using the official Couchbase SDKs: <Java> <.NET> <Node> <Python> <PHP> <Scala>**

So the reader can choose the path they want with a single click instead of being dropped to node and then having to figure out how to switch to a different SDK?

### DIFF
--- a/modules/getting-started/pages/try-a-query.adoc
+++ b/modules/getting-started/pages/try-a-query.adoc
@@ -275,7 +275,7 @@ The final step in the _Getting Started_ sequence, xref:choose-your-next-steps.ad
 
 == Other Destinations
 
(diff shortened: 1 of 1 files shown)


RichardSmedley:
Have fixed the link, but which SDK to link to is indeed another question, @mgroves

amarantha-k (approved):

mgroves:
It looks like github mangled my message, but would something like 'Execute N1QL queries programmatically using the official Couchbase SDKs: [.NET] [Java] [Node] ... etc' work?

RichardSmedley:
Yes - I think there may be an example of that somewhere.

mgroves:
Cool, if you can point me to that, I'll amend this PR with it

mgroves:
Looks like I can't amend the PR since you deleted the branch. No worries, I can submit another PR.

PR #1271, opened 2020-04-29
links to all languages under 'other destinations'

mgroves:
Side notes: the C page is blank, and the Scala page kicks me to the docs homepage.


mgroves:
Nope, I'm wrong about Scala kicking me to the homepage. BUT if I start on Python, whenever I use the dropdown at the top to switch to a different SDK, I get kicked to the homepage. Shall I file a DOC ticket for that?

RichardSmedley:
That's a temporary artefact of 3.0 being beta,
and 2.5 the current Python default.

RichardSmedley (approved):
Thanks @mgroves   :+1:

PR #1276, opened 2020-04-30
DOC-5475: Remove unnecessary version information

simon-dew:
Docs issue: [DOC-5475](https://issues.couchbase.com/browse/DOC-5475)

- Removed _(Introduced in ...)_ captions for versions earlier than 6.5.
- Removed `:page-status:` attribute for versions earlier than 6.5.
- Removed section `status` role for features that were not specific to a patch release.

The orange status label should now only appear in the following cases:

1. For pages only: to show that this feature is new in the current version.
1. For pages only: to show that this feature is specific to a developer preview.
1. For sections within a page: to show that a feature is specific to a patch release.

The italic _(Introduced in ...)_ caption should now only appear in the following cases:

1. For sections within a page: to show that this feature is new in the current version.

Anecdotal comments about versions, e.g. "Since Couchbase Server 5.5, you can..." still occur in the running text — I think this is okay.

### DIFF
--- a/modules/learn/pages/services-and-indexes/indexes/index_pushdowns.adoc
+++ b/modules/learn/pages/services-and-indexes/indexes/index_pushdowns.adoc
@@ -15,8 +15,6 @@ This helps performance, predominantly, in two ways:
 
 == Index Projection
 
-_(Introduced in Couchbase Server 5.0)_
-
 When processing a SELECT or DML with a WHERE clause, the Query engine picks one or more qualified indexes to be used for the query.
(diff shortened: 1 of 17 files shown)


simon-dew:
I also added `ifeval::['{page-component-version}' == '6.5']` directives to future-proof any captions stating _(Introduced in Couchbase Server 6.5)_. This means that when we create the release/6.6 branch, these captions will be hidden automatically.

amarantha-k (approved):
Thanks for taking care of these! lgtm

PR #1279, opened 2020-05-01
Please update the SEO fields for this page

Shweta0489:
Title: Database Clustering - Buckets, Memory, and Storage
Description: Couchbase Server keeps items in Buckets. Before an item can be saved, a bucket must exist for it. ✓ Learn more about database clustering with Couchbase!
H1: Database Clustering - Buckets, Memory and Storage


tonyjhillman (changes requested):
I don't agree with this. Can we please discuss before any such change is made? The term introduced is not supported by the content - in fact, we've never used it before anywhere. So, if there is a desire to change terminology in order to improve SEO, we need to determine how to attain consistency throughout (at least) the server docs.

amarantha-k:
Closing without merge as this needs further discussion before we can make the change.

PR #1280, opened 2020-05-01
Please update the SEO fields for this page

Shweta0489:
Title: vBuckets - Database Sharding Explained | Couchbase Docs

Description: vBuckets (sometimes referred to as shards) are virtual buckets that distribute data across a cluster and support replication. ✓ Learn more with Couchbase!

H1: vBuckets - Database Sharding

### DIFF
--- a/modules/learn/pages/buckets-memory-and-storage/vbuckets.adoc
+++ b/modules/learn/pages/buckets-memory-and-storage/vbuckets.adoc
@@ -1,4 +1,4 @@
-= vBuckets
+= vBuckets - Database Sharding
 :page-aliases: understanding-couchbase:buckets-memory-and-storage/vbuckets,architecture:core-data-access-bucket-disk-storage,architecture:core-data-access-vbuckets-bucket-partition
 
 [abstract]


tonyjhillman (changes requested):
My comment is the same as for 1279: I don't agree with this. Can we please discuss before any such change is made? The term introduced is not supported by the content - in fact, we've never used it before anywhere. So, if there is a desire to change terminology in order to improve SEO, we need to determine how to attain consistency throughout (at least) the server docs.

Shweta0489:
Sure. Thanks for the feedback.

I will keep you updated on the same.

tonyjhillman:
No. As I asked, can we discuss? There is a specific issue with the terms 'shard' and 'sharding' within Couchbase Server, such that this would, I fear, be a mistake. Please let's talk before we iterate further.

Shweta0489:
Sure, Tony. I agree with you.

It will be helpful if you can provide some suggestions so that we can ask our SEO agency to work on the same and then you can review it again. Let’s discuss in detail about the same in the meeting.

I met Amarantha today and we decided to set up a call including you and Marc to discuss the SEO metadata updates for the Docs pages.

I will send you an invite for this meeting.

Please let me know in case of any concerns.

Thank you

amarantha-k:
Closing without merge as this needs further discussion before we can make the change.

PR #1314, opened 2020-05-26
Update Conventions To Include Backticks

CodeItQuick:


mgroves:
Should this mention "escaping" or that it's an "escape" character?

simon-dew (commented):
Thanks very much for the contribution!

This page is about the notational conventions used in the N1QL syntax diagrams. This contribution seems to be about the N1QL syntax itself, so it doesn't belong on this page.

The use of backticks for escaping is described in the [Escaped Identifiers section on the Identifiers page](https://docs.couchbase.com/server/current/n1ql/n1ql-language-reference/identifiers.html#escaped-identifiers). Please feel free to submit another pull request against that page if you think it could be improved.

Leaving this pull request open for the moment in case there's anything I've misunderstood.

CodeItQuick:
Yeah looks like this is addressed elsewhere in the code, as you pointed out. You can close the PR.

PR #1319, opened 2020-05-29
DOC-3949a: Added ephemeral port text based on review

tonyjhillman:


tonyjhillman:
Have added text provided by David Saadeh, who expressed that we should go ahead based on that. So this is thereby approved.

amarantha-k (approved):
thanks for following up on the review. lgtm

PR #1323, opened 2020-06-01
Doc 6683

tonyjhillman:


tonyjhillman:
I have the following specific questions. Thanks in advance for your help.

In my description of the Certificate option, I have written that 'If Couchbase is 
selected, the root certificate for the Couchbase-Server cluster is used for validation: 
this assumes that the same root certificate is installed for the LDAP server.' I
have also written that 'Paste Cert should be selected if the LDAP server root certificate 
is not the same certificate as has been installed for the Couchbase-Server cluster.' Are
these statements correct?

In the Contact LDAP host section, in talking about the Bind DN field, it says 'The uid 
will be the username with which the user has logged into Couchbase Server.' But does
it need to be? If this is the UID of the Couchbase Administrator who is setting things
up, shouldn't this actually be their username as known to the LDAP server (whether or 
not it is their username on Couchbase Server)?

Regarding specifying Client Certificate and key: we appear to allow these to be specified
either with or without an LDAP root certificate also being specified. I'm assuming that
if no LDAP root certificate is specified, the LDAP server examines the client cert's
chain, in order to determine the client's root authority. But what happens if an LDAP root 
certificate *is* also specified? Does the LDAP server ignore it, or does it perhaps
accept it as the client's root authority, and therefore not bother examining the client's
chain?

In the Configure LDAP with the CLI section, our example specifies a '--bind-password ', 
but doesn't specify a '--bind-dn'. Should it?

Thanks again.

modules/manage/pages/manage-security/configure-ldap.adoc
```diff
@@ -121,40 +125,64 @@ image::manage-security/configureLDAPencryptionPullDownMenu.png[,200,align=left]
...
-
+** *Credentials*.
+Checking this checkbox allows credentials to be specified by the user, and submitted to the LDAP host for authentication.
+The UI expands, to display two additional fields:
++
+image::manage-security/contactLDAPhostCredentials.png[,450,align=left]
++
+*Bind DN* requires the LDAP _Distinguished Name_ of the user who will perform user-search and groups synchronization.
+This user needs to have read only access to the LDAP server, in order to be able to search for users and groups.
+The `uid` will be the username with which the user has logged into Couchbase Server.
```
timofey-barmin: The line about uid is not needed.

timofey-barmin (commented):
> In my description of the Certificate option, I have written that 'If Couchbase is
selected, the root certificate for the Couchbase-Server cluster is used for validation:
this assumes that the same root certificate is installed for the LDAP server.' I
have also written that 'Paste Cert should be selected if the LDAP server root certificate
is not the same certificate as has been installed for the Couchbase-Server cluster.' Are
these statements correct?

Yes, this is correct

>In the Contact LDAP host section, in talking about the Bind DN field, it says 'The uid
will be the username with which the user has logged into Couchbase Server.' But does
it need to be? If this is the UID of the Couchbase Administrator who is setting things
up, shouldn't this actually be their username as known to the LDAP server (whether or
not it is their username on Couchbase Server)?

This sentence about uid in this context doesn't make sense to me. We need to remove it. "Bind DN" is essentially a name of the user that will be used for searching. Actually current description on our web page looks pretty accurate: "The LDAP Distinguished Name of the user to perform user-search and groups synchronization. This user needs to have read only access to the LDAP server, in order to be able to search for users and groups."

>Regarding specifying Client Certificate and key: we appear to allow these to be specified
either with or without an LDAP root certificate also being specified. I'm assuming that
if no LDAP root certificate is specified, the LDAP server examines the client cert's
chain, in order to determine the client's root authority. But what happens if an LDAP root
certificate is also specified? Does the LDAP server ignore it, or does it perhaps
accept it as the client's root authority, and therefore not bother examining the client's
chain?

I think the UI dialog is a bit confusing. There is a big difference between those certs.
_LDAP root certificate_ is used to verify LDAP server's identity. Couchbase Server connects to LDAP server and Couchbase Server needs uses this cert to make sure that it talks to a correct LDAP server (make sure that LDAP server wasn't replaced by an attacker, for example).

_Client certificate_ is used by Couchbase Server for authentication on LDAP server. So, this time LDAP server checks Couchbase Server's identity. Another way of doing this is to authenticate using BindDN and password.

As you see, it's perfectly fine if both certs are specified.
   
> In the Configure LDAP with the CLI section, our example specifies a '--bind-password ',
but doesn't specify a '--bind-dn'. Should it?

Yes, --bind-dn should be specified.
```

File: modules/manage/pages/manage-security/configure-ldap.adoc [hunk 2]
```diff
@@ -121,40 +125,64 @@ image::manage-security/configureLDAPencryptionPullDownMenu.png[,200,align=left]
 +
 The options are *None* (to connect without encryption &#8212; this is insecure, and therefore is _not_ recommended), *TLS* (to connect to a TLS-encrypted port), and *StartTLSExtension* (to upgrade an existing connection).
 
-* *Certificate Validation*. Whether to validate Couchbase Server with the server certificate.
-This set of radio-buttons is enabled only if *TLS* or *StartTLSExtension* has been selected from the *Encryption* pull-down menu.
+* *Certificate*. Whether to validate Couchbase Server by presenting the root certificate of the LDAP server.
```
timofey-barmin: In my opinion, word 'with' was better in this case, because this cert is used by Couchbase Server. If you think it's still clear after the change, then I don't mind the change.

modules/manage/pages/manage-security/configure-ldap.adoc
```diff
@@ -121,40 +125,64 @@ image::manage-security/configureLDAPencryptionPullDownMenu.png[,200,align=left]
...
++
 With data entered into these fields, the dialog might appear as follows:
-
++
 [#configure-ldap-dialog-half-complete]
 image::manage-security/configureLDAPhalfComplete.png[,520,align=left]
++
+[#client-certificate]
+** *Client Certificate* allows Couchbase Server to authenticate with LDAP by means of a client certificate.
+For information on client-certificate configuration, see xref:manage:manage-security/configure-server-certificates.adoc[Configure Client Certificates].
```
timofey-barmin: I'm not sure it is the right place for this link here. Client-cert authentication for Couchbase users is a completely different topic and this link rather confuses than helps.

modules/manage/pages/manage-security/configure-ldap.adoc
```diff
@@ -266,14 +300,17 @@ When the *LDAP Query* radio-button is selected, the *Query for Groups Using* pan
 image::manage-security/ldapConfigurationLDAPquery.png[,400,align=left]
 
 Selection of *LDAP Query* instructs Couchbase Server to perform an LDAP search, in order to retrieve a list of the user's groups.
-For explanations of *Base*, *Filter*, *Scope*, see https://ldap.com/the-ldap-search-operation/[The LDAP Search Operation^].
-When the search is conducted, `%u` is replaced with the specified username; and
-`%D` is replaced with the user's DN.
+*Base* is the _search base_ https://ldap.com/the-ldap-search-operation/[DN].
+*Filter* is the https://ldap.com/the-ldap-search-operation/[search filter].
```
timofey-barmin: Probably this link is better for 'filter': https://ldap.com/ldap-filters/

modules/manage/pages/manage-security/configure-ldap.adoc
```diff
@@ -438,18 +475,18 @@ To configure an LDAP server to be used as a point of authentication for Couchbas
 --password password \
 --hosts 172.23.124.25 --port 389 \
 --encryption startTLS \
---ca-cert '/path/to/cert' \
---query-dn 'cn=admin,dc=example,dc=com' \
---query-pass 'password' \
+--ldap-cacert '/path/to/cert' \
+--user-dn-query 'cn=admin,dc=example,dc=com' \
```
timofey-barmin: This is actually --bind-dn

modules/manage/pages/manage-security/configure-ldap.adoc
```diff
@@ -438,18 +475,18 @@ To configure an LDAP server to be used as a point of authentication for Couchbas
...
+--bind-password 'password' \
 --authentication-enabled 1 \
---user-dn-mapping '{"template":"uid=%u,ou=users,dc=example,dc=com"}' \
+--user-dn-template '{"template":"uid=%u,ou=users,dc=example,dc=com"}' \
```
timofey-barmin: should be
```
--user-dn-template 'uid=%u,ou=users,dc=example,dc=com' \
```

modules/manage/pages/manage-security/configure-ldap.adoc
```diff
@@ -438,18 +475,18 @@ To configure an LDAP server to be used as a point of authentication for Couchbas
...
 --authorization-enabled 1 \
 --group-query 'ou=groups,dc=example,dc=com??one?(member=%D)'
 ----
 
 This call references the LDAP server at `172.23.125.25`, on port `389`, and specifies `--authorization-enabled` and `--authentication-enabled` for the user-credentials that will be passed from Couchbase Server.
 The argument specified for `--group-query` is the query that retrieves the LDAP groups of which the user is a member.
-A `--user-dn-mapping` is specified as a distinguished name template.
+A `--user-dn-template` is specified as a distinguished name template.
```
timofey-barmin, line 485: Please also update the 'couchbase-cli-setting-ldap' page.
 --user-dn-mapping option is removed.
Instead we now have two options
`--user-dn-template` and `--user-dn-query`
You can take description for those options from CLI's help

modules/rest-api/pages/rest-configure-ldap.adoc
```diff
@@ -122,4 +122,12 @@ Value must be an integer between 1 and 100: the default is 10.
 | cacheValueLifetime
 | Lifetime of values in cache in milliseconds. Default 300000 ms.
 
+| clientTLSCert
+| Path to a certificate to be used by Couchbase Server as a client, in authentication with LDAP.
```
timofey-barmin: Not the path to a certificate, but certificate itself

modules/rest-api/pages/rest-configure-ldap.adoc
```diff
@@ -122,4 +122,12 @@ Value must be an integer between 1 and 100: the default is 10.
...
+The `serverCertValidation` flag must be set to `true`, and a value must be specified for the `clientTLSkey` flag.
```
timofey-barmin: This is not correct. `serverCertValidation` has nothing to do with client cert authentication.

modules/rest-api/pages/rest-configure-ldap.adoc
```diff
@@ -122,4 +122,12 @@ Value must be an integer between 1 and 100: the default is 10.
...
+
+| clientTLSkey
+| Path to the private key that corresponds to Couchbase Server's client certificate, for use in authentication with LDAP.
```
timofey-barmin: The same as for clientTLSCert. Not the path, but key itself.

modules/rest-api/pages/rest-configure-ldap.adoc
```diff
@@ -122,4 +122,12 @@ Value must be an integer between 1 and 100: the default is 10.
...
+The `serverCertValidation` flag must be set to `true`, and a value must be specified for the `clientTLSCert` flag.
```
timofey-barmin: This is not correct. serverCertValidation has nothing to do with client cert authentication.

modules/manage/pages/manage-security/configure-ldap.adoc
tonyjhillman: Yes, I've modified this.

modules/manage/pages/manage-security/configure-ldap.adoc
tonyjhillman: Thanks. Have removed.

modules/manage/pages/manage-security/configure-ldap.adoc
tonyjhillman: I've removed.

modules/manage/pages/manage-security/configure-ldap.adoc
tonyjhillman: Have changed the link.

modules/manage/pages/manage-security/configure-ldap.adoc
tonyjhillman: I've changed --user-dn-query to --bind-dn, and have kept the value the same.

modules/manage/pages/manage-security/configure-ldap.adoc
tonyjhillman: I've made the substitution.

modules/manage/pages/manage-security/configure-ldap.adoc
tonyjhillman: Looks like this has been handled by Patrick's team: https://docs.couchbase.com/server/6.5/cli/cbcli/couchbase-cli-setting-ldap.html

modules/rest-api/pages/rest-configure-ldap.adoc
tonyjhillman: Have removed both instances of the sentence.

timofey-barmin (approved):

tonyjhillman:
Thanks!

PR #1330, opened 2020-06-02
Added an image for a FTS cluster.

keshavmr:


tonyjhillman (changes requested):
I guess I'm not clear on why there should be a cluster diagram at this point (fts-performing-searches.adoc, first section). We have such information elsewhere, and can optionally point to it, and even develop it, if appropriate. Specifically, this particular diagram has three instances of Full Text Search, two each of Data, one each of Query and Index. This may imply that that is how the cluster needs to be configured: although in fact, a single instance of Search on a single node would be the minimum. I might suggest that, rather than putting in a diagram, we discuss the options for deploying the Search Service, and point the reader to the other sections that cover service-deployment generically.

PR #1349, opened 2020-06-12
Removing the section "Supported token-filter languages"

abhinavdangeti:
This information is misleading, the supported language
    analyzers are already tabulated in a later section.


abhinavdangeti:
Ping @amarantha-k  - updating the analyzers' list on the FTS page, note that although we've support for some of these, they're not made available in the couchbase's CreateIndex page, also see here: http://bleveanalysis.couchbase.com/analysis

amarantha-k (approved):
lgtm

PR #1355, opened 2020-06-15
Fix up formatting of FTS query links so they all look consistent

abhinavdangeti:


abhinavdangeti:
Hi @amarantha-k , the links that show up in this section https://docs.couchbase.com/server/6.5/fts/full-text-intro.html#features-of-full-text-search for the supported FTS queries do not look consistent at all, this PR aims to fix that.

amarantha-k (approved):

PR #1359, opened 2020-06-17
Doc 6735

tonyjhillman:


modules/learn/pages/clusters-and-availability/hard-failover.adoc
```diff
@@ -21,6 +21,29 @@ Hard failover can be initiated by means of:
...
 
+[#safe-and-unsafe]
+== Hard-Failover Modes: Safe and Unsafe
+
+Either of the following two _modes_ can be specified for hard failover:
+
+* _Safe_ (the default mode).
+When hard failover has been initiated, it is continued by Couchbase Server only in the following circumstances:
+
+** Data-loss cannot result.
```
dave-finlay: Actually not true. This needs to be removed. Graceful failover should not result in data loss. Regular hard failover will likely result in data loss. Unsafe hard failover is hard failover where the cluster manager can contact only a minority of nodes. 

The problem with this is that simultaneously another admin could be failing over the minority of nodes from the other side of the network partition. We have no way of stopping this and if these two hard failovers were to happen simultaneously the cluster configuration would be entirely corrupt when the network partition heals.

So in unsafe failover we require the administrator to tell us that no such failover is being performed from the other side of the partition and that the nodes being failed over are never going to "come back"

modules/learn/pages/clusters-and-availability/hard-failover.adoc
```diff
@@ -21,6 +21,29 @@ Hard failover can be initiated by means of:
 
 For further information on initiating hard failover, see xref:manage:manage-nodes/failover-hard.adoc[Perform Hard Failover].
 
+[#safe-and-unsafe]
+== Hard-Failover Modes: Safe and Unsafe
```
dave-finlay: I don't want to call "safe" hard failover as it sounds like contradiction in terms. I'd prefer to just call the whole thing "hard failover" and then refer to the fact that within the sphere of hard failover sometimes the admin is failing over some nodes and the orchestrator cannot contact a majority of nodes to agree that the nodes should be failed over. This is an extra wrinkle and we call it unsafe because we can't guarantee that another admin isn't doing the opposite failover from the other side of the network partition.

modules/learn/pages/clusters-and-availability/hard-failover.adoc
```diff
@@ -21,6 +21,29 @@ Hard failover can be initiated by means of:
...
+Either of the following two _modes_ can be specified for hard failover:
+
+* _Safe_ (the default mode).
+When hard failover has been initiated, it is continued by Couchbase Server only in the following circumstances:
+
+** Data-loss cannot result.
+
+** A majority of nodes can still be contacted, following a node’s becoming unresponsive.
+
+** A majority of groups can still be contacted, following a group’s becoming unresponsive.
```
dave-finlay: Not true. The only condition that determines whether a hard failover is unsafe is if a majority of nodes cannot be contacted to get agreement that the nodes being failed over should be failed over.

tonyjhillman: I have modified the passage, and updated other pages accordingly.  Hopefully, this is closer.

dave-finlay:
Basically looks good to me. But @mcpaddy should take a look at what you wrote as it relates to the CLI. Thanks Tony!

mcpaddy (approved):
Looks good from the Couchbase-cli side.

PR #1371, opened 2020-06-25
DOC-6681: Document FTS FLEX index

simon-dew:

* [SELECT Syntax › USE Clause › index-ref](https://simon-dew.github.io/docs-site/DOC-6681/server/6.6/n1ql/n1ql-language-reference/select-syntax.html#index-ref) — index name optional
* [USE Clause › USE INDEX Clause](https://simon-dew.github.io/docs-site/DOC-6681/server/6.6/n1ql/n1ql-language-reference/hints.html#use-index-clause) — index name optional
* [Full-Text Search Indexes in N1QL](https://simon-dew.github.io/docs-site/DOC-6681/server/6.6/n1ql/n1ql-language-reference/flex-indexes.html) — new page on N1FTY FLEX
* [Search Service › Accessing the Search Service with N1QL](https://simon-dew.github.io/docs-site/DOC-6681/server/6.6/learn/services-and-indexes/services/search-service.html#search_via_query) — add link to new N1FTY FLEX page

Docs issue: [DOC-6681](https://issues.couchbase.com/browse/DOC-6681)


modules/n1ql/examples/n1ql-language-reference/n1ql.ebnf
```diff
@@ -0,0 +1,134 @@
...
+unnest-type ::= 'INNER' | ( 'LEFT' 'OUTER'? )
+
+/* USE Clause */
+
+use-clause ::= use-keys-clause | use-index-clause
+use-keys-clause ::= 'USE' use-keys-term
+use-keys-term ::= 'PRIMARY'? 'KEYS' expr
+use-index-clause ::= 'USE' use-index-term
+use-index-term ::= 'INDEX' '(' index-ref ( ',' index-ref )* ')'
+index-ref ::= index-type | index-name index-type?
```
sitaramv: USE INDEX ( USING GSI)
USE INDEX ( ix1 USING GSI)
both valid, index-name is is optional

sitaramv (approved):

modules/n1ql/examples/n1ql-language-reference/n1ql.ebnf
simon-dew: sure, but `USE INDEX ( ix1 )` is valid as well (defaults to `USING GSI`)

simon-dew: `USE INDEX ()` doesn't actually generate an error... 🤷‍♂️ but it seems meaningless

binhquocle:
The Overview -> Restrictions section regarding pushdown could be misinterpreted as something that the user cannot do. So while it is good that we call these out, we should make it clear that this is only for performance consideration. Functionally, Limit/Offset/Agg will still work.

simon-dew:
> The Overview -> Restrictions section regarding pushdown could be misinterpreted as something that the user cannot do. So while it is good that we call these out, we should make it clear that this is only for performance consideration. Functionally, Limit/Offset/Agg will still work.

Done

modules/n1ql/examples/n1ql-language-reference/n1ql.ebnf
simon-dew: Corrected the syntax to reflect this

modules/n1ql/pages/n1ql-language-reference/hints.adoc
```diff
@@ -227,6 +224,23 @@ WHERE sourceairport = "SFO";
 ----
 ====
 
+.Use any suitable Full Text Search index
+====
+Specify that the query service should prefer an FTS index, without specifying the index by name.
+To qualify for this query, there must be an FTS index on state, city, and type, using the keyword analyzer.
```
abhinavdangeti: The FTS index doesn't need to index "city" for this case.
Optionally if it was a type-mapped index on "hotel", it doesn't need to have "type" indexed as well.

abhinavdangeti (commented):
Looking good @simon-dew

modules/n1ql/pages/n1ql-language-reference/hints.adoc
simon-dew: Done

PR #1385, opened 2020-07-06
DOC-TQD1: Added sentence about queue depth and threads

tonyjhillman:


owend74 (commented):
Hi Tony,  I think the "A starting-point for experimentation" is great for setting the correct tone.  However I recommend the following change: 

"A starting-point for experimentation is to set the numbers for reader threads and writer threads  _equal_ to the _queue depth_ of the underlying I/O subsystem."

PR #1389, opened 2020-07-07
DOC-6873for6.5: Added LWW to timestamp explanation

tonyjhillman:


modules/learn/pages/clusters-and-availability/xdcr-conflict-resolution.adoc
```diff
@@ -100,11 +100,13 @@ For more information, see xref:manage:manage-buckets/create-bucket.adoc[Create a
 Choosing a conflict resolution method requires consideration of the logic of the applications that require the data.
 This is illustrated by the following examples:
 
-* _Most Updates Wins_: A hit-counter, for a website, is stored as a document within Couchbase Server: a value within the document is incremented each time the website is accessed.
+* _Sequence-Number-based_, whereby the document with the higher number of updates wins.
```
amarantha-k, line 103: I think we should keep the "Most updates wins" and "most recently updates/last write" references in these examples.

amarantha-k (changes requested):
I think we should keep the "Most updates wins" and "most recently updates/last write" references in these examples.

amarantha-k (approved):

PR #1393, opened 2020-07-07
Add new "pause" description and add password

jon-strabala:
A description of "Pause" was missing, also the "Deploy" block doubles as "Resume" if the function was paused.

Fix credentials instead of Administrator@ it should be Administrator:password@ for all occurances.

### DIFF
--- a/modules/eventing/pages/eventing-api.adoc
+++ b/modules/eventing/pages/eventing-api.adoc
@@ -65,20 +65,20 @@ Sample API:
 
 ----
 curl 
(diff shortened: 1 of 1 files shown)


eric-schneider (approved):
lgtm! 👍

eric-schneider:
@jon-strabala Will you be creating PRs for any necessary backports of this documentation?

jon-strabala:
> @jon-strabala Will you be creating PRs for any necessary backports of this documentation?

Yes but only after I get some more feed back from engineering on some undocumented API calls I just came across.

PR #1410, opened 2020-07-11
DOC-7048: Update Docker installation page

eric-schneider:
Although the original ticket called for just fixing an image on the Docker installation page ([DOC-7048](https://issues.couchbase.com/browse/DOC-7048)), I ended up doing a full refresh of the doc.


tonyjhillman (changes requested):
Not sure if 'Web browser' needs 'W' rather than 'w' (line 71, also 202) - is that standard?

I think we are missing an italics closure ('_') on line 134, or so. The same symbology on line 255 switches the italics off. Maybe use the html code for the underscore?

On 217 we have 'dialogue'. Do we follow a standard on this? I typically use 'dialog'. And 309 and 314.

I think 240 needs rewording, eg '' all physical hosts must be able to discover one another".

eric-schneider:
> Not sure if 'Web browser' needs 'W' rather than 'w' (line 71, also 202) - is that standard?

Changed to lower case 'w'

> I think we are missing an italics closure ('_') on line 134, or so. The same symbology on line 255 switches the italics off. Maybe use the html code for the underscore?

This is actually an underscore in the Docker Hub URL. The hyperlink seems to generate fine and it appears that Antora/Asciidoctor are properly interpreting it as part of the URL.

> On 217 we have 'dialogue'. Do we follow a standard on this? I typically use 'dialog'. And 309 and 314.

Not sure what the AP Stylebook says, but W3C appears to use "dialog", so I went ahead and changed it on this page.

> I think 240 needs rewording, eg '' all physical hosts must be able to discover one another".

I made some slight adjustments to the wording.

amarantha-k (approved):

tonyjhillman (approved):

PR #1414, opened 2020-07-13
DOC-6662: Added notes and text mods on deprecation of subject common …

tonyjhillman:
…name

### DIFF
--- a/modules/learn/pages/security/certificates.adoc
+++ b/modules/learn/pages/security/certificates.adoc
@@ -39,7 +39,8 @@ Each certificate includes:
 This information associates the _identified entity_ (that is, the server-node or client bearing the certificate) with the issuing CA, such that third parties who know and trust the CA, but are unfamiliar with the identified entity, may elect to trust the identified entity on the basis of their trust in the CA.
 Note that all systems that use certificates for authentication maintain their own collection of trusted CA certificates: if, when an entity attempts to authenticate, the entity proves that it has itself been authorized by a CA whose certificate is included in the system's collection, the entity may be trusted by the system.
 
-Additionally, certificates may each be configured to contain _extensions_, which might be used to constrain the certificate's capabilities (for example, by indicating the IP address of the server on which the certificate must reside; or by indicating the uses to which its public key must be restricted); or to identify a _username_, based on which client-authentication and authorization can proceed.
+Additionally, certificates can each be configured to contain _extensions_, which are used to constrain the certificate's capabilities (for example, by indicating the IP address of the server on which the certificate must reside; or by indicating the uses to which its public key must be restricted).
+Every certificate should be extended with at least one _Subject Alternative Name_: see xref:learn:security/certificates.adoc#deprecation-of-subject-common-name[Deprecation of Subject Common Name], below.
 
 [#certificate-hierarchies]
 == Certificate Hierarchies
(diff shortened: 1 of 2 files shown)


modules/learn/pages/security/certificates.adoc
```diff
@@ -39,7 +39,8 @@ Each certificate includes:
...
 Note that all systems that use certificates for authentication maintain their own collection of trusted CA certificates: if, when an entity attempts to authenticate, the entity proves that it has itself been authorized by a CA whose certificate is included in the system's collection, the entity may be trusted by the system.
 
-Additionally, certificates may each be configured to contain _extensions_, which might be used to constrain the certificate's capabilities (for example, by indicating the IP address of the server on which the certificate must reside; or by indicating the uses to which its public key must be restricted); or to identify a _username_, based on which client-authentication and authorization can proceed.
+Additionally, certificates can each be configured to contain _extensions_, which are used to constrain the certificate's capabilities (for example, by indicating the IP address of the server on which the certificate must reside; or by indicating the uses to which its public key must be restricted).
```
timofey-barmin: I'm ok with "may -> can" and "might be -> are" changes, but I think the last part of the sentence should no be changed.
> or to identify a _username_, based on which client-authentication and authorization can proceed.

The old version is pretty accurate, as we don't want to insist on using of SAN for client certificates. I would leave it as it was before.

modules/learn/pages/security/certificates.adoc
```diff
@@ -192,6 +193,8 @@ Each specified _path_ can be one of the following path-types:
 
 * `subject.cn`.
 The _Subject Common Name_ specified in the certificate will be extracted.
+Note that use of Subject Common Name is now deprecated.
```
timofey-barmin: It sounds as it is deprecated in Couchbase Server. If I were a customer I would decide that I need to regenerate my certs, but that's not true. Customers can actually continue to use CN, at least right now we don't have any plans to remove support for username in CN for client certs.
In my opinion we can mention that it's deprecated in rfc6125 but we still support it.

modules/learn/pages/security/certificates.adoc
```diff
@@ -233,6 +236,20 @@ For example, a delimiter of `.` causes the substring `.couchbase.com` to be disc
 
 For step-by-step instructions, see xref:manage:manage-security/enable-client-certificate-handling.adoc[Enable Client Certificate Handling].
 
+[#deprecation-of-subject-common-name]
+== Deprecation of Subject Common Name
+
+Use of the _Subject Common Name_ to identify either a server or a client is now deprecated (see https://tools.ietf.org/html/rfc6125#section-6.4.4[section 6.4.4 of RFC 6125^]).
+For Couchbase Server, this means that:
+
+* The node-certificate for each server in the cluster must specify at least one _Subject Alternative Name_, in order to be identified: typically, the node's IP address or DNS name should be provided in this way.
```
timofey-barmin: I would not use word "must". I would say "We recommend", as we are not removing the support for CN.

modules/manage/pages/manage-security/enable-client-certificate-handling.adoc
```diff
@@ -53,7 +53,9 @@ Note that imposing this level of security likely requires the additional measure
 . If *Require Client Certificate* has been set to either *Enable* or *Mandatory*, establish how the username within each client certificate is to be determined.
 +
 Each certificate is expected to provide a _username_ as part of its content.
-Frequently, the username is specified as the certificate's _Subject Common Name_; but it may also be expressed as a _Subject Alternative Name_.
+This should be expressed as a _Subject Alternative Name_.
```
timofey-barmin: I would not change this piece. As we are talking about _client_ certs here it's pretty accurate I beleive.

modules/learn/pages/security/certificates.adoc
tonyjhillman: Ok. I've restored the previous phrase.

modules/learn/pages/security/certificates.adoc
tonyjhillman: I've added a link to the RFC, and a statement that we still support it.

modules/learn/pages/security/certificates.adoc
tonyjhillman: Ok, I've changed accordingly.

modules/manage/pages/manage-security/enable-client-certificate-handling.adoc
tonyjhillman: Ok, I've restored the original.

modules/learn/pages/security/certificates.adoc
tonyjhillman: I've also moved this to a place a little earlier in the section, where client's use of cn is first described.

timofey-barmin (approved):

PR #1427, opened 2020-07-22
DOC-7096: Updated pam routine

tonyjhillman:


modules/manage/pages/manage-security/configure-pam.adoc
```diff
@@ -88,40 +89,45 @@ If output to the above command does not confirm that `MECHANISMS` is set to `pam
...
 image::manage-security/securityViewInitial.png[,820]
 
-. Left-click on the btn:[Add User] button, situated near the right.
+. Left-click on the *ADD USER* button, situated near the right.
 This brings up the [.ui]*Add New User* dialog.
 Select the [.ui]*External* radio-button, in the [.ui]*Authentication Domain* panel at the upper left.
-Then, enter the name of the new user you are creating, and specify a suitable role, such as [.ui]*Cluster Admin*.
-The panel appears as follows:
+Then, enter the name of the new user you are creating.
+Note that since PAM authentication has not yet taken effect, the notification `not found` now appears above the username-field.
```
timofey-barmin: Actually this is a bug. It actually tries to search for LDAP user with provided name. It only makes sense when native LADP is enabled. We need to fix that in the UI I think. I will create a ticket for that.
Probably we can say something like that: "you might see 'not found' notification if native LDAP is enabled as well"

timofey-barmin (approved):

PR #1438, opened 2020-07-24
DOC-7018: Upgrade feature list for 6.6

tonyjhillman:


modules/learn/pages/clusters-and-availability/connectivity.adoc
```diff
@@ -53,6 +53,13 @@ The address family can be established as part of a cluster's initial configurati
 
 For information on using the CLI to change the address family for an existing cluster, see xref:manage:manage-nodes/manage-address-families.adoc[Manage Address Families].
 
+[#services-and-ip-address-families]
+=== Services and IP-Address Families
+
+_All_ Couchbase xref:learn:services-and-indexes/services/services.adoc[Services] bind on _all_ their assigned ports with the same address family &#8212; either IPv4 or IPv6, depending on which address family has been established for the cluster; either at initial configuration, or subsequently by means of the CLI.
```
amithk, line 59: > either IPv4 or IPv6

For index service, this is a regression from earlier 6.5.1 behaviour. In 6.5.1, the server used to bind to all the available ip-address families. 

Question: Do we need to mention about this behavioural regression explicitly? or this much explanation should be enough? Especially, this can impact the externally accessible ports (like index service HTTP port).

tonyjhillman: I can make this explicit in the Upgrade section. I'll add some text to the row in the table.

amarantha-k (approved):

PR #1450, opened 2020-07-30
DOC-7155: Separated cert upload and reload commands

tonyjhillman:


modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -329,6 +324,16 @@ image::manage-security/rootCertificateWithSignedCert.png[600,align=left]
 +
 As this indicates, the signed certificate has now been substituted for the default certificate (an example of whose appearance is provided in xref:manage:manage-security/manage-security-settings.adoc#root-certificate-security-screen-display[Root Certificate]).
 
+. _Reload_ the node certificate, for the current node:
```
orthogonal3: Perhaps to make it clear where the certificate has to come from: 

> _Reload_ the node certificate from disk, for the current node.

Just a little prompt to remind them that as well as performing the reload, they'll have to copy `chain.pem` and `pkey.key` onto each individual node in the `inbox` directory on that node.

orthogonal3 (approved):
One slight suggestion to tweak, but looks really good. Helps separate the 2 different bits more clearly.
Thanks!

PR #1467, opened 2020-08-06
fixed a comment and corrected a URL typo

mgroves:
there's an extra slash in one of the URLs

also, the comment says "rename node" but it's really the cluster name, not a node name


amarantha-k (approved):
thanks!

PR #1493, opened 2020-08-13
Doc 7190

tonyjhillman:


tonyjhillman:
This should be merged only after the new non-root install is made available for 6.5 and 6.6.

tonyjhillman:
This can now be merged, since the binary is available.

tonyjhillman:
The 6.6 version has already been merged. This is the 6.5 equivalent, which is also now ready to be merged.

amarantha-k (approved):

PR #1515, opened 2020-08-25
DOC-7236: Revised access scanner descriptions

tonyjhillman:


modules/cli/pages/cbepctl/set-flush_param.adoc
```diff
@@ -57,13 +57,15 @@ NOTE: Changes of thresholds are NOT persistent and must be reapplied after the b
...
++
+The `cbepctl flush_param` command is used to change the initial time and the interval for the access scanner.
 For example, the initial time and interval might be changed to accommodate a peak time when an application needs these keys to be quickly available.
 +
-By default, the scanner process runs once every 24 hours at 10:00 AM GMT.
-To reduce the cluster-wide impact of running this task, stagger the start time to a different value on each node in the cluster.
+By default, the access scanner runs once every 24 hours at 10:00 AM GMT.
+The scanner is highly CPU-intensive: therefore, to reduce the cluster-wide impact of running this task, its start time should be staggered to a different value on each node in the cluster.
+Note also that the scanner shares its thread pool with the Index Service; and may therefore adversely affect the performance of any simultaneous index-updates: if instances of this problem occur, the scanner should be either run with less frequency, or disabled.
```
malarky: More specifically, the `projector` process - not sure how this is represented in the rest of the docs - calling it the Index Service here may cause confusion if people aren't running the Index Service on this node.

modules/install/pages/sizing-general.adoc
```diff
@@ -261,6 +261,13 @@ The larger the fragmentation, the more disk you will need.
 
 The above index consumes 6 GB of disk space.
 
+=== Accommodating the Access Scanner
+
+The _access scanner_ is used to ensure high-performance data-loading, when buckets are restarted.
+The access scanner is highly CPU-intensive, and shares a thread pool with the Index Service: consequently, if scanning is performed simultaneously with index updates, the performance of the index updates may be adversely affected.
```
malarky: Same as above, need to be clear what is actually being affected here - the index updates are performed elsewhere, but this can block the mutations from making out of the Data Service across the network to the Index Service.

modules/learn/pages/buckets-memory-and-storage/memory.adoc
```diff
@@ -57,13 +57,27 @@ Minimum-required memory-quotas (which are the quota-defaults initially provided
...
-During this warmup process, data on disk is sequentially reloaded into memory.
+During this warmup process, data on disk is sequentially reloaded into memory for each bucket.
 The time required for the reload depends on the size and configuration of the system, the amount of data persisted on the node, and the ejection policy configured for the buckets.
 
-Frequently used items are identified via a scanner process, which examines an _access log_, and obtains the appropriate keys.
-Corresponding items are then loaded with the highest priority.
-The scanner process is configurable, via the CLI utility `cbepctl`, with the `flush_param` parameter.
+On a given Data Service node, a bucket's data is loaded in accordance with determinations made by the _access scanner_.
+This program runs by default once every day.
+It checks the _resident ratio_ &#8212; which is the percentage, on the current node, of bucket-items currently in memory:
```
malarky: Think it needs to be clear that this is the combined Resident Ratio across both Active and Replica vBuckets.

malarky (changes requested):
Would also suggest that folks from both Data and Index teams review these changes and agree on all the terminology.

modules/cli/pages/cbepctl/set-flush_param.adoc
owend74: "Note also that the scanner shares its thread pool with the Index Service" (or projector process).  I'm not sure I follow / agree with.  Which thread pool are we referring to? - Do we mean general CPU resources? Memcached and KV are separate processes and so don't have any direct interact other than communicating over DCP.

modules/learn/pages/buckets-memory-and-storage/memory.adoc
owend74: Maybe rephrase to .... It checks the _resident ratio_ &#8212; which is the percentage of items in active and replica vBuckets that are currently in memory on that node.

modules/install/pages/sizing-general.adoc
owend74: See previous comment - not sure I agree with "and shares a thread pool with the Index Service"

PR #1547, opened 2020-09-07
Document Update

NiteshKoushik:
Grammatical correction


amarantha-k (commented):
Since 'FTS' begins with a vowel sound, it should precede it with an. Will close this PR without merge.

PR #1527, opened 2020-08-28
DOC-6817: Document cluster-level settings

simon-dew:
The following documentation is ready for review:

* [Settings and Parameters](https://simon-dew.github.io/docs-site/DOC-6817/server/6.6/settings/query-settings.html) — reworked to include node-level and cluster-level settings
* [General Settings › Configure General Settings with the CLI › Query Settings via CLI](https://simon-dew.github.io/docs-site/DOC-6817/server/6.6/manage/manage-settings/general-settings.html#query-settings-via-cli) — new section
* [Cluster Query Settings API](https://simon-dew.github.io/docs-site/DOC-6817/server/6.6/rest-api/rest-cluster-query-settings.html) — new REST API reference page
* [Admin REST API › Definitions › Settings](https://simon-dew.github.io/docs-site/DOC-6817/server/6.6/n1ql/n1ql-rest-api/admin.html#_settings) — added links to cluster-level settings

Docs issue: [DOC-6817](https://issues.couchbase.com/browse/DOC-6817)

PR #1548, opened 2020-09-07
DOC-6817: Document cluster-level settings

simon-dew:
Backport of #1527 to release/6.5

Docs issue: [DOC-6817](https://issues.couchbase.com/browse/DOC-6817)


eric-schneider (approved):
I'm not sure if there is a way to compare this PR to the 6.6 PR (or its squashed commit), but if there is, it would be really nice to quickly understand the different deleted/added lines in backports, especially ones this large.

simon-dew:
> I'm not sure if there is a way to compare this PR to the 6.6 PR (or its squashed commit), but if there is, it would be really nice to quickly understand the different deleted/added lines in backports, especially ones this large.

Thanks for the approval. This is an unusually big PR. I really should have split the task up into subtasks when I realized how big it was going to be.

eric-schneider:
> Thanks for the approval. This is an unusually big PR. I really should have split the task up into subtasks when I realized how big it was going to be.

@simon-dew I'm not averse to changes this large when so much of the subject matter is related (as it was here). But it would be nice if we could consult a GitHub expert to help us determine whether there's an effective and efficient way to view a diff between two PRs. If I could quickly zoom in on the areas where a line was added/modified/deleted in one branch and not in the other, I could more easily review the content that is most likely to contain problems with the backport. Does that make sense?

PR #1599, opened 2020-09-23
DOC-7402for7.0: Limitation notice on mandatory client certs with node…

tonyjhillman:
…-to-node encryption

### DIFF
--- a/modules/manage/pages/manage-security/enable-client-certificate-handling.adoc
+++ b/modules/manage/pages/manage-security/enable-client-certificate-handling.adoc
@@ -50,7 +50,7 @@ If the client does not present a certificate, the certificate-based authenticati
 No other form of client-authentication is handled over the secure connection.
 Note that imposing this level of security likely requires the additional measure of disabling non-secure console access: see xref:manage:manage-security/manage-console-access.adoc[Manage Console Access].
 +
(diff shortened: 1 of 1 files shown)


modules/manage/pages/manage-security/enable-client-certificate-handling.adoc
```diff
@@ -49,6 +49,9 @@ If the client does not present a certificate, the certificate-based authenticati
 * *Mandatory*: Specifies that all clients _must_ present a certificate, in order to authenticate.
 No other form of client-authentication is handled over the secure connection.
 Note that imposing this level of security likely requires the additional measure of disabling non-secure console access: see xref:manage:manage-security/manage-console-access.adoc[Manage Console Access].
++
+A current limitation is that mandatory handling of client certificates cannot be enabled if node-to-node encryption is set to _all_.
```
dave-finlay: This is a little confusing because client certs also support a setting of "enabled". Suggest modify to read:

> Note: A current limitation is that client certificate authentication cannot be set to mandatory if if node-to-node encryption is set to _all_.

tonyjhillman: I've modified as suggested.

PR #1608, opened 2020-09-28
Refined words

gurmeetsinghdke:
At 3 places 'left-click' was written which is much more descriptive than required. So changed it to 'click' as it is a web based documentation (not a window tool based) and click is enough I guess and looks more refined.


amarantha-k (approved):

amarantha-k:
Agree! Thanks for your contribution!

PR #1646, opened 2020-10-19
DOC-7506for6.6.1: Updated description of Eventing Node hard failover

tonyjhillman:


amarantha-k (approved):
Approving as Tony clarified that this PR is to back porting changes made with Jon.

PR #1661, opened 2020-10-27
DOC-7578: Added note on backup service and hard failover

tonyjhillman:


cg14823 (approved):
Looks correct to me 😄

PR #1669, opened 2020-10-29
DOC-7529: System keyspaces

simon-dew:

* [Getting System Information](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-intro/sysinfo.html)

Docs issue: [DOC-7529](https://issues.couchbase.com/browse/DOC-7529)

### DIFF
--- a/modules/n1ql/pages/n1ql-intro/sysinfo.adoc
+++ b/modules/n1ql/pages/n1ql-intro/sysinfo.adoc
@@ -12,33 +12,56 @@ The keyspace names are plural in order to avoid conflicting with N1QL keywords.
 
 N1QL has the following artifacts:
 
-* *Datastores*
-+
+Datastores::
+
 Datastores are similar to sites.
 A datastore is a database deployment, for example, a server cluster, cloud service, or mobile installation.
 It is analogous to a relation database instance.
 
-* *Namespaces*
-+
+Namespaces::
+
 Namespaces are similar to pools.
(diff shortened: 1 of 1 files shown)


marcogrecopriolo:
All perfect. Just one little note (and it's probably my bad...)
Aside from system:scopes, system:keyspaces, system:indexes, we also have system:all_scopes, system:all_keyspaces, system:all_indexes.
They look exactly the same as the non 'all_' counterpart, but also list system objects, so for instance system:all_scopes would list '_default' scopes, which system:scopes, wouldn't, system:all_keyspaces would list all the system keyspaces and would list  the same keyspace twice, once as default:bucket1 and once as default:bucket1._default._default, and system:all_indexes would also list indexes on system:keyspaces.

If you think these are confusing, we can just leave them undocumented.

simon-dew:
Thanks @marcogrecopriolo. I've actually documented `system:all_scopes` as a note under the **Querying Scopes** heading, `system:all_keyspaces` as a note under the **Querying Collections** heading, and  `system:all_indexes` as a note under the **Querying Indexes** heading. I wanted to make sure they were documented, but didn't want to get bogged down in details because as you say the behavior of `system:all_keyspaces` is confusing.

marcogrecopriolo:
perfect!

marcogrecopriolo (approved):

simon-dew:
Update links to staging docs

PR #1670, opened 2020-10-29
DOC-7542for6.5: Added information on quota reset and button

tonyjhillman:


anuthan:
I have asked PM to have a look. There were some discussions offline about how we present the auto-failover reset count information. I just want to make sure we are all on the same page.

tonyjhillman:
Based on our discussions, I'm thinking it might be as well also to be explicit as to a manual hard failover (optionally introduced in circumstances where auto won't happen) itself incrementing the count.

Correction, based on still further discussion, hard failover does not do so. So we can scratch the above suggestion.

tonyjhillman:
Approved.

amarantha-k (approved):

PR #1685, opened 2020-11-06
DOC-7484: Added scopes and collections to XDCR management sequence

tonyjhillman:


modules/manage/pages/manage-xdcr/replicate-using-scopes-and-collections.adoc
```diff
@@ -0,0 +1,178 @@
+= Replicate Using Scopes and Collections
+
+[abstract]
+XDCR can be performed with reference to scopes and collections within source and target buckets.
+
+[#understanding-scopes-collections-and-replication]
+== Understanding Scopes, Collections, and Replication
+
+As described in xref:learn:data/scopes-and-collections.adoc[Scopes and Collections], scopes and collections are containers for data within a bucket.
+A bucket can contain up to 100 scopes; and a scope can contain up to 1000 buckets.
```
nelio2k, line 10: > buckets 

collections

modules/manage/pages/manage-xdcr/replicate-using-scopes-and-collections.adoc
```diff
@@ -0,0 +1,178 @@
...
+
+As well as specifying a source bucket and a target bucket, XDCR can specify scopes and collections within those buckets.
+This allows data to be reassigned from one collection to another.
+It also provides a convenient way of _migrating_ data that currently resides in legacy buckets that do not support (or have not yet been established with) administrator-defined scopes and collections, into buckets, on a Version 7.0+ Couchbase-Server cluster, that _do_ contain administrator-defined scopes and collections.
```
nelio2k, line 14: > in legacy buckets that do not support (or have not yet been established with) administrator-defined scopes and collections

Technically, that's not possible. Remember that for the migration mode operation to even show up in the first place, customers will have already upgraded the (source) cluster to 7.0. This means that the "legacy" bucket mentioned here technically supports admin-defined scopes and collections. 

I think the point to convey here is that instead of defining it as legacy, it should be though of as "upgraded" bucket, of which all the data now all reside in the default collection by design.

> on a Version 7.0+ Couchbase-Server cluster, that _do_ contain administrator-defined scopes and collections

Consequently, the detail describing the target cluster bucket may not be necessary, since we're talking about replicating from 7.0+ to 7.0+ here.

modules/manage/pages/manage-xdcr/replicate-using-scopes-and-collections.adoc
```diff
@@ -0,0 +1,178 @@
...
+For more information, see xref:learn:clusters-and-availability/xdcr-overview.adoc#xdcr-with-scopes-and-collections[XDCR with Scopes and Collections].
+
+[#examples-on-this-page]
+== Examples on this Page
+The examples provided on this page &#8212; xref:manage:manage-xdcr/replicate-using-scopes-and-collections.adoc#migrate-data-to-a-collection[Migrate Data to a Collection] and xref:manage:manage-xdcr/replicate-using-scopes-and-collections.adoc#replicate-data-between-collections[Replicate Data Between Collections] &#8212; demonstrate how XDCR can be performed with reference to scopes and collections within source and target buckets, using Couchbase Web Console.
+The examples assume the following:
+
+* A single-node cluster named `10.144.210.101` has been created, and features the `travel-sample` bucket.
```
nelio2k, line 22: > named `10.144.210.101`

(Applies to all other clusters)
Normally, customers don't name their clusters with IP addresses. They should be named with easily distinguishable names, such as 'cluster1', 'cluster2', etc, which also has the added benefit of readability.

This also means the screenshots will have to be updated.

modules/manage/pages/manage-xdcr/replicate-using-scopes-and-collections.adoc
```diff
@@ -0,0 +1,178 @@
...
+* A single-node cluster named `10.144.210.103` has been created, and features a single, empty bucket, named `tsBackup`.
+This bucket contains a scope named `usAirlineBackupScope`; which itself contains a collection named `usAirlineBackupCollection`.
+
+* The cluster `10.144.210.101` has defined `10.144.210.102` as a remote cluster for XDCR.
+In turn, the cluster `10.144.210.102` has defined `10.144.210.103` as a remote cluster for XDCR.
+
+Before following the examples, familiarity should be gained with XDCR management-basics, explained on other pages. In particular, see xref:manage:manage-xdcr/prepare-for-xdcr.adoc[Prepare for XDCR], xref:manage:manage-xdcr/create-xdcr-reference.adoc[Create a Reference], xref:manage:manage-xdcr/create-xdcr-replication.adoc[Create a Replications], and xref:manage:manage-xdcr/filter-xdcr-replication.adoc[Filter a Replication].
+
+[#migrate-data-to-a-collection]
+== Migrate Data to a Collection
```
nelio2k, line 39: Given the context of the document is titled "Replicate using Scopes and Collections", Data migration is not necessarily the primary use case for replication. I'd argue that it is a specialized use case of upgrading/migration. My recommendation is that this section should be presented after the implicit/explicit mapping concepts.

modules/manage/pages/manage-xdcr/replicate-using-scopes-and-collections.adoc
```diff
@@ -0,0 +1,178 @@
...
+
+By specifying a scope and collection within a target bucket, XDCR can be used to replicate data selectively from the `_default` collection within one bucket to the purpose-created collection within another.
+This is demonstrated in the following example.
```
nelio2k, line 42: There are some caveats that should be presented here, like a warning/note subsection:
(Please rephrase and you may choose to omit details as you see fit):

Each migration rule that is established in the following example will be CPU intensive, and adds a noticeable degradation to the XDCR replication performance. The more migration rules, the slower each migration replication will be. It is recommended to keep the total number of migration rules (across all replications) per source cluster under 3.

If migrating a large number of collections, the best use case is to iteratively migrate a few rules at a time. Once the expected documents have finished replicating, switch over any application from the default collection to the newly migrated target collection if necessary. Then, edit the source migration rules/mapping, Save Mappings, Save Replication, and repeat the procedure until all the applications have been switched over to the target bucket/collections.

(The above performance reason is also another reason why I would prefer not advertising migration as the first-and-foremost use case. If someone is trying out Couchbase, they should be playing with the most powerful and performant feature: implicit mapping)

modules/manage/pages/manage-xdcr/replicate-using-scopes-and-collections.adoc
```diff
@@ -0,0 +1,178 @@
...
+The documents within the collection are now displayed:
++
+image::manage-xdcr/xdcr-target-collection-filled.png[,680,align=left]
++
+This indicates that the `airline` documents from `travel-sample` have been successfully filtered and replicated to the `airlineCollection` collection, within the remote bucket `ts`.
+
+[#replicate-data-between-collections]
+== Replicate Data Between Collections
+
+XDCR allows data to be replicated between collections.
```
nelio2k, line 110: In this section, it looks like it forces the user to *have* to specify collections mappings to be able to replicate collections (explicit mapping), which shouldn't be the case. XDCR was designed so that implicit mapping would be offered first. Implicit mapping should be intuitive enough and hides a lot of the management complexity from the user. Only if the implicit mapping is not enough, should the user opt in for explicit mapping.

I would recommend a section to showcase implicit mapping and try to pursuade the users to go that route. Then, explicit mapping can come in for those who need a more powerful mapping experience.

I have sent you a copy of the blog draft, which explains the concepts.

One point to note for implicit mapping: XDCR monitors the target bucket for created/removed collections at a rate of once per minute. If a target collection was missing when a replication is set up, and then created, it may take up to a minute for XDCR to replicate data to the target collection.

modules/manage/pages/manage-xdcr/replicate-using-scopes-and-collections.adoc
```diff
@@ -0,0 +1,178 @@
...
++
+In the same way, modify the remote-collection name associated with `airlineCollection` to `usAirlineBackupCollection`:
++
+image::manage-xdcr/xdcr-modified-remote-collection-name.png[,225,align=left]
++
+Note that the rules now appear in the *Mapping Rules* column, at the upper right of the screen:
++
+image::manage-xdcr/xdcr-mapping-rules-for collections.png[,300,align=left]
+
+. Filter the replication, to ensure that only documents whose `country` value is `United States` are replicated.
```
nelio2k, line 170: One clarification for users that may be helpful: The filtering, if enabled, will take place across all collections mapping, and there is no option right now to have individual filters on a specific collection->collection mapping.

modules/learn/pages/clusters-and-availability/xdcr-overview.adoc
```diff
@@ -96,6 +96,25 @@ Views and indexes can only be replicated manually, or by administrator-provided
...
+The current source may be:
+
+* A bucket on a 7.0+ cluster, housing its data in administrator-defined collections.
+Thus, data can be replicated (optionally using XDCR Advancing Filtering), from one collection to another within the same bucket; or from a collection in one bucket to a collection in another bucket.
+
+* A bucket on a 7.0+ cluster, housing its data in the `_default` collection, within the `_default` scope (this being the default initial residence for all data in a bucket whose cluster has been upgraded from a pre-7.0 Couchbase Server version to a 7.0+ version).
+Thus, XDCR can subsequently be used to redistribute the data into administrator-defined collections, either within the same or within different buckets (again, optionally using XDCR Advancing Filtering).
+
+* A bucket on a pre-7.0 cluster (which being pre-7.0, does not support scopes and collections).
+Thus, XDCR can be used to _migrate_ the data from its bucket on the pre-7.0 cluster to a bucket the 7.0+ cluster, and thereby to assign the data to appropriate scopes and collections that have been defined within the new bucket (again, optionally using XDCR Advancing Filtering).
```
nelio2k, line 113: This isn't possible. See note below.

modules/manage/pages/manage-xdcr/replicate-using-scopes-and-collections.adoc
tonyjhillman: Thanks. Have fixed. The constraints have changed, so I've updated them also.

modules/manage/pages/manage-xdcr/replicate-using-scopes-and-collections.adoc
tonyjhillman: I understand. So I'll remove references to pre 7.0 in terms of migration.

modules/manage/pages/manage-xdcr/replicate-using-scopes-and-collections.adoc
tonyjhillman: Right. So, unfortunately, this is the standard naming currently used throughout all the management doc, not just for XDCR, but for every feature - including UI, CLI, and REST examples. I can do this, but it might be better to wait until we are through the current Beta, first. The new standard I use for naming should probably be adopted elsewhere in the docs also.

modules/manage/pages/manage-xdcr/replicate-using-scopes-and-collections.adoc
tonyjhillman: I understand. Unfortunately, the problem here is that at this moment, I don't have a starting point for the examples without migration from _default. Because the reader is likely to have no collections yet set up - or at least, has not yet populated them. The plan is to have a new version of travel_sample that will be collections ready: so it would make sense to use that, and go directly to the implicit/explicit example, with source collections ready to go. But at this moment, the new sample bucket is not yet available.  When it arrives, I can make the change. (Or is there another way around this, I wonder?)

modules/manage/pages/manage-xdcr/replicate-using-scopes-and-collections.adoc
tonyjhillman: I will add this info.

modules/manage/pages/manage-xdcr/replicate-using-scopes-and-collections.adoc
tonyjhillman: I'll put in implicit examples first, and try to explain the concepts a bit more. (I'll leave the major explanation for the work I do based on your blog.)

modules/manage/pages/manage-xdcr/replicate-using-scopes-and-collections.adoc
tonyjhillman: I'll make this clear.

modules/learn/pages/clusters-and-availability/xdcr-overview.adoc
tonyjhillman: I've removed.

tonyjhillman: I'll continue this work on a new PR.

PR #1699, opened 2020-11-11
Doc 7624a

tonyjhillman:


tonyjhillman:
I've not yet tried to document the Limitations section in the doc (section 6 of https://docs.google.com/document/d/175UaKmvrjIWCelJv6UfQ1O4-QSv5D46agsXV8xQsGH4/edit?pli=1#heading=h.hofb77oeslv9) because I find I don't yet understand it fully. 

I hope I've gotten my statements about defaults correct, but I may be mistaken. One thing I'm not sure of is what happens when a service uses the global setting for cipher suites or the underlying default (assuming I'm right that it does): presumably, either of these could contain cipher-suites not included in the service' own 'supportedCipherSuites' list. In that cases, does it just ignore them?

anuthan:
@tonyjhillman : The default for cipherSuites is all supported ciphers. 
Also, the expectation is that if the cipherSuites list set by the user contains unsupported ciphers for a service, the service just ignores those ciphers and limit themselves to other ciphers on the list. However, limitation 6.1 in the doc still applies.

tonyjhillman:
I will do some more work on this before any further review. I will incorporate the information you provided on the google doc. I will rework the explanation of defaults, and will add the 1.3 limitations.

tonyjhillman:
I will close this PR and open a new one, to add revised and extended info.

PR #1718, opened 2020-11-13
Updated the link to point right URL

Shweta0489:
Updated the link to point the right URL in order to avoid SEO error.

It will be great if you can approve the change.

Thank you


amarantha-k (changes requested):
@Shweta0489 In this context, the user expects a direct link to the specific presentation and not the top level URL. Do we have a specific link to replace the current one with?

amarantha-k (approved):

PR #1721, opened 2020-11-14
DOC-7455: Updates to Index Advisor

simon-dew:

* [ADVISE](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-language-reference/advise.html) — Updated advisor output, updated examples
* [ADVISOR Function](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-language-reference/advisor.html) — removed DP flag, made changes recommended by QE

Docs issue: [DOC-7455](https://issues.couchbase.com/browse/DOC-7455)


pierrecouch (approved):

miaobingjie:
In Advise, under Recommendation Rule, second paragraph, instead of saying "cost-based optimizer fails to recommend an index", change to something like "preconditions are not met and cost-based optimizer cannot be used to recommend an index"

simon-dew:
> In Advise, under Recommendation Rule, second paragraph, instead of saying "cost-based optimizer fails to recommend an index", change to something like "preconditions are not met and cost-based optimizer cannot be used to recommend an index"

Done

simon-dew:
Update links to staging docs

PR #1752, opened 2020-11-26
Change the positioning of the note

McSpooder:
It makes more sense to have the note about spreadsheets after you talk about editing documents via spreadsheet setting.


amarantha-k (approved):
Thanks for the contribution!

PR #1780, opened 2020-12-15
DOC-7825for6.5: Added ports for future use

tonyjhillman:


tonyjhillman:
If this works, I'll add for 6.6 and 7.0 also.

timofey-barmin (changes requested):
Thanks,
We already have description for those ports (in two tables). I suggest we add a foot-note for them in both tables (similar to port 18095).
The note can say something like:
Ports 21250 and 21350 are not currently used, but are reserved for future use.

Probably we can merge similar kinds of footnotes and have one note for all "not used, reserved for future use" ports (21250, 21350, and 18095). I don't think we need to mention ports themselves in the note in this case.

tonyjhillman:
Does this look okay? Thanks.

timofey-barmin (approved):

timofey-barmin:
> Does this look okay? Thanks.

Looks great! Thanks!

amarantha-k (approved):

PR #1787, opened 2020-12-23
6.6.1 Eventing updates and two new Eventing Scriptlets

jon-strabala:


RichardSmedley (approved):
LGTM
*within my understanding of Eventing, at least.

PR #1796, opened 2021-01-13
Doc 7635

tonyjhillman:


modules/learn/pages/clusters-and-availability/xdcr-overview.adoc
```diff
@@ -110,33 +111,39 @@ Thus, data can be replicated (optionally using XDCR Advancing Filtering), from o
...
+* The document meets whatever filtering criteria may have been configured.
+For information, See xref:clusters-and-availability/xdcr-filtering.adoc[XDCR Advanced Filtering].
 
-* If XDCR Advanced Filtering _is_ applied, each document in the source bucket is examined; but only those documents that meet the specified filtering-criteria are replicated.
+* The source collection within which the document resides can be mapped to a collection within the target bucket.
+For information, see xref:learn:clusters-and-availability/xdcr-with-scopes-and-collections.adoc#xdcr-with-scopes-and-collections[XDCR with Scopes and Collections].
 
-Once this initial process is complete, only _mutated_ documents will be considered for replication.
-_Mutated_ means one of the following:
+If, for a given document, one or both criteria are not satisfied, the document is _dropped_ from memory by XDCR, and therefore not replicated: however, the attempted replication of other documents is continued.
```
nelio2k: > is _dropped_ from memory

This phrasing may induce panic as it sounds like a document is going to be removed from a source cluster's memory (bucket).

Perhaps a more correct phrase could be along the lines of: 
"document is dropped from the XDCR replication pipeline"

modules/learn/pages/clusters-and-availability/xdcr-overview.adoc
```diff
@@ -110,33 +111,39 @@ Thus, data can be replicated (optionally using XDCR Advancing Filtering), from o
...
 
-* Modified
+* The document is _mutated_: which is to say, it is created, modified, deleted, or expired.
++
+Replication of a deleted or expired document means that the document will be correspondingly deleted or expired on the target.
+Note that this is the default behavior; although options are provided for _not_ replicating deletion or expiration mutations &#8212; so that the replicated documents are not removed.
+See the reference information for the CLI xref:cli:cbcli/couchbase-cli-xdcr-replicate.adoc[xdcr-replicate] command.
 
-* Deleted or expired
+* On the target bucket, a collection is created that allows a new, _implicit_ mapping to occur between a source collection and the new target collection: a _backfill pipeline_ is automatically created, to replicate documents between these collections, as appropriate.
```
nelio2k: 1. The mapping can be either implicit or explicit. Implicit is not the only option here. Perhaps a generic concept of mapping should be used instead of a limited scope of "implicit".
2. Technically, both backfill and main pipelines work in tandem (main pipeline replicate ongoing, while backfill catches up). Given this is an "overview" section, introducing a technical term "backfill pipeline" without context may cause more confusion.

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
```diff
@@ -0,0 +1,167 @@
+= XDCR with Scopes and Collections
+
+[abstract]
+When XDCR is established between a source bucket and a target bucket, data can be either _implicitly_ or _explicitly_ replicated between _collections_.
```
nelio2k: > _implicitly_ or _explicitly_ replicated between _collections_

Should it be "mapped" instead of replicated? 
Replication mechanism is the same. It's the mapping (routing) between collection that makes the difference.

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
```diff
@@ -0,0 +1,167 @@
...
+== Understanding XDCR Collections-Mapping
+
+XDCR allows only a single replication to be established between a specific source bucket and a specific target bucket.
+However, within the replication, documents can be _mapped_ between different source and target _collections_.
+(An overview of scopes and collections is provided in xref:learn:data/scopes-and-collections.adoc[Scopes and Collections].)
+
+A _mapping_ is a link, established by XDCR, between a _namespace_ within the source bucket and a namespace within the target bucket.
+A _namespace_ indicates the scope and collection, within the source or target bucket, within which exists data to be replicated.
+Each namespace is represented as the name of the scope, followed by the name of the collection, separated by a period.
+For example, if a bucket contains a scope named `HotelCollection`, and this scope contains collections named `USHotelCollection` and `UKHotelCollection`, then `HotelCollection.USHotelCollection` and `HotelCollection.UKHotelCollection` are both valid namespaces, specifying different collections within the same scope.
```
nelio2k: > if a bucket contains a scope named `HotelCollection`

"HotelScope"?

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
```diff
@@ -0,0 +1,167 @@
...
+
+. Every bucket contains a default scope and collection, whose namespace is `_default._default`.
+Implicit replication always occurs, therefore, between the default collection on the source bucket and the default collection on the target bucket.
+
+. In this example, the source bucket contains a scope named `ScopeA`, within which a collection named `CollectionA` exists.
+The target bucket also contains `CollectionA` within `ScopeA`.
+Therefore, since source and target buckets contain the identical namespace, `ScopeA.CollectionA`, implicit replication occurs automatically between `ScopeA.CollectionA` on the source, and `ScopeACollectionA` on the target.
++
+(Note that XDCR does not automatically create scopes or collections.
+Therefore, for an implicit mapping to be achieved, identical namespaces must have been previously created by the administrator.)
```
nelio2k: > previously created by the administrator

2 things:
1. Technically, either "previously" created by the admin and "later created" once the replication has been ongoing. Admins can technically create a missing target namespace after the replication has started, and it will _still_ be considered an "achievable implicit or explicit mapping".

2. One more important note that I'd like to see documented:
Given a target cluster is missing a namespace (i.e. source has scopeA.collectionC while target does not), an admin wants to now create a "scopeA.collectionC" on the target. Once he creates the namespace on the target, it will actually take the source XDCR up to 1 minute (default) to be able to detect the newly created collectionC on the target. It is not instantaneous.
Once XDCR detects it, _then_ XDCR will perform the backfill and also replicate any ongoing mutations over. Users will need to understand this concept: "source XDCR periodically checks the target for any changes in the way collections is set up", and know that the delay is not a defect.

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
```diff
@@ -0,0 +1,167 @@
...
+The target bucket also contains `CollectionA`, within `ScopeB`.
+Therefore, since source and target buckets contain the identical namespace, `ScopeB.CollectionA`, implicit replication occurs automatically between `ScopeB.CollectionA` on the source, and `ScopeB.CollectionA` on the target.
+
+. Within `ScopeB` in the source bucket, a collection named `CollectionB` exists.
+However, although the target bucket contains a scope named `ScopeB`, this does _not_ contain a collection named `CollectionB`: instead, it contains a collection named `CollectionC`.
+Therefore, since `ScopeB.CollectionB` is a namespace unique to the source, no implicit mapping is established with the target, and no replication is automatically initiated.
++
+Note, however, if an identical namespace is _subsequently_ established within the target bucket, a _backfill pipeline_ is automatically created, to replicate all appropriate data from the source to the target collection: this is described below, in xref:learn:collections-and-availability/xdcr-overview.adoc#target-collection-removal-and-addition[Target-Collection Removal and Addition].
```
nelio2k: > automatically created

> all appropriate data

automatically (and temporarily?) created for the purpose to replicate any previously dropped data

(Main pipeline, not mentioned here, is in charge of replicating the ongoing data)

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
```diff
@@ -0,0 +1,167 @@
...
+
+For the practical steps required to set up implicit mappings, see xref:manage:manage-xdcr/replicate-using-scopes-and-collections.adoc#replicate-data-between-collections-implicitly[Replicate Data Between Collections Implicitly].
+
+[#explicit-mapping]
+== Explicit Mappping
+
+_Explicit_ mapping is administrator-specified, and allows replication to occur between namespaces that are not identical.
```
nelio2k: > identical

identically named

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
```diff
@@ -0,0 +1,167 @@
...
+An explicit mapping can be specified either between scopes, or between collections:
+
+* _An explicit mapping between scopes_ ensures that an _implicit_ mapping occurs between collections within the scopes that are identically named.
```
nelio2k: > _An explicit mapping between scopes_ ensures that an _implicit_ mapping occurs between collections within the scopes that are identically named

Explicit mapping between scopes [without any specific collection mapping of their members] ensures...

(This is because if a specific collection mapping rule exists underneath, it would override the _implicit_ propety of the scope->scope rule)

nelio2k: Another point to make is that: An explicit mapping between scopes ensures that any future matching collection names on both sides are implicitly mapped.
This is not present when explicit collection mappings are used (i.e. s1.c1 -> s2.c1)
People may be unaware and wondering what's the difference is between the two.

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
```diff
@@ -0,0 +1,167 @@
...
+For example, if on the source, `ScopeA` contains `CollectionA`, and on the target, `ScopeB` contains `CollectionA`, an explicit mapping between `ScopeA` on the source and `ScopeB` on the target automatically produces an implicit mapping between the two collections named `CollectionA`.
++
+An explicit mapping between scopes produces no implicit mapping between dissimilarly named collections.
+For example, if on the source, `ScopeA` contains `CollectionA`, and on the target, `ScopeB` contains `CollectionB`, an explicit mapping between `ScopeA` on the source and `ScopeB` on the target produces no implicit mapping between `CollectionA` and `CollectionB`.
+
+* _An explicit mapping between collections_ allows a collection on the source to be mapped to a dissimilarly named collection on the target.
+Such collections may reside within dissimilarly named scopes.
+For example, an explicit mapping might be specified between `ScopeA.CollectionX` on the source, and `ScopeB.CollectionY` on the target.
+
```
nelio2k, line 87: One thing missing here is the concept of "denyList" (previously called blacklist) rule.
https://docs.google.com/document/d/1T7LglNmBx0LsupHXsoAME5oswgAHqCwJUcUNND07Hg0/edit#heading=h.lgmu5o1zjeoq

This may or may not be the right place to document this... but this needs to be documented somewhere

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
```diff
@@ -0,0 +1,167 @@
...
+The conditions under which a document is replicated from a source bucket to a target bucket are explained in xref:learn:clusters-and-availability/xdcr-overview.adoc#xdcr-process[XDCR Process].
+These include the existence of a valid collection-to-collection mapping, which may be any of the following:
+
+* The _implicit_ mapping that always exists between the `_default` collections of the source and target buckets.
+
+* The _implicit_ mapping that is automatically recognized between other identical namespaces within the source and target buckets.
+
+* An _explicit_ mapping that has been previously configured by the administrator; and which correctly corresponds to an existing pair of non-identical namespaces on the source and target buckets.
+
+If no such mapping exists for a given document, the document is not replicated.
```
nelio2k: Or if a denylist mapping rule exists for a given document, the document is not replicated.

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
```diff
@@ -0,0 +1,167 @@
...
+
+XDCR continuously monitors the target bucket for the addition or removal of collections.
```
nelio2k, line 243: The monitoring period is 1 minute by default, and is adjustable.

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
```diff
@@ -0,0 +1,167 @@
...
+Where collection-removal on the target bucket invalidates a mapping, documents previously eligible for replication are no longer so; and are therefore, on examination, dropped from memory by XDCR, and are not replicated.
+
+Where collection-addition occurs on the target bucket such that a new _implicit_ mapping is created, but occurs _after_ replication between the source and target bucket has been commenced, the following occur:
+
+* XDCR creates a _backfill pipeline_, which replicates to the target collection all documents from the source collection that were previously dropped by XDCR, due to the previous lack of an implicit mapping.
+The documents to be considered candidates for this replication are determined based on the _replication sequence number_ that XDCR was handling at the point the new implicit mapping was recognized: documents whose sequence number is lower than this are re-examined.
```
nelio2k: > _replication sequence number_

source document sequence number

modules/learn/pages/clusters-and-availability/xdcr-overview.adoc
tonyjhillman: Thanks. I've changed accordingly.

modules/learn/pages/clusters-and-availability/xdcr-overview.adoc
tonyjhillman: Okay. I've changed this to say "On the target bucket, a collection is created that allows a new mapping to occur between a source collection and the new target collection." I've thus left out reference to backfill here. Does that seem okay?

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
tonyjhillman: I've changed to "mapped".

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
tonyjhillman: Thanks! I've fixed.

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
tonyjhillman: For 1, I've changed to "identical namespaces must be created by the administrator".

tonyjhillman: For 2, I've modified here to say "Note, however, if an identical namespace is _subsequently_ established within the target bucket, this is eventually detected by XDCR, by means of a periodic check." Then, there's a pointer to the section Target-Collection Removal and Addition. In that section, it now says: "XDCR checks the target namespaces every minute: when a check is performed, any new collections that have been added to the target are detected.
(Note, therefore, that it may indeed take XDCR up to 60 seconds to detect a newly created collection on the target: detection is _not_ instantaneous.)"

BTW, shall I assume that detecting collection removal is by the same process, and will therefore also take up to 60 seconds?

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
tonyjhillman: I've added: "Note that once an explicit mapping between scopes has been established, implicit mapping will occur between identically named collections that are _subsequently created_ within those scopes.
(Note also that this only occurs when the explicit mapping is indeed between the scopes themselves: it does not occur when the explicit mapping is between individual collections within the scopes.)"

modules/learn/pages/clusters-and-availability/xdcr-overview.adoc
nelio2k: Yes, that's good.

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
nelio2k: > detecting collection removal

Just being explicitly clear, I think customers prob want to know about the XDCR behavior due to _target_ collection removal. (Source collection removal just means that no more data will be replicated since the source namespace is now gone). Anyway, the following is how the system is going to work in this scenario:

If a source collection has a mapped target collection, and there are ongoing mutations that are sent from the source collection, then XDCR will be able to detect the target collection removal as soon as it tries to send a mutation and finds out that the target collection is now gone. This detection mechanism does not depend on the per-minute interval and should be much quicker as it depends on the failure returned by the target cluster when sending a mutation.

If a mapped target collection is removed, but there is no mutation being sent to it, XDCR will detect the removal via the per-minute pull mechanism but will remain silent. 

(Essentially, it is replication-data driven)

tonyjhillman: I've added a note on source-collection removal as follows:
If a source collection has a mapped target collection, and replication from the source collection is ongoing; if the source collection is removed, this is detected by XDCR as soon as XDCR tries to send a mutation. 

The detection of source-collection removal, therefore, does not depend on the 60-second interval required for the detection of target-collection removal; and is likely to occur much more quickly.

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
tonyjhillman: I'm adding a section.

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
tonyjhillman: I've added a clause to state this.

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
tonyjhillman: I've added.

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
tonyjhillman: I've modified.

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
tonyjhillman: I've now added a section entitles Rules for Explicit Mappings.

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
```diff
@@ -98,6 +102,129 @@ However, as an alternative to an explicit mapping between `ScopeB` and `ScopeY`,
...
+[#priority-0]
+=== Priority 0
+
+A single, unique collection under a single unique scope on the source is mapped to a single, unique collection under a single unique scope on the target, and is affirmed for replication.
+For example, the expression `{"inventory.airport":"MyInventory.MyAirport"}` affirms that the collection `airport`, within the source-scope `inventory`, should be replicated to the collection `MyAirport`, within the target-scope `MyInventory`.
+
+A source collection is not permitted to be mapped (by means of multiple rules) to multiple target collections.
+For example, the expression `{"inventory.airport":"MyInventory.MyAirport","inventory.airport":"MyInventory.airport"}` generates an error.
+
+However, multiple source collections can be mapped (by means of multiple rules) to a single target collection.
```
nelio2k: Per design, we stated that:
> A single, unique collection under a single unique scope is mapped to a single, unique collection under a single unique scope on the target.

If this worked for you, then unfortunately it is a bug. It seems to be, as I am able to create a 2-to-1 rule like you mentioned.

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
```diff
@@ -98,6 +102,129 @@ However, as an alternative to an explicit mapping between `ScopeB` and `ScopeY`,
...
+For example, the expression `{"inventory.airport":"MyInventory.MyAirport", "inventory.MyAirport":"MyInventory.MyAirport"}` is valid, and affirms replication from the collections `airport` and `MyAirport`, located within the source-scope `inventory`, to the collection `MyAirport`, located within the target-scope `MyInventory`.
+
+If the replication of multiple source-collections to a single target collection results in conflicts, the procedure for xref:learn:clusters-and-availability/xdcr-conflict-resolution.adoc[XDCR Conflict Resolution] that has been established for the bucket is applied.
```
nelio2k: Once MB-44932 is fixed, we won't need this section.

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
tonyjhillman: I've changed to "At this point, a _backfill pipeline_ is automatically created, and is maintained for a temporary period; for the purpose of replicating any dropped data"

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
tonyjhillman: Thanks, I've fixed.

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
tonyjhillman: I've changed to say: _An explicit mapping between scopes_ ensures that an _implicit_ mapping occurs between identically named collections within those scopes.
For example, if on the source, `ScopeA` contains `CollectionA`, and on the target, `ScopeB` contains `CollectionA`, an explicit mapping between `ScopeA` on the source and `ScopeB` on the target automatically produces an implicit mapping between the two collections named `CollectionA`.
(Note, however, that this implicit mapping between the collections only occurs if no _explicit_ mapping of any of the collections is specified.)

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
tonyjhillman: I've changed this to read: Multiple source collections are not permitted to be mapped (by means of multiples rules) to a single target collection.
For example, the expression `{"inventory.airport":"MyInventory.MyAirport","inventory.MyAirport":"MyInventory.MyAirport"}` generates an error.

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
tonyjhillman: I've removed the paragraph about conflict resolution.

tonyjhillman:
I've added a subsection at the end, on migration rule syntax.

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
```diff
@@ -23,6 +23,7 @@ Between a keyspace on the source cluster and a keyspace on the target cluster, t
 The correct pairings of keyspaces on source and target must therefore be specified by the administrator.
 
 These forms of mapping are further described below.
+Note that whereas replication based on _implicit_ mapping is available in both Couchbase Server Enterprise and Community Edition, replication based on _explicit_ mapping is available only in Couchbase Server Enterprise Edition.
```
nelio2k: and "Migration rules feature"

tonyjhillman: I've modified this to: Note that whereas replication based on _implicit_ mapping is available in both Couchbase Server Enterprise and Community Edition, replication based on _explicit_ mapping and _migration rules_ is available only in Couchbase Server Enterprise Edition.

modules/learn/pages/clusters-and-availability/xdcr-with-scopes-and-collections.adoc
```diff
@@ -0,0 +1,310 @@
...
+For the practical steps, see xref:manage:manage-xdcr/replicate-using-scopes-and-collections.adoc#migrate-data-to-a-collection[Migrate Data to a Collection].
+
+[#rules-for-migration]
+=== Rules for Migration
+
+Each XDCR Migration is performed according to an administrator-specified rule.
+The rule must be expressed as a  key-value pair:
+
+* If _all_ the documents in the `&#95;default` collection of the source bucket are to be migrated to the specified target collection, the keyspace `_default._default` must be the key, and the destination keyspace must be the value.
+For example, the rule `{_default._default: California.SanFrancisco}` specifies that all documents in the `_default` collection of the source bucket should be migrated to the `SanFrancisco` collection, within the `California` scope, on the target bucket.
```
nelio2k, line 307: Just as a follow up since our last discussion with our PM...
We have decided that if such a rule is present (default.default : California.SanFrancisco), then it means that no other rule can exist. In other words, a rule such as (default.default : NewYork:NewYork) is not allowed if the SanFrancisco one is present, nor is a rule that has a filter expression.

tonyjhillman: I've added: Note that for a given replication, only one migration-rule can be used to migrate data from the _default collection on the source. If a second migration-rule, specifying a different target collection, attempts to migrate data from the same _default collection as does the first migration-rule, an error is generated. (This is the case regardless of whether a filter expression is specified for the second migration-rule.) 

(Do I have this right?)

nelio2k: That's correct! Thanks.

nelio2k (approved):

PR #1804, opened 2021-01-29
DOC-7936 Update desc when expiry pager intrvl = 0

amarantha-k:


amarantha-k:
Once this PR is approved, I will backport the changes to other versions.

modules/cli/pages/cbepctl/set-flush_param.adoc
```diff
@@ -112,10 +112,13 @@ The following are the command options:
 | `couch_response_timeout`
 | timeout in receiving a response from CouchDB.
 
+| `exp_pager_enabled`
+| Checks if the expiry pager task is enabled. The value is set to true by default.
```
JFlath: Is "Checks" the right word here (not sure if this has come from kv internal docs)? I think this should be "Sets" in terms of how it works...

JFlath (changes requested):

modules/cli/pages/cbepctl/set-flush_param.adoc
amarantha-k: Thanks. It should be "Sets"

PR #1805, opened 2021-02-05
DOC-8019: Update server examples with scopes and collections — N1QL query pages

simon-dew:
Updated pages:

* [Query: Fundamentals](https://docs-staging.couchbase.com/server/7.0/n1ql/query.html)
* [Queries and Results](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-intro/queriesandresults.html)
* [Getting System Information](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-intro/sysinfo.html)
* [ADVISE](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-language-reference/advise.html)
* [Advisor function](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-language-reference/advisor.html)
* [EXPLAIN](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-language-reference/explain.html)
* [FROM clause](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-language-reference/from.html)
* [GROUB BY clause](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-language-reference/groupby.html)
* [USE clause](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-language-reference/hints.html)
* [N1QL Language Reference](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-language-reference/index.html)
* [INFER](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-language-reference/infer.html)
* [JOIN clause](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-language-reference/join.html)
* [LET clause](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-language-reference/let.html)
* [LIMIT clause](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-language-reference/limit.html)
* [NEST clause](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-language-reference/nest.html)
* [OFFSET clause](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-language-reference/offset.html)
* [ORDER BY clause](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-language-reference/orderby.html)
* [PREPARE](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-language-reference/prepare.html)
* [SELECT clause](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-language-reference/selectclause.html)
* [SELECT Overview](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-language-reference/selectintro.html)
* [UNION, INTERSECT, and EXCEPT](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-language-reference/union.html)
* [UNNEST clause](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-language-reference/unnest.html)
* [WHERE clause](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-language-reference/where.html)
* [WITH clause](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-language-reference/with.html)
* [N1QL REST API](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-rest-api/index.html)

Docs issue: [DOC-8019](https://issues.couchbase.com/browse/DOC-8019)


modules/n1ql/pages/n1ql-language-reference/advise.adoc
```diff
@@ -395,7 +399,7 @@ Assume that the cost-based optimizer is enabled, and that statistics exist for t
 
 [source,n1ql]
 ----
-ADVISE SELECT * FROM `travel-sample` a WHERE a.type = 'hotel' AND a.country = 'France';
+ADVISE SELECT * FROM `travel-sample`.inventory.hotel a WHERE a.country = 'France';
```
osfameron, line 402: Ooo, this is really nice example to get head around advantage/difference of scope/collection usage.

modules/n1ql/pages/n1ql-language-reference/advise.adoc
```diff
@@ -483,7 +487,7 @@ ADVISE SELECT airportname FROM `travel-sample` WHERE type = 'airport' AND geo.al
 ====
 [source,n1ql]
 ----
-ADVISE SELECT * FROM `travel-sample` WHERE type LIKE 'air%'or type LIKE 'rou%';
+ADVISE SELECT * FROM `travel-sample`.inventory.landmark WHERE city LIKE "Par%" OR city LIKE "Lon%";
```
osfameron, line 490: This is an example where the type field -> scope translation loses some expressiveness, isn't it?

modules/n1ql/pages/n1ql-language-reference/let.adoc
```diff
@@ -104,10 +104,9 @@ Find all Sunday flights (`day = 0`) to the Charles De Gaulle airport (`CDG`) on
 [source,N1QL]
 ----
 SELECT t1.airline, t1.destinationairport, sch AS schedule
-FROM `travel-sample` AS t1
-LET sch = ARRAY v FOR v IN t1.schedule WHEN v.day = 0 END <1>
-WHERE type = "route"
-AND t1.destinationairport = "CDG"
+FROM `travel-sample`.inventory.route AS t1
+LET sch = ARRAY v FOR v IN t1.schedule WHEN v.day = 0 END -- <1>
```
osfameron, line 108: Ah, I'm guessing the SQL comment markers are for the benefit of people copy/pasting the sql from this source, rather than for Asciidoc itself?

modules/n1ql/pages/n1ql-language-reference/limit.adoc
```diff
@@ -1,4 +1,5 @@
 = LIMIT clause
+:imagesdir: ../../assets/images
```
osfameron, line 2: This was an unrelated driveby fix?

modules/n1ql/pages/n1ql-language-reference/nest.adoc
```diff
@@ -807,9 +804,9 @@ FROM _lhs-expression_
 
 There are three important changes in the index scan syntax example above:
 
-* `CREATE INDEX` on the `ON KEY` expression `route.airlineid` to access `route` documents using `airlineid` (which are produced on the left-hand side).
-* The `ON KEY route.airlineid FOR airline` enables N1QL to use the index `route.airlineid`.
-* Create any optional index, such as `route.airline` that can be used on `airline` (left-hand side).
+* `CREATE INDEX` on the `ON KEY` expression `rte.airlineid` to access `route` documents using `airlineid` (which are produced on the left-hand side).
```
osfameron, line 807: `rte` seems ugly abbreviation here, could the alias be named `route` to keep the example as it was before?

modules/n1ql/pages/n1ql-language-reference/orderby.adoc
```diff
@@ -153,33 +153,41 @@ ORDER BY city DESC, name ASC;
...
 [source,n1ql]
 ----
-SELECT name
-  FROM `travel-sample`
-  WHERE type = "landmark"
-UNION SELECT name
-  FROM `travel-sample`
-  WHERE type = "hotel"
-ORDER BY name ASC;
+SELECT name AS sort, type
```
osfameron, line 156: Ah, so does `type` still exist in the new sample?  

Not sure why the updated query brings in type, and renames name to sort.

osfameron: Aha, nice, you explain the rename just below!

modules/n1ql/pages/n1ql-language-reference/limit.adoc
simon-dew: This attribute is purely for the benefit of authors. It means you get to see the images when you preview the document as a standalone AsciiDoc file, e.g. in your browser or in GitHub. It is stripped out / ignored by Antora. It's not conventional to add this attribute but I find it useful.

modules/n1ql/pages/n1ql-language-reference/orderby.adoc
simon-dew: Yes, each collection still has the type field.

simon-dew (commented):
Thanks for the review @osfameron !

modules/n1ql/pages/n1ql-language-reference/selectclause.adoc
```diff
@@ -1,4 +1,5 @@
 = SELECT Clause
+:imagesdir: ../../assets/images
```
osfameron, line 2: One question: how are the images above created? The output in the preview looks good.

This commit has a lot of Asciidoc features I don't know enough to comment on, will study...

modules/n1ql/pages/n1ql-language-reference/selectintro.adoc
```diff
@@ -448,76 +441,76 @@ In this example:
...
-   },
-   {
-      "airlineid": "airline_24",
-      "callsign": "AMERICAN",
-      "sourceairport": "GGT"
-   }
+  {
+    "airlineid": "airline_24",
+    "callsign": "AMERICAN",
+    "sourceairport": "ABE"
```
osfameron, line 458: why have the source airports changed? new data?
Ah, because there's no sort order specified, and the implicit one has changed because of new PLAN?

modules/n1ql/pages/n1ql-language-reference/where.adoc
```diff
@@ -11,48 +12,43 @@ For example, list airports in France, detailed in <<examples_section,Example 1>>
...
+cond:: [Required] Conditional expression that represents a filter to be applied to the resultset.
+Records for which the condition resolves to TRUE are propagated to the resultset.
 
-bool_type::
-`AND`;; String expression of an inclusive filter to be applied to the resultset.
-
-`OR`;; String expression of a union filter to be applied to the resultset.
-
-`NOT`;; String expression of an exclusive filter to be applied to the resultset.
+You can construct complex conditional expressions, for example by using the xref:xref:n1ql-language-reference/logicalops.adoc[logical operators] `AND`, `OR`, and `NOT`.
```
osfameron, line 33: I like the simplification here.

modules/n1ql/pages/n1ql-language-reference/join.adoc
```diff
@@ -248,40 +246,39 @@ NOTE: The LEFT OUTER JOIN will list all left-side results regardless of matching
 
 [source,N1QL]
 ----
-SELECT DISTINCT  MIN(aport.airportname) AS Airport_Name,
-                 MIN(lmark.name) AS Landmark_Name,
-                 MIN(aport.tz) AS Landmark_Time
-FROM `travel-sample` aport
-RIGHT JOIN `travel-sample` lmark
+SELECT DISTINCT MIN(aport.airportname) AS Airport__Name,
```
osfameron, line 249: What's this double-underscore convention?

modules/n1ql/pages/n1ql-language-reference/join.adoc
```diff
@@ -665,11 +666,10 @@ This join hint instructs the planner to use nested-loop join (NL join) for the j
 [source,N1QL]
 ----
 SELECT COUNT(1) AS Total_Count
-FROM `travel-sample` rte
-INNER JOIN `travel-sample` aline
+FROM `travel-sample`.inventory.route rte
```
osfameron, line 669: I noticed in some of the other n1ql examples you omitted the alias name, and just used `route` which I preferred
(and which feels like a happy feature of the scope rewrite).

modules/n1ql/pages/n1ql-language-reference/join.adoc
```diff
@@ -1078,31 +1075,31 @@ LIMIT 4;
...
 [
   {
-    "airline": "VX",
-    "callsign": "REDWOOD",
-    "destinationairport": "SAN",
-    "name": "Virgin America",
+    "airline": "B6",
+    "callsign": "JETBLUE",
+    "destinationairport": "AUS",
+    "name": "JetBlue Airways",
```
osfameron, line 1081: again, it's slightly confusing to see data changes in a docs commit.

If this is just an artefact of not having ORDER, would it be worth adding an explicit ordering?
e.g. to at least minimize data changes in future?

(Obviously if the example is clearer without ORDER, then fair enough, is all trade-offs)

osfameron:
I've commented (mostly questions really). Looks good to me, to the extent I understand it, and I do think that the new documentation and examples seem cleaner, which feels like it validates the feature nicely 😄

modules/n1ql/pages/n1ql-language-reference/limit.adoc
osfameron: Nice! Yes, I already took advantage of that by clicking through to the github preview, so definitely useful!

modules/n1ql/pages/n1ql-language-reference/nest.adoc
osfameron: In fact from other examples later, it looks like simply omitting the alias would let you refer to it as `route`

modules/n1ql/pages/n1ql-language-reference/selectclause.adoc
simon-dew: [Railroad Diagram Generator](https://www.bottlecaps.de/rr/ui) by Gunther Rademacher. There's a standalone tool as well as the online converter.

modules/n1ql/pages/n1ql-language-reference/selectintro.adoc
simon-dew: When I tested the original query (with the default scope and collection) I got the new results. I get consistent results each time I run the new query. It could be that something has changed in the way the query engine generates an unsorted resultset. In a few cases, unfortunately, the results in the docs were just wrong. I'm testing them all as I updated them.

modules/n1ql/pages/n1ql-language-reference/join.adoc
simon-dew: I wondered this as well, I think it's just so that all the result field names have the same length and the values are aligned. 🤷‍♂️ left it as-is

modules/n1ql/pages/n1ql-language-reference/join.adoc
simon-dew: My approach was, if the original example applied an alias, leave it as is, even if unnecessary. However,  if the original example applied an alias which had the same name as the collection, e.g. `route` or `airline`, I omitted the alias in the rewrite. By default, the query service automatically applies an alias which is the same as the name of the keyspace. So no need to assign it explicitly.

modules/n1ql/pages/n1ql-language-reference/selectintro.adoc
osfameron: Great, yes, it's really nice to see how this fits in with the docs/QE work!

modules/n1ql/pages/n1ql-language-reference/join.adoc
simon-dew: This is one where (i) the original query gives different results in the new release, and (ii) the updated query gives different results again — but they're consistent each time I run it. So something's obviously changed in the query engine. Another good reason for automated testing!

The point about reproducibility of non-deterministic results is a good one. If an example isn't guaranteed to return the same results each time we should probably have a note explaining that. Or as you say add an ORDER BY clause to guarantee the same results.

modules/n1ql/pages/n1ql-language-reference/nest.adoc
simon-dew: The query actually says `ON KEY rte.airlineid`. Updated the explanation to match the query.

modules/n1ql/pages/n1ql-language-reference/let.adoc
simon-dew: Yes. Antora hides the comment markers and makes the callout number uncopyable. The comment markers are there so that the query engine does not trip over the callout number during automated testing.

amarantha-k (approved):

simon-dew:
All docs staging site links updated

PR #1816, opened 2021-02-16
DOC-5322: Note about FTS auto-upgrade to Scorch

eric-schneider:
This PR seeks to update the 6.0 release notes and upgrade docs to clarify an edge case where indexes can be automatically upgraded to Scorch.

### DIFF
--- a/modules/install/pages/upgrade-strategy-for-features.adoc
+++ b/modules/install/pages/upgrade-strategy-for-features.adoc
@@ -4,7 +4,7 @@ Upgrading to the latest version of Couchbase Server allows you to leverage the l
 The following table summarizes the new features in this release, and describes what parts of the cluster need to be upgraded to enable them.
 
 .Couchbase Server 6.0 Feature Upgrade Matrix
-[cols="1,1,1,1,1,1"]
+[cols="1,1,1,1,1,2"]
 |===
 | Feature | Upgrade Level | xref:upgrade-strategies.adoc[Upgrade Type] | xref:upgrade.adoc#supported-upgrade-paths[Upgrade Path] | Mixed Mode<<feature-mixed-mode,†>> | Dependencies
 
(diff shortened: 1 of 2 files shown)


abhinavdangeti (approved):

modules/install/pages/upgrade-strategy-for-features.adoc
```diff
@@ -24,15 +24,18 @@ If you plan to use the production release of Couchbase Analytics in version 6.0,
...
 
-| Scorch indexing system (Full Text Search)
+| Scorch indexing system (Search Service)
 | Cluster
 | Offline/Online
 | Direct
 | Yes*
-| * FTS nodes running version 6.0 use the Scorch indexing system by default for new indexes.
-You can continue to use previous indexes in version 6.0, but they will not use Scorch.
+a| *After upgrading to Couchbase Server 6.0, existing Full Text Indexes will continue to use the Moss xref:fts:fts-creating-indexes.adoc#index-type[index type], while _new_ Full Text Indexes will use the Scorch index type by default.
```
osfameron, line 32: Question: what's the `a|` syntax here?

eric-schneider: @osfameron Good question. The `a` is the [cell specifier](https://docs.asciidoctor.org/asciidoc/latest/tables/add-cells-and-rows/#specifiers); more specifically the [style operator](https://docs.asciidoctor.org/asciidoc/latest/tables/format-cell-content/). While I was drafting this change, I originally had a NOTE admonition in this cell. Since an admonition is a block, I needed to add an `a` in front of the cell separator for the block to render properly in the cell. However, I ended up removing the admonition but seemed to have left behind the cell specifier. It's harmless in this case and not really worth removing.

A note about adding the `a` to a cell: It truly turns the cell into a mini asciidoc document within the page. It allows you to treat the content like you would any other content outside of the table, but causes some funny issues with certain asciidoc features. For example, if the cell contains a footnote reference, the footnote gets rendered at the bottom the cell, not the page.

osfameron: Ah! Thanks for that. Searching the docs for `a|` wasn't getting me anywhere...

amarantha-k (approved):

PR #1858, opened 2021-03-09
DOC-6994 add note about RHEL 8.2 on v6.5

amarantha-k:


malarky (commented):
Would it be better to explicitly say "8.0, 8.1" for clarity, the same way that the 6.6 docs currently say? Still with the footnote though, think that's needed.

PR #1867, opened 2021-03-17
DOC-8190: Updated description of docs_written

tonyjhillman:


amarantha-k:
@nelio2k and @tonyjhillman This PR has been open for a long time. Please let me know if there any further changes expected. If not, I'll plan to merge by 07/15.

modules/rest-api/pages/rest-xdcr-statistics.adoc
```diff
@@ -187,6 +187,10 @@ The above command produces the following output that shows that XDCR transferred
 1000000,1000000,1000000,1000000,1000000,1000000,1000000]}}
 ----
 
+The output shows that `60` samples were taken.
+A sample was taken every `1000` milliseconds, and each sample is represented by its `timestamp`.
+1 million documents are shown to have been written in each sampling window.
```
nelio2k: It's not that during the duration of each sampling window, 1 million documents have been written in each window.
It is that there is a cumulative of 1 million document already written by the 1st sampling window (in the output). And, during each successive window, the number of document remained constant at 1 million. In other words, since the first window has began, where 1 million docs have been written already, there has been no increase in the total of number of documents written.

nelio2k (commented):
Apologies for the late review as this fell through the cracks

modules/rest-api/pages/rest-xdcr-statistics.adoc
tonyjhillman: I've written: "1 million documents are shown already to have been written when the first sample was taken; and this number is shown to have remained consistent at the time of each successive sample." Is this better?

nelio2k (approved):
Thanks! Looks great

PR #1879, opened 2021-03-25
DOC-7831: Description of schedule merge window-definition

tonyjhillman:


cg14823 (approved):
Yes this looks correct and the examples with dates makes it very clear, thank you.

PR #1880, opened 2021-03-25
DOC-8379: Note on avoiding task overlap in backup service

tonyjhillman:


modules/learn/pages/services-and-indexes/services/backup-service.adoc
```diff
@@ -83,6 +83,14 @@ Optionally, the data itself can be retained, on the local filesystem.
 A deleted repository whose data still exists can be _imported_ back into the cluster, if required.
 Once imported, the repository can be _read_ from, but no longer receives data backups.
 
+[#avoiding-task-overlap]
+== Avoiding Task Overlap
+
+Although the Backup Service allows automated tasks to be scheduled for repetition at intervals as small as one minute, administrators are recommended typically not to lower the interval below fifteen minutes; and always to ensure that the established interval allows each scheduled task-repetition ample time to complete before the next; even in the event of unanticipated, high network-latency.
+
+If, due to an interval-specification that is too small, a given task attempts to start while its prior repetition is still ongoing, the new task fails, stops running, and is not restarted.
```
owend74: If, due to an interval-specification that is too small, a given task attempts to start while its prior repetition is still ongoing, the new task fails to start.

modules/learn/pages/services-and-indexes/services/backup-service.adoc
```diff
@@ -83,6 +83,14 @@ Optionally, the data itself can be retained, on the local filesystem.
...
+In such circumstances, all data continues to be backed up by the prior repetition; and will be backed up again by any subsequent repetition that starts after the completion of the prior repetition.
```
owend74: In such circumstances, all data continues to be backed up by the prior repetition; and will be backed up again by any subsequent repetition that starts after the completion of the previous backup.

owend74 (changes requested):
I got a little lost in repetitions :)
Not sure I improved - so treat just suggestions..

@cg14823 probably worth you reading through.

modules/learn/pages/services-and-indexes/services/backup-service.adoc
tonyjhillman: ok.

modules/learn/pages/services-and-indexes/services/backup-service.adoc
tonyjhillman: ok

modules/learn/pages/services-and-indexes/services/backup-service.adoc
```diff
@@ -83,6 +83,14 @@ Optionally, the data itself can be retained, on the local filesystem.
 A deleted repository whose data still exists can be _imported_ back into the cluster, if required.
 Once imported, the repository can be _read_ from, but no longer receives data backups.
 
+[#avoiding-task-overlap]
+== Avoiding Task Overlap
+
+Although the Backup Service allows automated tasks to be scheduled for repetition at intervals as small as one minute, administrators are recommended typically not to lower the interval below fifteen minutes; and always to ensure that the interval is large enough to allow each scheduled task-repetition ample time to complete before the next is due to commence; even in the event of unanticipated network latency.
```
cg14823: I think `repetition` may be a bit confusing in some of the places I would suggest altering the following

> Backup Service allows automated tasks to be scheduled for repetition at intervals as small as one minute

To

> Backup Service allows automated task to be scheduled at intervals as small as one minutes

cg14823 (commented):
I think it looks good but maybe the word repetition can be a bit confusing in some context maybe iteration or prior instance could be replaced in some of the places.

modules/learn/pages/services-and-indexes/services/backup-service.adoc
tonyjhillman: The reason I used 'repetition' is that I imagined we are talking about the same task, scheduled to be repeated at intervals?.But perhaps that is not correct. If *different* tasks overlap, does this also produce failure of this kind?

cg14823 (approved):

PR #1882, opened 2021-03-31
DOC-8397: New REST API statistics retrieval

tonyjhillman:


modules/rest-api/pages/rest-statistics-single.adoc
```diff
@@ -0,0 +1,487 @@
+= Getting a Single Statistic
+
+[abstract]
+The REST API allows individual cluster-statistics to be retrieved, from one or more nodes.
+
+[#http-methods-and-uris]
+== HTTP Method and URI
+
+----
+GET /pools/default/stats/range/<metric_name>/<function-expression>
```
dave-finlay, line 10: We should indicate that the function expression is optional.

modules/rest-api/pages/rest-statistics-single.adoc
```diff
@@ -0,0 +1,487 @@
...
+[#curl-syntax]
+== Curl Syntax
+
+----
+curl -X GET -u <username:password> \
+http://<ip-address-or-domain-name>:<port-number>/pools/default/stats/range/<metric-name>/
+  <function-expression>
+----
+
+The `function-expression` must take the following form:
```
dave-finlay, line 28: Suggest: "When present the function expression should take the following form ...."

modules/rest-api/pages/rest-statistics.adoc
```diff
@@ -0,0 +1,24 @@
...
+The REST API allows cluster statistics to be retrieved; either individually, or in bulk.
+
+== Methods and URIs
+
+[cols="76,215,249"]
+|===
+| HTTP method | URI | Description
+
+| GET
+| `/pools/default/stats/range/<metric_name>/<function-expression>`
```
dave-finlay, line 13: Suggest indicate the function-expression and labels are optional, that is,

/pools/default/stats/range/<metric_name>/[<function-expression>[label=value&...]]

modules/rest-api/pages/rest-statistics-single.adoc
```diff
@@ -0,0 +1,487 @@
...
+=== `function`
+
+The `function` parameter can be specified as follows:
+
+----
+<function> := rate | irate | increase | avg_over_time | min_over_time | max_over_time | deriv | delta | idelta | sum | min | max | avg | p[1-99]
+----
+
+* `rate`.
+The per-second average rate of increase of the time series in the range vector.
```
dave-finlay, line 46: I think we should tell the user that the metrics for each node are stored in an instance of Prometheus which runs on the node. In this way, we can then refer the user to the Prometheus concepts. We might want to explain this conceptual idea (that we use Prometheus to store stats) somethere other than in the stats API reference, though. 

In any case, I might say something like this to describe rate. 

The per-second average rate of increase of the time series over the time window used when retrieving the time series. If the time window is not explicitly specified, it default to 1m. (See the description of the timeWindow optional parameter.) The rate function exposed by this API is the same as the Prometheus rate function.

Something like this anyway.

modules/rest-api/pages/rest-statistics-single.adoc
```diff
@@ -0,0 +1,487 @@
...
+Breaks in monotonicity (such as counter resets due to target restarts) are automatically adjusted for.
+Also, the calculation extrapolates to the ends of the time range, allowing for missed scrapes or imperfect alignment of scrape cycles with the range's time period.
+
+* `irate`.
+The per-second instant rate of increase of the time series in the range vector.
+This is based on the last two data points. Breaks in monotonicity (such as counter resets due to target restarts) are automatically adjusted for.
```
dave-finlay, line 52: I would just say:

The per-second instant rate of increase of the time series in the range vector.
This is based on the last two data points. 

And then repeat the idea that this is the same as the Prometheus irate fucntion.

modules/rest-api/pages/rest-statistics-single.adoc
```diff
@@ -0,0 +1,487 @@
...
+
+* `increase`.
+The increase in the time series in the range vector.
```
dave-finlay, line 55: As before. I would just say:
The increase in the time series in the range vector.

And then say that this is the same as the Prometheus increase function.

modules/rest-api/pages/rest-statistics-single.adoc
```diff
@@ -0,0 +1,487 @@
...
+
+* `avg`.
+The average of values retrieved over the sample period.
+
+* `p[1-99]`
+The number of results to be displayed per page.
+
+=== `label` and `value`
+
+The possible specifications of `label`, and their associated possible values, are as follows.
```
dave-finlay, line 95: I think we should probably divide the optional parameters into 2 groups.

(1) the fixed optional parameters: start, end, step, nodes, nodesAggregation, timeWindow, timeout and alighTimestamps.
(2) labels that are in the underlying metrics (such as bucket, scope, collection, ...) -- basically all the labels that show in the metrics reference.

dave-finlay (commented):
Hi Tony - just got through an initial review. Have a look - I will do a second pass later.

modules/rest-api/pages/rest-statistics-single.adoc
tonyjhillman: I have done so, and so have used square brackets rather than angle.

modules/rest-api/pages/rest-statistics-single.adoc
tonyjhillman: Okay, I've fixed.

modules/rest-api/pages/rest-statistics-single.adoc
tonyjhillman: I've added a note to each description, stating that metrics are stored in the node's Prometheus instance. I've redone the rate description as per the above.

modules/rest-api/pages/rest-statistics-single.adoc
tonyjhillman: Done.

modules/rest-api/pages/rest-statistics-single.adoc
tonyjhillman: Done.

modules/rest-api/pages/rest-statistics-single.adoc
tonyjhillman: Okay I have called them 'General' and 'Couchbase-Specific'. Does that work?

tonyjhillman:
This PR is now closed: I've put the above review feedback into a new PR, which is https://github.com/couchbase/docs-server/pull/2022

PR #1884, opened 2021-04-02
DOC-8058 operators collections updated.

sixthcodebrewer:


sixthcodebrewer:
### Need Help
```
SELECT DISTINCT city FROM `travel-sample`.inventory.landmark AS l
WHERE EXISTS (SELECT city FROM `travel-sample`.inventory.airport AS a );
```
 This query from collectionsops page results in the 626 cities that contain both an airport and a landmark but shows a different result with respect to the ordering of the city with an internal error. 

- Attaching the image of the query and the result 

query result
error message

modules/n1ql/pages/n1ql-language-reference/collectionops.adoc
```diff
@@ -424,8 +424,8 @@ Return Values:: The `EXISTS` operator evaluates to `TRUE` if the value is an arr
 *EXISTS Example 1:* Of the 1641 cities with an airport, search for all cities that have a landmark as well.
```
simon-dew:
```suggestion
*EXISTS Example 1:* Of the 274 cities with a hotel, search for all cities that have hotels with reviews.
```

modules/n1ql/pages/n1ql-language-reference/collectionops.adoc
```diff
@@ -424,8 +424,8 @@ Return Values:: The `EXISTS` operator evaluates to `TRUE` if the value is an arr
...
 
 ----
-SELECT DISTINCT city FROM `travel-sample` AS l WHERE type = "landmark"
-AND EXISTS (SELECT city FROM `travel-sample` AS a WHERE type = "airport");
+SELECT DISTINCT city FROM `travel-sample`.inventory.landmark AS l
+WHERE EXISTS (SELECT city FROM `travel-sample`.inventory.airport AS a );
```
simon-dew:
```suggestion
SELECT DISTINCT h.city FROM `travel-sample`.inventory.hotel AS h
WHERE EXISTS h.reviews;
```

modules/n1ql/pages/n1ql-language-reference/collectionops.adoc
```diff
@@ -424,8 +424,8 @@ Return Values:: The `EXISTS` operator evaluates to `TRUE` if the value is an arr
...
 ----
 
 This results in the 626 cities that contain both an airport and a landmark.
```
simon-dew:
```suggestion
This results in 255 cities that contain hotels with reviews.
```

simon-dew: Please also replace the query results (lines 434 to 446) with the updated results:
```
[
  {
    "city": "Medway"
  },
  {
    "city": "Giverny"
  },
  {
    "city": "Glasgow"
  },
  {
    "city": "Highland"
  },
  ```

simon-dew (changes requested):
Thanks very much for the contribution! Great work. Regarding the query that you flagged as problematic — I'm not sure what this query was trying to do. (Not your fault, I know.) As originally written, the EXISTS expression in this query always returns `true`, which means that the query would just return the 626 distinct cities in the `landmark` collection, even if the subquery didn't generate an error. In fact, it would be much more natural to find the data that this example is supposed to find using a JOIN.

I've therefore made some suggestions to replace this example with a new one using the `hotel` collection. Please note you will also have to update the result for this query — I've added the result as a comment. I couldn't make a suggestion for the query results because you didn't edit this part of the document.

All other changes look good to me 👍

simon-dew:
> with an internal error

This is a known issue 😄

sixthcodebrewer:
> > with an internal error
> 
> This is a known issue

Ok 😃

sixthcodebrewer:
> Thanks very much for the contribution! Great work. Regarding the query that you flagged as problematic — I'm not sure what this query was trying to do. (Not your fault, I know.) As originally written, the EXISTS expression in this query always returns `true`, which means that the query would just return the 626 distinct cities in the `landmark` collection, even if the subquery didn't generate an error. In fact, it would be much more natural to find the data that this example is supposed to find using a JOIN.
> 
> I've therefore made some suggestions to replace this example with a new one using the `hotel` collection. Please note you will also have to update the result for this query — I've added the result as a comment. I couldn't make a suggestion for the query results because you didn't edit this part of the document.
> 
> All other changes look good to me 

Thank You, Simon, I have committed all the suggested changes.

simon-dew (approved):
Fantastic, thanks. LGTM 👍

sixthcodebrewer:
> Fantastic, thanks. LGTM

Thank You 🕺🏻 🙌🏻

PR #1897, opened 2021-04-10
Remove repeated information about bucket

avivmu:
The removed information appears just in the paragraph above it, so it's redundant to repeat it


amarantha-k:
Closing without merge as we'll plan to revise the statement with additional information in the near future.

PR #1900, opened 2021-04-13
Add new 6.6.2 capabilities, notations, and scriptlets

jon-strabala:
This is my first pass at adding the 6.6.2 changes to the Eventing documents
I could add pictures if needed, but I think it should be good at this point.


amarantha-k (approved):
Approving after a quick look. Will run a editorial review before merging.

PR #1920, opened 2021-04-22
DOC-8472: Updated rest info for getting bucket information, and conso…

tonyjhillman:
…lidated the two old rest pages into one


tonyjhillman:
Also made small edit to bucket-streamingURI page.

PR #1925, opened 2021-04-27
DOC-8491: Description of metadata management

tonyjhillman:


modules/learn/pages/clusters-and-availability/hard-failover.adoc
```diff
@@ -25,7 +25,7 @@ For further information on initiating hard failover, see xref:manage:manage-node
 [#default-and-unsafe]
 == Hard Failover in Default and Unsafe Modes
 
-In the event of a cluster’s xref:learn:clusters-and-availability/cluster-manager.adoc#master-services[Master Services] not being able to contact a majority of the cluster’s nodes, an attempted hard failover will, by default, _not_ be executed.
+In the event of a cluster’s xref:learn:clusters-and-availability/cluster-manager.adoc#master-services[Master Services] not being able to contact a majority of the cluster’s nodes (sometimes referred to as a _quorum failure_), an attempted hard failover will, by default, _not_ be executed.
```
dave-finlay: At some point, we should change our terminology and move away from using terms like "master" in our docs (and code). This is a task that we should talk about separately, but just noting it for the record here.

dave-finlay: > an attempted hard failover will, by default, _not_ be executed.

Suggest we adjust to say: "an attempted failover will not be executed."

If the orchestrator can't contact a majority of the nodes, no kind of failover can be performed.

modules/learn/pages/clusters-and-availability/metadata-management.adoc
```diff
@@ -0,0 +1,127 @@
+= Metadata Management
+
+:description: pass:q[A Couchbase Cluster's _metadata_ describes its configuration. Some categories of metadata are maintained by a _consensus_ methodology, based on the Raft algorithm; others by the _Master Services_ of the Cluster Manager. ]
```
dave-finlay: We need ways to refer to the new consensus-protocol based metadata management system introduced in 7.0 and the legacy gossip-replicated metadata management system. 

For the new system, we could refer to it as:
* the consensus-protocol based metadata management system
* the Raft-based metadata management system
* Chronicle

It might be worth introducing the term Chronicle, because it's much more compact than the other two alternatives.

For the old system, we could say:
* the legacy metadata management system
* the gossip-replicated metadata management system
* ns_config

I'd prefer not to use "ns_config", since it includes the letters "ns", which refer to Northscale, the company's first name. But if it makes things a lot simpler, maybe it's a good idea.

modules/learn/pages/clusters-and-availability/metadata-management.adoc
```diff
@@ -0,0 +1,127 @@
...
+
+[abstract]
+{description}
+
+[#understanding-couchbase-server-metadata]
+== Understanding Couchbase-Server Metadata
+
+In Couchbase Server 7.0+, metadata is considered to be of two kinds: metadata related to cluster topology; and all other metadata.
```
dave-finlay: We shouldn't say that metadata is "of two kinds" in public documentation. I know I've said something like this in the past, but this was more to motivate that the topology-related metadata was the first metadata to move to Chronicle. I would just rework this section to say that in 7.0 we store some metadata in Chronicle and some that we don't.

modules/learn/pages/clusters-and-availability/metadata-management.adoc
```diff
@@ -0,0 +1,127 @@
...
+[#all-other-metadata]
+=== All Other Metadata
+
+All other metadata for the cluster is maintained by _ns_sever_.
+This metadata includes:
+
+* Compaction settings.
+* Security settings.
+* XDCR replications.
+* Settings and metadata for all services.
```
dave-finlay, line 28: Add: Per-node configuration settings

modules/learn/pages/clusters-and-availability/metadata-management.adoc
```diff
@@ -0,0 +1,127 @@
...
+
+Every node in the cluster stores a copy of all such metadata.
+Metadata-changes are asynchronously replicated across the nodes by the xref:learn:clusters-and-availability/cluster-manager.adoc#master-services[Master Services] of the Cluster Manager: this leadership role is occupied by one node at a time; with a replacement node being elected if the current leader has become non-communicative.
```
dave-finlay: Only the gossip-replicated configuration is "asynchronously replicated through the cluster". I think you understand this, but the way it reads like this is true for all metadata, so it might need to be reworded a bit.

And actually this activity is not managed by the orchestrator / master node. What happens is that whenever a change happens on any node it will try to replicate this change to all other nodes in the cluster. In addition to this, nodes periodically pull other another random node's configuration. So over time configuration spreads throughout the cluster.

modules/learn/pages/clusters-and-availability/metadata-management.adoc
```diff
@@ -0,0 +1,127 @@
...
+
+A leader can only transmit metadata-changes when it is granted a _quorum lease_; which means that a majority of nodes in the cluster are continuing to communicate with it.
```
dave-finlay: As mentioned in my previous comment, ns_config based configuration is replicated by all nodes independently at all times. If there's a partition, then of course configuration can't be replicated across the partition, but replication will resume if the partition is healed. 

Leases are used by the orchestrator (the master node) when it's doing some cluster-wide activity like rebalance or failover. This is to ensure that another master node doesn't stand up and start trying to do cluster-wide activities at the same time.

modules/learn/pages/clusters-and-availability/metadata-management.adoc
```diff
@@ -0,0 +1,127 @@
...
+Should at least half of the cluster's nodes become non-communicative, due to an unanticipated network partition, the leader can make no further changes to metadata until the partition is healed, or the cluster is otherwise fixed.
```
dave-finlay: In 6.6 and earlier (i.e. before Chronicle) this is kind of true, but it's also kind of the reverse. 

What I mean is that anyone can make any changes to ns_config based metadata anytime. It's just that it may not be safe to do so if another leader is trying to make changes to the same metadata at about the same time. So, what happens in 6.6 and earlier is that the orchestrator grabs the lease, allowing it to change the metadata confident in the knowledge that it's the only entity changing the metadata at that time.

In 7.0 under Chronicle, it is the case that the orchestrator can make no changes to the metadata if half or more of the nodes in the cluster can't be reached. This is guaranteed by the Raft consensus protocol. You would think that we don't need leases anymore, but it's the case that we are not at the point where they can be removed so we still use them. At some point in the future, leases may go away. 

So, this sentence is correct in 7.0 as it applies to the orchestrator and Chronicle, but is not accurate as it relates to ns_config. As it appears to be in a section on ns_config, it's a bit confusing.

modules/learn/pages/clusters-and-availability/metadata-management.adoc
```diff
@@ -0,0 +1,127 @@
...
+(See xref:learn:clusters-and-availability/hard-failover.adoc#default-and-unsafe[Hard Failover in Default and Unsafe Modes], for information on failing over a cluster in such circumstances.)
+
+The asynchronous replication of metadata-changes by the Master Services means that updates occur across the cluster with _eventual consistency_.
+
+For a more detailed description of the cluster manager, see xref:learn:clusters-and-availability/cluster-manager.adoc[Cluster Manager].
+
+[#consensus-based-metadata-management]
+== Consensus-Based Metadata-Management
+
+Topology data for the Couchbase-Server cluster &#8212; referring to nodes, services, buckets, scopes and collections, and vBuckets &#8212; is maintained by a consensus methodology, based on the https://raft.github.io/[Raft^] algorithm.
```
dave-finlay: I would avoid the idea that only topology information is stored in Chronicle. As we go forward this will change. Ultimately we want to move everything to Chronicle. 

In any case, in this section I wouldn't mention what configuration we store in Chronicle and leave that to elsewhere.

modules/learn/pages/clusters-and-availability/metadata-management.adoc
```diff
@@ -0,0 +1,127 @@
...
+Basic concepts are summarized below.
```
dave-finlay: I would say that what follows is just a sketch of how Chronicle works. For full details on how Raft can be used to build a replicated log and on things like leadership election, refer readers to the Raft paper. 

I would also say that users don't really need to know all the details below, but that we're providing it as they may be interested to read about it. What users mostly need to know is that the Chronicle is strongly consistent, supports full linearizability and we have tested it with Jepsen. (https://jepsen.io/)

modules/learn/pages/clusters-and-availability/metadata-management.adoc
```diff
@@ -0,0 +1,127 @@
...
+
+* Add a key-value pair.
+* Update the value of a key-value pair.
+* Update the value of a key-value pair, based on logical constraints.
+* Update the values of key-value pairs transactionally.
+
+Each key-value pair has a revision number that corresponds to the sequence of change-commands applied to the store.
+
+[#quorum-failure]
+=== Quorum Failure
```
dave-finlay, line 114: I think the section you have here is probably more or less fine. 

The section on unsafe failover we'll definitely need to update as it has changed behavior from 6.6 in the following way:
* 6.6: after the unsafe failover the nodes are still in the cluster (but in a failed-over state) and can be recovered. In this way unsafe failover is just like a regular failover, the only difference between the two is that the system cannot guarantee safety because a majority of nodes can't be reached to get a lease.
* 7.0: after the unsafe failover the failed-over nodes are immediately removed from the cluster. However, in addition these nodes are not contacted to inform them that they've been removed - they still think they are in the cluster. So unsafe failover is very different from regular failover.

modules/learn/pages/clusters-and-availability/hard-failover.adoc
```diff
@@ -36,9 +36,17 @@ This protects the cluster from the situation where:
 
 Note that this restriction also protects the cluster in situations where multiple xref:learn:clusters-and-availability/groups.adoc[Server Groups] are accidentally divided into separate sets, in the same way.
 
-However, if an administrator is confident that simultaneous hard failovers will not occur, a single hard-failover can be performed in _unsafe_ mode; this is executed even when a majority of cluster-nodes cannot be contacted.
+[#performing-an-unsafe-failover]
+=== Performing an Unsafe Failover
```
dave-finlay, line 40: This section needs a bit of work. 

The first idea is since Chronicle is a consensus protocol that achieves strong consistency by synchronously replicating configuration to a majority of the nodes in the cluster _and_ since topology-related metadata is stored in Chronicle, if a majority of nodes in the cluster become unavailable for some reason, then no topology related changes can be made. This means: no rebalance, no failover, no node addition, no bucket creation / deletion.

So essentially: if you lose a majority of nodes, your cluster becomes stuck. There is no way to make topology related changes within the consensus protocol. 

To allow customers recover clusters where a majority of nodes are lost, we allow unsafe failovers to be made in these situations. Unsafe failovers don't work through the consensus protocol, they simply force the metadata changes and override the protocol. This means that strong consistency can no longer be guaranteed - and in particular metadata may be lost. 

Unsafe failovers have the following requirements:
* any node that will survive the failover must be up and reachable by the other nodes that will survive the failover
* the administrator should attempt to ensure that the nodes being unsafely failed over are permanently down and will not restart and resume communication with the cluster. These nodes cannot be recovered and they will need to be re-initialized. If the administrator can get to these nodes they may want to preserve the data files in the event they need to use a tool like cbtranfer to manually copy the data back to the restored cluster.
* the administrator should ensure that no other administrator is simultaneously performing a separate unsafe failover. Only one should be run at a time. 
* administrators should attempt to a safe (i.e. regular failover) first and only if this doesn't succed consider the unsafe failover option.

A vital point is that unsafe failover in 7.0 is different from that in 6.6. In 7.0, after the unsafe failover the failed over nodes are permanently removed from the cluster, cannot be recovered and need to be re-initialized. More on this in a later comment.

modules/learn/pages/clusters-and-availability/hard-failover.adoc
tonyjhillman: Okay. I've left this for now, since I'll need to look at multiple pages to make this change.

tonyjhillman: I've changed accordingly.

modules/learn/pages/clusters-and-availability/hard-failover.adoc
tonyjhillman: I have rewritten this section. Hopefully, I've included all of these points.

modules/learn/pages/clusters-and-availability/metadata-management.adoc
tonyjhillman: I've used 'Chronicle' and 'legacy metadata-management system'.

modules/learn/pages/clusters-and-availability/metadata-management.adoc
tonyjhillman: I've now attempted to make the distinction in this way.

modules/learn/pages/clusters-and-availability/metadata-management.adoc
tonyjhillman: Have added.

modules/learn/pages/clusters-and-availability/metadata-management.adoc
tonyjhillman: I've added and clarified accordingly.

modules/learn/pages/clusters-and-availability/metadata-management.adoc
tonyjhillman: I've updated the description, hopefully to make this more clear.

modules/learn/pages/clusters-and-availability/metadata-management.adoc
tonyjhillman: I've added a short introduction, saying this.

modules/learn/pages/clusters-and-availability/metadata-management.adoc
tonyjhillman: I've attempted to emphasize this appropriately.

tonyjhillman:
Replaced by https://github.com/couchbase/docs-server/pull/1985

PR #1929, opened 2021-04-29
DOC-7889 transactions updates

RichardSmedley:
DOC-7846 / CBSE-9312
DOC-8139

### DIFF
--- a/modules/learn/pages/data/transactions.adoc
+++ b/modules/learn/pages/data/transactions.adoc
@@ -16,7 +16,7 @@ xref:java-sdk:howtos:distributed-acid-transactions-from-the-sdk.adoc[Java].
 
 _Distributed ACID Transactions_ are operations that ensure that when multiple documents need to be modified such that only the successful modification of all justifies the modification of any, either all the modifications do occur successfully; or none of them occurs.
 This _Atomicity_ supports _insert_, _update_, and _delete_ operations, across any number of documents.
-It also enables seemless integration between KV and xref:n1ql:n1ql-language-reference/begin-transaction.adoc#purpose[Query] operations within transactions.
+It also enables seamless integration between KV and xref:n1ql:n1ql-language-reference/begin-transaction.adoc#purpose[Query] operations within transactions.
 
 Only following commit of the documents modified by the transaction -- when the modified data-copy is removed from the meta data area (which uses xref:learn:data/extended-attributes-fundamentals.adoc[Extended Attributes]), and is written over the main body of the document’s data -- do the corresponding changes become readable by other transactions and operations.
 This is https://jepsen.io/consistency/models/read-committed[Read Committed^] isolation.
--- a/modules/learn/pages/data/transactions.adoc
+++ b/modules/learn/pages/data/transactions.adoc
@@ -72,7 +72,7 @@ This setting can be changed via xref:manage:manage-buckets/create-bucket.adoc#co
 
 * A Transactional N1QL Query using `AT_PLUS` will be automatically upgraded to `REQUEST_PLUS`.
 
-=== Additional storage use
+=== Additional Storage Use
 
 Transactions make use of the xref:learn:data/durability.adoc[Durability] provided by _synchronous writes_.
 As well as making use of individual documents' extended attributes, transactions also create additional documents in each bucket that they access.
--- a/modules/learn/pages/data/transactions.adoc
+++ b/modules/learn/pages/data/transactions.adoc
@@ -51,7 +51,11 @@ If left at this default, then all Key-Value writes performed at with durability 
 In turn this will cause all transactions (which perform all Key-Value writes durably) to fail.
 This setting can be changed via xref:manage:manage-buckets/create-bucket.adoc#couchbase-bucket-settings[GUI] or xref:cli:cbcli/couchbase-cli-bucket-create.adoc#options[command line].
 
+//  as "I've actually made it through the API that you can't do an AT_PLUS query."
+// commenting out, but keeping on hand as reminder that maybe this AT_PLUS limitation needs noting in another form!
+////
 * A Transactional N1QL Query using `AT_PLUS` will be automatically upgraded to `REQUEST_PLUS`.
+////
 
 * Non-transactional updates should not be made to any document involved in a transaction while the transaction is itself in progress: this prevents the non-transactional update from being overwritten.
 
--- a/modules/learn/pages/data/transactions.adoc
+++ b/modules/learn/pages/data/transactions.adoc
@@ -1,6 +1,7 @@
 = Transactions
 :description: pass:q[_Transactions_ guarantee that multiple documents can be updated atomically.]
 :page-aliases: acid-transactions,transactions,learn:data/distributed-acid-transactions,introduction:distributed-acid-transactions
+:page-edition: Enterprise Edition
 
 [abstract]
 {description}
--- a/modules/learn/pages/data/transactions.adoc
+++ b/modules/learn/pages/data/transactions.adoc
@@ -65,7 +65,10 @@ See xref:learn:clusters-and-availability/xdcr-conflict-resolution.adoc#ensuring_
 
 * Since transactions make use of the xref:learn:data/durability.adoc[Durability] provided by _synchronous writes_, in order to use transactions in development on a _single-node cluster_, the number of replicas assigned to any bucket to be used transactionally must be established as _zero_: if the number is greater than zero, the required durability level is unattainable, and the write fails.
 +
-If, on a single-node cluster, a bucket's replica-assignment was previously greater than zero, and the number is reduced to zero in order to support synchronous writes, a rebalance must be performed prior to any synchronous write.
+NOTE: If using a single node cluster (for example, during development), then note that the default number of replicas for a newly created bucket is 1.
+If left at this default, then all Key-Value writes performed at with durabiltiy will fail with a `DurabilityImpossibleException`.
+In turn this will cause all transactions (which perform all Key-Value writes durably) to fail.
+This setting can be changed via xref:manage:manage-buckets/create-bucket.adoc#couchbase-bucket-settings[GUI] or xref:cli:cbcli/couchbase-cli-bucket-create.adoc#options[command line].
 
 * A Transactional N1QL Query using `AT_PLUS` will be automatically upgraded to `REQUEST_PLUS`.
 
--- a/modules/learn/pages/data/transactions.adoc
+++ b/modules/learn/pages/data/transactions.adoc
@@ -66,7 +66,7 @@ See xref:learn:clusters-and-availability/xdcr-conflict-resolution.adoc#ensuring_
 * Since transactions make use of the xref:learn:data/durability.adoc[Durability] provided by _synchronous writes_, in order to use transactions in development on a _single-node cluster_, the number of replicas assigned to any bucket to be used transactionally must be established as _zero_: if the number is greater than zero, the required durability level is unattainable, and the write fails.
 +
 NOTE: If using a single node cluster (for example, during development), then note that the default number of replicas for a newly created bucket is 1.
-If left at this default, then all Key-Value writes performed at with durabiltiy will fail with a `DurabilityImpossibleException`.
+If left at this default, then all Key-Value writes performed at with durability will fail with a `DurabilityImpossibleException`.
 In turn this will cause all transactions (which perform all Key-Value writes durably) to fail.
 This setting can be changed via xref:manage:manage-buckets/create-bucket.adoc#couchbase-bucket-settings[GUI] or xref:cli:cbcli/couchbase-cli-bucket-create.adoc#options[command line].
 
--- a/modules/learn/pages/data/transactions.adoc
+++ b/modules/learn/pages/data/transactions.adoc
@@ -1,7 +1,6 @@
 = Transactions
 :description: pass:q[_Transactions_ guarantee that multiple documents can be updated atomically.]
 :page-aliases: acid-transactions,transactions,learn:data/distributed-acid-transactions,introduction:distributed-acid-transactions
-:page-edition: Enterprise Edition
 
 [abstract]
 {description}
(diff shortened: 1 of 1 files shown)


modules/learn/pages/data/transactions.adoc
```diff
@@ -5,38 +5,23 @@
...
-Collectively, these documents record all currently active transactions for the bucket.
-The name of each document is prefixed with `&#95;txn:atr-`.
+== Understanding Transactions
 
-* _Transaction Client Records_.
-One of these exists per bucket, covering all clients currently accessing the bucket transactionally.
-The document is named `&#95;txn:client-record`.
+_Distributed ACID Transactions_ are operations that ensure that when multiple documents need to be modified such that only the successful modification of all justifies the modification of any, either all the modifications do occur successfully; or none of them occurs.
+This _Atomicity_ supports _insert_, _update_, and _delete_ operations, across any number of documents.
+It also enables seemless integration between KV and xref:n1ql:n1ql-language-reference/begin-transaction.adoc#purpose[Query] operations within transactions.
```
programmatix: seamless, no?

modules/learn/pages/data/transactions.adoc
```diff
@@ -5,38 +5,23 @@
...
-Multiple Active Transaction Records can exist per bucket.
-Each of these documents can contain entries for multiple transactions.
-Collectively, these documents record all currently active transactions for the bucket.
-The name of each document is prefixed with `&#95;txn:atr-`.
+== Understanding Transactions
 
-* _Transaction Client Records_.
-One of these exists per bucket, covering all clients currently accessing the bucket transactionally.
-The document is named `&#95;txn:client-record`.
+_Distributed ACID Transactions_ are operations that ensure that when multiple documents need to be modified such that only the successful modification of all justifies the modification of any, either all the modifications do occur successfully; or none of them occurs.
```
programmatix, line 17: This just seems too wordy to me?  Also occur/occurs is a weird mix.

modules/learn/pages/data/transactions.adoc
```diff
@@ -5,38 +5,23 @@
...
-Consequently, the changes that occur during the transaction prior to commitment &#8212; these constituting _uncommitted_ (or _dirty_) data &#8212; are non-readable by any operation other than this transaction itself.
-Only following commit of the documents modified by the transaction &#8212; when the modified data-copy is removed from the meta data area, and is written over the main body of the document’s data &#8212; do the corresponding changes become readable by other transactions and operations.
-This isolation-level is known as https://jepsen.io/consistency/models/read-committed[Read Committed].
-Note that this use of a document’s extended attributes area results in an approximate doubling of its pre-transaction size, while the transaction is underway.
-In consequence, transactions can only be used on documents whose maximum size is 10 MB (which is half of the maximum-permitted document-size, 20 MB).
+Couchbase provides Distributed ACID Transactions for K-V (data) and Query operations -- these are available through APIs in 
+xref:cxx-txns::distributed-acid-transactions-from-the-sdk.adoc[C++],
```
programmatix, line 10: The somewhat limited APIs are one of our weaker points currently, maybe shouldn't be first thing we stress?  E.g. for any service we don't mention the APIs in the first line.

modules/learn/pages/data/transactions.adoc
programmatix: Also is it KV or K-V or Key-Value?

programmatix: I'd prefer not to link to the query transaction docs at all.  End-users should be using it through a transaction library - reading about BEGIN TRANSACTION is going to confuse people.

modules/learn/pages/data/transactions.adoc
```diff
@@ -5,38 +5,23 @@
...
-One of these exists per bucket, covering all clients currently accessing the bucket transactionally.
-The document is named `&#95;txn:client-record`.
+_Distributed ACID Transactions_ are operations that ensure that when multiple documents need to be modified such that only the successful modification of all justifies the modification of any, either all the modifications do occur successfully; or none of them occurs.
+This _Atomicity_ supports _insert_, _update_, and _delete_ operations, across any number of documents.
+It also enables seemless integration between KV and xref:n1ql:n1ql-language-reference/begin-transaction.adoc#purpose[Query] operations within transactions.
 
-These documents, which may persist indefinitely, are automatically maintained by Couchbase Server, and should not be modified by any application.
+Only following commit of the documents modified by the transaction -- when the modified data-copy is removed from the meta data area (which uses xref:learn:data/extended-attributes-fundamentals.adoc[Extended Attributes]), and is written over the main body of the document’s data -- do the corresponding changes become readable by other transactions and operations.
```
programmatix, line 21: Well, other transactions can read based on MAV.  I'm not sure how much we want to cover that here, admittedly - maybe one for PM.

modules/learn/pages/data/transactions.adoc
```diff
@@ -45,44 +30,78 @@ If two transactions simultaneously attempt to write to the same document, one is
...
 
-* The number of writes required by a transactional update is greater than the number required for a non-transactional update: a transactional update requires writes in order to stage and commit data, and also to maintain transaction records.
-Consequently, transactional updates may be less performant than non-transactional updates.
+* The number of writes required by a transactional update is greater than the number required for a non-transactional update.
+Thus transactional updates may be less performant than non-transactional updates.
 +
-Note that data within a single document is _always_ updated atomically: therefore, whenever multiple key-value pairs consistently require atomic update, their co-location within a single document may best ensure high performance.
+Note that data within a single document is _always_ updated atomically (without the need for transactions): therefore, whenever multiple key-value pairs consistently require atomic update, their co-location within a single document may best ensure high performance.
```
programmatix, line 71: Can this be rewritten more simply?  It feels wordy?

modules/learn/pages/data/transactions.adoc
```diff
@@ -45,44 +30,78 @@ If two transactions simultaneously attempt to write to the same document, one is
...
 See xref:learn:clusters-and-availability/xdcr-conflict-resolution.adoc#ensuring_safe_failover[Ensuring Safe Failover], for information on failing a transactional application from one data center to another.
 
 * Since transactions make use of the xref:learn:data/durability.adoc[Durability] provided by _synchronous writes_, in order to use transactions in development on a _single-node cluster_, the number of replicas assigned to any bucket to be used transactionally must be established as _zero_: if the number is greater than zero, the required durability level is unattainable, and the write fails.
 +
-If, on a single-node cluster, a bucket's replica-assignment was previously greater than zero, and the number is reduced to zero in order to support synchronous writes, a rebalance must be performed prior to any synchronous write.
+NOTE: If using a single node cluster (for example, during development), then note that the default number of replicas for a newly created bucket is 1.
+If left at this default, then all Key-Value writes performed at with durabiltiy will fail with a `DurabilityImpossibleException`.
```
programmatix: "durability" typo

modules/learn/pages/data/transactions.adoc
```diff
@@ -45,44 +30,78 @@ If two transactions simultaneously attempt to write to the same document, one is
...
+In turn this will cause all transactions (which perform all Key-Value writes durably) to fail.
+This setting can be changed via xref:manage:manage-buckets/create-bucket.adoc#couchbase-bucket-settings[GUI] or xref:cli:cbcli/couchbase-cli-bucket-create.adoc#options[command line].
```
programmatix: It's worth keeping the rebalance note, it's tripped up plenty of people.

modules/learn/pages/data/transactions.adoc
```diff
@@ -45,44 +30,78 @@ If two transactions simultaneously attempt to write to the same document, one is
...
+
+* A Transactional N1QL Query using `AT_PLUS` will be automatically upgraded to `REQUEST_PLUS`.
```
programmatix: I don't know.  This is already a somewhat scary list of limitations and gotchas, I don't know how much we want to put little details like this front and centre.

modules/learn/pages/data/transactions.adoc
```diff
@@ -45,44 +30,78 @@ If two transactions simultaneously attempt to write to the same document, one is
...
+
+=== Additional storage use
```
programmatix: I think Headers Are In This Format?

modules/learn/pages/data/transactions.adoc
```diff
@@ -45,44 +30,78 @@ If two transactions simultaneously attempt to write to the same document, one is
...
-For examples of creating and using transactions on Couchbase Server, see xref:3.0@java-sdk:howtos:distributed-acid-transactions-from-the-sdk.adoc[Distributed Transactions from the Java SDK].
+
+For examples of creating and using transactions on Couchbase Server, see:
+
+* xref:cxx-txns::distributed-acid-transactions-from-the-sdk.adoc[Distributed Transactions through the C++ API].
+* xref:dotnet-sdk:howtos:distributed-acid-transactions-from-the-sdk.adoc[Distributed Transactions from the .NET SDK].
+* xref:java-sdk:howtos:distributed-acid-transactions-from-the-sdk.adoc[Distributed Transactions from the Java SDK].
+
+NOTE: SDK Transactions Version 1.0.x do not support N1QL operations, but can be used with Couchbase Server 6.6.
+SDK Transactions Version 1.1.x require at least Server 7.0, and will work with N1QL operations.
```
programmatix, line 112: We don't have a consistent versioning scheme across all transaction libs so I wouldn't go with this.. Also AFAIK all versions of all transaction libs will be compatible with 6.6.1.
(We should explicitly call out 6.6.1 rather than 6.6 if we need to mention versions - 6.6.0 had a bad bug.)

programmatix: I think we should maybe just avoid mentioning specific transaction library versions at all?  If someone's using Java then they don't need to know that .NET doesn't yet support query.  It just looks bad.

modules/learn/pages/data/transactions.adoc
```diff
@@ -45,44 +30,78 @@ If two transactions simultaneously attempt to write to the same document, one is
...
 +
 See xref:learn:clusters-and-availability/xdcr-overview.adoc#xdcr-direction-and-topology[XDCR Direction and Topology], for a description of _bidirectional_ XDCR.
 See xref:learn:clusters-and-availability/xdcr-conflict-resolution.adoc#ensuring_safe_failover[Ensuring Safe Failover], for information on failing a transactional application from one data center to another.
 
 * Since transactions make use of the xref:learn:data/durability.adoc[Durability] provided by _synchronous writes_, in order to use transactions in development on a _single-node cluster_, the number of replicas assigned to any bucket to be used transactionally must be established as _zero_: if the number is greater than zero, the required durability level is unattainable, and the write fails.
 +
-If, on a single-node cluster, a bucket's replica-assignment was previously greater than zero, and the number is reduced to zero in order to support synchronous writes, a rebalance must be performed prior to any synchronous write.
+NOTE: If using a single node cluster (for example, during development), then note that the default number of replicas for a newly created bucket is 1.
```
programmatix: I don't like how much text this adds to the already huge limitations section, but I'm also very aware it's a common sticking point.
This isn't really a limitation.. Maybe we need a separate Gotchas section?  But with a better name...
But then also, this is meant to be a concepts doc, and this is very much a HOWTO point.. So, maybe it solely belongs in the transaction lib docs, which are the HOWTOs?

modules/learn/pages/data/transactions.adoc
```diff
@@ -45,44 +30,78 @@ If two transactions simultaneously attempt to write to the same document, one is
...
-Transactions, which constitute an evolving technology, currently have the following limitations:
+Couchbase Distributed ACID Transactions currently have the following limitations:
 
 * Only documents whose initial size is 10 MB or less can be included in a transaction.
 
 * Non-transactional updates should not be made to any document involved in a transaction while the transaction is itself in progress: this prevents the non-transactional update from being overwritten.
 
-* The number of writes required by a transactional update is greater than the number required for a non-transactional update: a transactional update requires writes in order to stage and commit data, and also to maintain transaction records.
-Consequently, transactional updates may be less performant than non-transactional updates.
+* The number of writes required by a transactional update is greater than the number required for a non-transactional update.
```
programmatix, line 68: Hmm, I'm not sure this is a limitation.. Maybe we need a separate section..

programmatix: (I'm very aware of the optics of the current very large limitations section.)

modules/learn/pages/data/transactions.adoc
programmatix: Combined with the above comment, I think we maybe need to split this out to a "When should I use transactions" or similar section?  This is not really limitation, this is best practices for data modelling.

modules/learn/pages/data/transactions.adoc
```diff
@@ -45,44 +30,78 @@ If two transactions simultaneously attempt to write to the same document, one is
 Use of transactions requires _Network Time Protocol_ (NTP) to be used to synchronize time across all cluster-nodes.
 See xref:install:synchronize-clocks-using-ntp.adoc[Clock Sync with NTP], for details.
 
-[#indexes-and-transactions]
 === Services and Transactions
 
 The indexes provided by the Index, Search, and Analytics Services are _not_ atomically updated with the commits performed by transactions: however, they are updated with https://en.wikipedia.org/wiki/Eventual_consistency[Eventual Consistency].
```
programmatix, line 35: We probs need to update this with talk about query.

programmatix (changes requested):
I think there's some stuff to do with the mechanics of transactions that we could usefully move over from the Java docs too.

modules/learn/pages/data/transactions.adoc
RichardSmedley: > seamless, no?

Fixed

RichardSmedley: > Also is it KV or K-V or Key-Value?

Key-value is abbreviated to KV / KV Service - https://docs.couchbase.com/home/contribute/word-list.html#k

RichardSmedley: K-V / KV may need revisiting across the docs team and beyond if people are gravitating to different forms :thinking:

modules/learn/pages/data/transactions.adoc
RichardSmedley: I don't disagree with you - but that's what's been up on the site since the doc was first written more than a year ago, let's not block the PR on that ;)

modules/learn/pages/data/transactions.adoc
RichardSmedley: otoh we're promoting that Query is now covered.

modules/learn/pages/data/transactions.adoc
RichardSmedley: Acknowledged.

modules/learn/pages/data/transactions.adoc
RichardSmedley: Thanks.
Fixed.

modules/learn/pages/data/transactions.adoc
RichardSmedley: Yep - fixed, thanks.

modules/learn/pages/data/transactions.adoc
```diff
@@ -1,6 +1,7 @@
 = Transactions
 :description: pass:q[_Transactions_ guarantee that multiple documents can be updated atomically.]
 :page-aliases: acid-transactions,transactions,learn:data/distributed-acid-transactions,introduction:distributed-acid-transactions
+:page-edition: Enterprise Edition
```
programmatix: Transactions are CE

modules/learn/pages/data/transactions.adoc
RichardSmedley: Shuffled sections

modules/learn/pages/data/transactions.adoc
RichardSmedley: Okay

modules/learn/pages/data/transactions.adoc
RichardSmedley: Section is now split.

modules/learn/pages/data/transactions.adoc
programmatix: Maybe I'm just unfamiliar with the syntax here, but it looks like the NOTE: line is still present and not commented out?  As it looks to be on a newline after the ////

modules/learn/pages/data/transactions.adoc
```diff
@@ -43,14 +42,25 @@ However, https://jepsen.io/consistency/models/snapshot-isolation[Snapshot Isolat
...
+* Since transactions make use of the xref:learn:data/durability.adoc[Durability] provided by _synchronous writes_, in order to use transactions in development on a _single-node cluster_, the number of replicas assigned to any bucket to be used transactionally must be established as _zero_: if the number is greater than zero, the required durability level is unattainable, and the write fails.
++
+NOTE: If using a single node cluster (for example, during development), then note that the default number of replicas for a newly created bucket is 1.
+If left at this default, then all Key-Value writes performed at with durability will fail with a `DurabilityImpossibleException`.
+In turn this will cause all transactions (which perform all Key-Value writes durably) to fail.
+This setting can be changed via xref:manage:manage-buckets/create-bucket.adoc#couchbase-bucket-settings[GUI] or xref:cli:cbcli/couchbase-cli-bucket-create.adoc#options[command line].
 
-* Only documents whose initial size is 10 MB or less can be included in a transaction.
+* A Transactional N1QL Query using `AT_PLUS` will be automatically upgraded to `REQUEST_PLUS`.
```
programmatix, line 57: Thanks for shuffling these bits around.  It's hard to picture it in this github format, but I think it'll be much clearer.
I still think this AT_PLUS notification is superfluous though.  I've actually made it through the API that you can't do an AT_PLUS query.

PR #1953, opened 2021-05-17
DOC-7481: Allow concurrent CREATE INDEX

simon-dew:

* [Global Secondary Indexes › Creating Global Secondary Indexes](https://docs-staging.couchbase.com/server/7.0/learn/services-and-indexes/indexes/global-secondary-indexes.html#creating-global-secondary-indexes) — described create phase and build phase
* [CREATE INDEX](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-language-reference/createindex.html) — added info about concurrent create operations, updated 5000 error
* [CREATE PRIMARY INDEX](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-language-reference/createprimaryindex.html) — added info about concurrent create operations, updated 5000 error
* [BUILD INDEX](https://docs-staging.couchbase.com/server/7.0/n1ql/n1ql-language-reference/build-index.html) — updated 5000 error, added note about building all indexes in a collection

Docs issue: [DOC-7481](https://issues.couchbase.com/browse/DOC-7481)

### DIFF
--- a/modules/learn/pages/services-and-indexes/indexes/global-secondary-indexes.adoc
+++ b/modules/learn/pages/services-and-indexes/indexes/global-secondary-indexes.adoc
@@ -1,10 +1,11 @@
 = Global Secondary Indexes
-:description: pass:q[A _Global Secondary Index_ (GSI) supports queries made by the _Query Service_ on attributes within documents. \
-Extensive filtering is supported.]
 :page-aliases: indexes:indexing-overview,understanding-couchbase:services-and-indexes/indexes/global-secondary-indexes,indexes:gsi-for-n1ql,architecture:global-secondary-indexes,architecture:gsi-versus-views
+:description: A Global Secondary Index (GSI) supports queries made by the Query Service on attributes within documents. \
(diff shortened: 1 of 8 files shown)


tonyjhillman (approved):

simon-dew:
- [x] As a part of this feature, multiple index creation requests will be admitted by the server. The indexes will get created in the background - so that they are still serialised - but the caller does not need to worry. The calling function will return only after index is created.
- [x] The behaviour of index build phase remains the same.
- [x] User can monitor the indexes that are scheduled for background creation via system:indexes n1ql qury as well as via UI.
- [x] When index creation happens in the background, it can fail and will be retried. Number of retries = 1000.
- [x] After 1000 retries, if index creation fails, index will become errored (state = error). And user has to explicitly drop the index.
- [x] User requests to scan/move/build/alter the index which is scheduled for background creation, will fail. Index drop will return a message but will succeed.

simon-dew:
Update links to staging docs

PR #1959, opened 2021-05-19
Update eventing-handler-convertXMLtoJSON.adoc

jon-strabala:
Bad character in the code


jon-strabala:
It might be just fine need to test ...

PR #1960, opened 2021-05-19
use counter instead of built-in example-number attribute to number Q&A examples

mojavelinux:
Using the built-in example-number attribute is not appropriate for this use case.

### DIFF
--- a/modules/n1ql/pages/n1ql-language-reference/indexing-arrays.adoc
+++ b/modules/n1ql/pages/n1ql-language-reference/indexing-arrays.adoc
@@ -268,8 +268,8 @@ The following examples use the {install-sample-buckets}[travel-sample] bucket th
 .Indexing all DISTINCT elements in an array
 ====
 
-.C{example-number}: Create an index on all schedules
-[[C1,C{example-number}]]
+C{counter:example-counter}: Create an index on all schedules
+[[C1,C{example-counter}]]
 [source,N1QL]
 ----
 CREATE INDEX idx_sched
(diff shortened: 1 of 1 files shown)


mojavelinux:
This change also needs to be backported to 6.6, 6.5, and 6.0.

amarantha-k (approved):

amarantha-k:
I'll backport the changes to 6.6, 6.5, and 6.0 releases.

modules/n1ql/pages/n1ql-language-reference/indexing-arrays.adoc
```diff
@@ -268,17 +268,17 @@ The following examples use the {install-sample-buckets}[travel-sample] bucket th
 .Indexing all DISTINCT elements in an array
 ====
 
-.C{example-number}: Create an index on all schedules
-[[C1,C{example-number}]]
+C{counter:example-counter}: Create an index on all schedules
```
simon-dew, line 271: Is there a reason why the C examples (CREATE INDEX) use plain paragraphs (no initial dot), whereas the Q examples (SELECT) use block titles (with initial dot)?

```
C{counter:example-counter}
```

versus

```
.Q{example-counter}
```

mojavelinux: My mistake. That's a search and replace typo. It should be the same as it was in the original. Are you able to correct it?

simon-dew: No problem, I'll update the titles in a  new PR.

I need to look at this in more detail, because after this change the example counters don't seem to work:

Screen Shot 2021-05-27 at 00 18 24

(Created with Antora 3.0.0 alpha 2)

PR #1961, opened 2021-05-19
fix footnote so it is compatible with Asciidoctor 2

mojavelinux:


mojavelinux:
Another way to solve this is to use the shorthand form of the xref macro in the footnote text. For example, `<<n1ql-language-reference/groupby.adoc#,GROUP BY>>`

mojavelinux:
This also needs to be backported to 6.6 and 6.5.

amarantha-k (approved):

mojavelinux:
@simon-dew or @amarantha-k, will you backport this to 6.6 and 6.5?

amarantha-k:
Yes @mojavelinux. I'll backport to 6.6 and 6.5.

PR #1975, opened 2021-05-26
fix syntax for block title

mojavelinux:
* restore missing . at start of line

### DIFF
--- a/modules/n1ql/pages/n1ql-language-reference/indexing-arrays.adoc
+++ b/modules/n1ql/pages/n1ql-language-reference/indexing-arrays.adoc
@@ -268,7 +268,7 @@ The following examples use the {install-sample-buckets}[travel-sample] bucket th
 .Indexing all DISTINCT elements in an array
 ====
 
-C{counter:example-counter}: Create an index on all schedules
+.C{counter:example-counter}: Create an index on all schedules
 [[C1,C{example-counter}]]
 [source,N1QL]
 ----
@@ -288,7 +288,7 @@ WHERE ANY v IN schedule SATISFIES v.flight LIKE 'UA%' END;
 
 .Partial index (with WHERE clause) of individual attributes from selected elements (using WHEN clause) of an array
 ====
-C{counter:example-counter}: Create an index on flights from San Francisco scheduled in the first 4 days of the week
+.C{counter:example-counter}: Create an index on flights from San Francisco scheduled in the first 4 days of the week
 [[C2,C{example-counter}]]
 [source,N1QL]
 ----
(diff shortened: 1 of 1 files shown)


mojavelinux:
@amarantha-k This fixes a search and replace error that @simon-dew discovered in a previous PR. Please apply to the 6.6, 6.5, and 6.0 releases as well.

Related PR: #1960

amarantha-k (approved):

modules/n1ql/pages/n1ql-language-reference/indexing-arrays.adoc
```diff
@@ -288,7 +288,7 @@ WHERE ANY v IN schedule SATISFIES v.flight LIKE 'UA%' END;
 
 .Partial index (with WHERE clause) of individual attributes from selected elements (using WHEN clause) of an array
 ====
-C{counter:example-counter}: Create an index on flights from San Francisco scheduled in the first 4 days of the week
+.C{counter:example-counter}: Create an index on flights from San Francisco scheduled in the first 4 days of the week
 [[C2,C{example-counter}]]
 [source,N1QL]
 ----
```
simon-dew, line 294: Antora 3.0.0 alpha 2. The example counter works in the running text, but is not being resolved in xref labels.

Screen Shot 2021-05-27 at 00 18 24

simon-dew: For comparison, the `{example-number}` attribute _is_ resolved in xrefs when using Antora 3.0.0 alpha 2.

Screen Shot 2021-05-27 at 00 56 38

simon-dew (commented):
The edit is fine, but I'm seeing problems with the output with Antora 3.0.0 alpha 2. Is this fixed by alpha 5?

modules/n1ql/pages/n1ql-language-reference/indexing-arrays.adoc
mojavelinux: Yes, this fix for Asciidoctor 2 doesn't seem to also work for Asciidoctor 1.5.x. Since the plan is to upgrade immediately, I think we can just make the leap. If we need to accommodate both Asciidoctor versions, then it would be necessary to add ifeval logic around the two alternatives.

simon-dew: Okay, I updated to Antora 3.0.0 alpha 5, and the example counter is now resolved in xref labels, but unfortunately it's still not working quite right. Looking at the same piece of text:

Screen Shot 2021-05-27 at 11 54 41

Both instances of *C1* (the first is highlighted) should be *C2*, as in [the comment above](https://github.com/couchbase/docs-server/pull/1975#discussion_r640195998).

mojavelinux:
That's why I'm saying just get rid of the counter altogether. It was never
intended to be used for this use case (in fact, there's almost no use case
for which a counter should be used because it's fraught with limitations
and side effects). I, myself, wasn't even aware this usage pattern worked
in 1.5.x.

PR #1976, opened 2021-05-26
Backport of PR 1960 to use counter instead of built-in example-number…

amarantha-k:
… attribute

### DIFF
--- a/modules/n1ql/pages/n1ql-language-reference/indexing-arrays.adoc
+++ b/modules/n1ql/pages/n1ql-language-reference/indexing-arrays.adoc
@@ -249,16 +249,16 @@ The following examples use the xref:manage:manage-settings/install-sample-bucket
 .Indexing all DISTINCT elements in an array
 ====
 
-.C{example-number}: Create an index on all schedules
-[[C1,C{example-number}]]
+.C{counter:example-counter}: Create an index on all schedules
+[[C1,C{example-counter}]]
 [source,N1QL]
 ----
 CREATE INDEX idx_sched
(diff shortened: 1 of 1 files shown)


mojavelinux:
As Simon pointed out, this fix doesn't work for both Asciidoctor 1.5.x and Asciidoctor 2. So we either need to make this fix, then make the jump to Asciidoctor 2 (aka the Antora upgrade), or go back and try to find something that works for both using ifeval.

mojavelinux (approved):

mojavelinux (changes requested):
On further review, I see that there are even problems with this fix. The counter should only be incremented for the "C", not the "Q". Otherwise, the numbers get out of sync.

I can also verify that the fix **does** work for both Asciidoctor 1.5.x and 2.

mojavelinux:
I want to revise my statement again. The use of counters on this page just need to be removed. Counters just weren't designed for this kind of use case. And they are only adding needless complexity anyway since they can only be used in certain cases and not in others. Just sticking to literal numbers would be much more clear, stable, and maintainable.

mojavelinux:
Here's the actual change I recommend applying: https://github.com/couchbase/docs-server/pull/1979

simon-dew (commented):
Further to my review of #1979, I think it would be better to rewrite this page not to use numbered code listings at all. Please await my pull request for release/7.0 and backport that.

simon-dew:
Please backport #1984 instead.

amarantha-k:
As noted in the comments, closing this PR without merge.

PR #1979, opened 2021-05-27
remove use of counter in block titles and reftext

mojavelinux:


simon-dew (commented):
I take your point that this is simpler, but examples often get moved around, inserted in the middle of a document, or removed from the middle of a document, which would mean renumbering all the code listings by hand — tedious and error-prone. It makes more sense to me to rewrite the cross-references so they don't use numbered code listings at all. I'll create a new pull request for this myself.

simon-dew:
I considered using the `listing-caption` attribute, but I think having all the code listings numbered straight through from start to finish was cluttered and less readable.

simon-dew:
Please backport #1984  instead.

mojavelinux:
> examples often get moved around, inserted in the middle of a document, or removed from the middle of a document, which would mean renumbering all the code listings by hand — tedious and error-prone.

My point is that the way it was written would have made it even more tedious and error-prone to do that. It was a mix of counters and explicit numbering, so the counters were pointless. It's part of the reason it took me so long to figure out what was going on. It's much, much simpler and clearer in this case to use the explicit numbering consistently.

I do understand your underlying point about auto-numbering to keep things DRY. While I generally advocate for that practice, counters are not designed for that purpose. There are some facilities in AsciiDoc to do auto-numbering, but not as advanced as would be needed for this case. So perhaps new ones need to be introduced. But that's an underlying issue with the language itself.

> It makes more sense to me to rewrite the cross-references so they don't use numbered code listings at all.

Totally agree.

amarantha-k:
This has been superseded by PR#1984.

PR #1981, opened 2021-05-27
DOC-8582for6.5: Added indexer https port

tonyjhillman:


tonyjhillman:
This change has been made.

PR #1982, opened 2021-05-27
DOC-8521for6.6: Added indexer https port

tonyjhillman:


tonyjhillman:
This change has been made.

PR #1983, opened 2021-05-27
Doc 8584

tonyjhillman:


modules/manage/pages/manage-backup-and-restore/manage-backup-and-restore.adoc
```diff
@@ -821,6 +827,9 @@ The time at which the task is commenced is that specified in the *Time* field.
 [#use-cloud-storage]
 == Use Cloud Storage
 
+Cloud storage can be used for backups.
+Note that it _cannot_ be used for merges.
```
cg14823: I would say more than it *cannot* be used for merges I would word it like it will skip merge tasks if the plan has any.

So you can in theory add a plan that has merges to a cloud repo, it will just skip them as it is not supported.

cg14823 (approved):

PR #1984, opened 2021-05-27
remove use of counter in block titles and reftext

simon-dew:
Reworded so that code listings are no longer referenced using an auto-generated number based on the example number. Examples now have their own IDs. WIthin each example, code listings are referenced using a short title, including a distinguishing letter if necessary, which is hand-generated. References to code listings in other examples include a cross reference to the example.

### DIFF
--- a/modules/n1ql/pages/n1ql-language-reference/indexing-arrays.adoc
+++ b/modules/n1ql/pages/n1ql-language-reference/indexing-arrays.adoc
@@ -265,11 +265,11 @@ They can efficiently index and query recursive trees of arbitrary depth.
 
 The following examples use the {install-sample-buckets}[travel-sample] bucket that is shipped with Couchbase Server.
 
+[[example-1]]
 .Indexing all DISTINCT elements in an array
 ====
-
-.C{counter:example-counter}: Create an index on all schedules
-[[C1,C{example-counter}]]
+.Index: Create an index on all schedules
+[[C1,Index]]
 [source,N1QL]
 ----
 CREATE INDEX idx_sched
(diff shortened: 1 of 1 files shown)


simon-dew:
My follow-up to #1960 and #1975

modules/n1ql/pages/n1ql-language-reference/indexing-arrays.adoc
```diff
@@ -265,31 +265,32 @@ They can efficiently index and query recursive trees of arbitrary depth.
 
 The following examples use the {install-sample-buckets}[travel-sample] bucket that is shipped with Couchbase Server.
 
+[[example-1]]
```
mojavelinux, line 268: The ID on the example blocks themselves aren't really necessary unless yo'are planning to link directly to them from an xref.

simon-dew: Sure. Some of them are indeed linked to directly, including this one. It was easier to give them all an ID than to work out which ones needed it. In fact I generally like to give all examples an explicit ID, because it's useful to be able to link to them from other documents.

mojavelinux: That make sense. Though I would encourage you to use a semantic ID rather that one that reinforces the document order. That way, if you do move them around, all you need to change are the textual labels...and you don't break any existing links.

mojavelinux:
The only change I would recommend is to use semantic IDs for the examples instead of encoding the example number. Though that could be a follow-up change since it might make sense to change all the numbered IDs in the same way.

amarantha-k:
Thank you Simon and Dan. I'll merge this for now and we'll plan to take a look at Dan's recommendation to use semantic IDs separately.

amarantha-k (approved):

PR #1985, opened 2021-05-28
DOC-8491b: Metadata management doc for 7.0

tonyjhillman:


dave-finlay (approved):
Hi Tony - thanks for putting this together in one PR tor review. Looks good. The section which includes some informational material on Chronicle is possibly a little confusing since it introduces a bunch of concepts that people don't really 100% need to know (followers, replicas, replicated state machines and so on.) But it does say that it's for informational purposes only, so I think it's good. 

I may suggest some other changes after I see it included in the build, but I think this is good for now. Thanks for your patience. :-)

PR #1998, opened 2021-06-08
DOC-8594a: Updated information on mctimings

tonyjhillman:


tonyjhillman:
I've updated this page for DOC-8594. It's not a complete rewrite. I have change references to username and password to correspond with current RBAC-based usage, and I have stated that hex codes need to be specified in back-ticks (which is the only way I can get them recognized on Linux and MacOS). I have updated the example commands accordingly.

mcpaddy (commented):
Is the commit message correct, with the "a" at the end of the defect ID ("DOC-8594a")?
With regards to the change itself `mctimings` is owned by the kv-team , that said  I don't see an issue with this commit.

PR #2024, opened 2021-06-24
Doc 7032

tonyjhillman:


modules/install/pages/upgrade-procedure-selection.adoc
```diff
@@ -0,0 +1,178 @@
...
+All clusters are assumed to be _multi-node_ clusters, except in the final section, xref:install:upgrade-strategies.adoc#upgrade-for-single-node-clusters[Upgrade for Single-Node Clusters].
+
+[#online-upgrade]
+=== Upgrade with Cluster Online
+
+For a multi-node cluster, an _online upgrade_ means that the cluster continues to serve data while its nodes are progressively upgraded.
+Online upgrade can be performed in either of the following ways.
+
+[#cluster-online-swap-rebalance-with-capacity-reduced]
+==== Cluster Online: Swap Rebalance at Full Capacity
```
mcpaddy, line 112: I wonder if we should make a recommendation for cloud environments, given that is where most of our customers operate and with what marketing is positioning with.

mcpaddy (commented):
It looks good. There're bit that are duplicated I wonder if it's worth putting them into their own adoc file and getting the doc generation to inline them? That is what we do in Couchbase-cli and backup for shared text. 

To be clear I'm not suggesting that the text be on it's own web page this is just for easy of management of the docs.

tonyjhillman:
I've added the note on cloud. I will put some of the procedure-steps into asciidoc partials under a new ticket.

modules/install/pages/upgrade-cluster-offline.adoc
```diff
@@ -0,0 +1,238 @@
...
++
+Once the archive and repository have been created, a command such as the following performs a full backup:
++
+[source,bash]
+----
+cbbackupmgr backup --archive ${ABS_PATH_TO_ARCHIVE} --repo ${REPO_NAME} --cluster ${CLUSTER_ADDRESS} --username ${USERNAME} --password ${PASSWORD} --full-backup
+----
++
+Here, the `CLUSTER_ADDRESS` is the IP address or domain name of the cluster.
+The `full-backup` flag ensures that the backup is indeed a _full_ backup.
```
jamesl33, line 54: I think we should refer to the flag as '--full-backup' (with the '--' prefix) unless that's not something we do elsewhere.

modules/install/pages/upgrade-cluster-offline.adoc
```diff
@@ -0,0 +1,238 @@
...
++
+For the equivalent procedure as performed by the Backup Service, see xref:manage:manage-backup-and-restore/manage-backup-and-restore.adoc#run-an-immediate-backup[Run an Immediate Backup].
++
+In the event of node-upgrade failing, the backed-up data can be restored by means of xref:backup-restore:cbbackupmgr-restore.adoc[cbbackupmgr restore] or xref:cli:cbtools/cbrecovery.adoc[cbrecovery]; or by means of the Backup Service (see xref:manage:manage-backup-and-restore/manage-backup-and-restore.adoc#restore-backups[Restore Backups]).
+
+. Disable auto-failover (to prevent auto-failover from occuring when individual nodes stop communicating with the rest of the cluster, during their upgrade).
+Disablement is performed by means of the xref:manage:manage-settings/general-settings.adoc[General Settings] screen, by means of the xref:manage:manage-settings/general-settings.adoc#node-availability[Node Availability] panel.
```
jamesl33, line 61: It seems a little odd that we're suggesting that the user should disable auto-failover using the WebUI, but the following example is using the CLI (a clearer separation here could be beneficial). 

I imagine the linked adoc may cover the whole procedure by the WebUI, so this might not be an issue.

modules/install/pages/upgrade-cluster-offline.adoc
```diff
@@ -0,0 +1,238 @@
...
+. If using Couchbase-provided repositories/PPAs, the Couchbase-Server version should be _unpinned_, to allow it to be upgraded.
+Proceed as follows, for the appropriate platform:
++
+[{tabs}]
+====
+RedHat & Centos::
++
+--
+
+Move the reference to the `couchbase-server` (or `couchbase-server-community`, if using Community Edition) package to the `exclude` section of the `yum.conf` file, located in `/etc/yum/yum.conf`.
```
jamesl33, line 114: I might be reading this incorrectly, but this section/example appears to be suggesting that I should add `couchbase-server` to the `exclude` section of the `yum.conf` file (specifically, the use of "Move" and "to the `exclude`" are throwing me off here).

The example below (and description above) suggests that we're unpinning, in which case something like "**Remove** the reference to the `couchbase-server` package **from** the `exclude` section" would be slightly more clear in my opinion.

modules/install/pages/upgrade-cluster-offline.adoc
```diff
@@ -0,0 +1,238 @@
...
++
+Enter the following commands (substituting, if using Community Edition, `couchbase-server-community` for `couchbase-server`).
++
+----
+systemctl enable couchbase-server.service
+
+systemctl is-active --quiet couchbase-server.service || systemctl start couchbase-server.service
+----
+
+. _Repin_ (or _hold_) future package-upgrades for Couchbase Server, so that none occurs before the administrator's next, elective, manually driven upgrade.
```
jamesl33, line 177: I think this would read better with "occur" instead of "occurs" but feel free to ignore

modules/install/pages/upgrade-cluster-offline.adoc
```diff
@@ -0,0 +1,238 @@
...
+The status of warmup can be checked for each node as follows:
++
+----
+cbstats ${NODE_ADDRESS}:${NODE_KV_PORT} -u ${USERNAME} -p ${PASSWORD} -b ${BUCKET} warmup | grep state
+----
++
+For example:
++
+----
+/opt/couchbase/bin/cbstats localhost:11210 -u Administrator -p password -b travel-sample warmup | grep state
```
jamesl33, line 224: Should we have examples for other commands which use the ${VARIABLE} substitution style?

modules/install/pages/upgrade-cluster-offline.adoc
```diff
@@ -0,0 +1,238 @@
...
+
+. Enable and start the `couchbase-server` (or `couchbase-server-community`) service.
+(The package installer may already have performed enablement: however, explicit enablement is nevertheless recommended.)
++
+Enter the following commands (substituting, if using Community Edition, `couchbase-server-community` for `couchbase-server`).
++
+----
+systemctl enable couchbase-server.service
+
+systemctl is-active --quiet couchbase-server.service || systemctl start couchbase-server.service
```
jamesl33, line 174: I know this was the command that I provided, but considering we're using the `--now` flag when disabling, we could also do that here i.e. `systemctl enable --now couchbase-server.service` (this command is just more backwards compatible, same applies to disable; I don't recall exactly when the `--now` flag was added).

modules/install/pages/upgrade-cluster-online-full-capacity.adoc
```diff
@@ -0,0 +1,236 @@
+= Upgrade a Full-Capacity, Online Cluster
+
+:description: A cluster can be upgraded while still online; and maintained at full capacity.
```
jamesl33, line 3: Should this be "and maintain" instead of "and maintained"?

modules/install/pages/upgrade-cluster-online-full-capacity.adoc
```diff
@@ -0,0 +1,236 @@
...
+:tabs:
+
+[abstract]
+{description}
+
+[#online-upgrade-with-full-capacity]
+== Understanding Full-Capacity Upgrade
+
+The context and overall requirements for upgrading a live cluster by means of the introduction of one or more additional nodes are described in xref:install:upgrade-strategies.adoc[Upgrade-Procedure Selection]: this fully explains the node-by-node upgrade of the cluster, using _swap rebalance_ to minimize overhead.
+It also explains how, if additional nodes are available, the procedure can be completed with cluster performing at full capacity, during the upgrade.
```
jamesl33, line 13: "with the cluster"?

modules/install/pages/upgrade-cluster-online-full-capacity.adoc
```diff
@@ -0,0 +1,236 @@
...
+
+The precise steps for this procedure are detailed on this page, below.
+A full understanding of the information in xref:install:upgrade-strategies.adoc[Upgrade-Procedure Selection] should be acquired, before proceeding.
+The procedure assumes that:
+
+* The cluster to be upgraded must continue to serve data throughout the cluster-upgrade process.
+
+* A single, additional node is available; to act as a _swap_ node.
```
jamesl33, line 21: Should we mention that it could be one or more nodes (as long as it's an equal amount being rebalanced in/out)?

(reading ahead, it looks like we may be keeping it simple here, so maybe we don't need to mention "one or more")

modules/install/pages/upgrade-cluster-online-full-capacity.adoc
```diff
@@ -0,0 +1,236 @@
...
+
+* During the cluster-upgrade process, _swap_ nodes and _upgraded_ nodes will be introduced to the cluster by the _node-addition_ procedure; while _failed over spare nodes_ and _nodes to be upgraded_ will be withdrawn from the cluster by means of the _node removal_ and _swap rebalance_ procedures.
++
+For information on node-addition, see xref:learn:clusters-and-availability/nodes.adoc#clusters[Clusters].
+For information on node-removal, see xref:learn:clusters-and-availability/removal.adoc[Removal].
+For information on _swap rebalance_, see xref:install:upgrade-strategies.adoc#swap-rebalance[Swap Rebalance].
+
+* Nodes will be upgraded _one at a time_.
+
+The overall procedure is described in seven stages, below.
```
jamesl33, line 33: This might have been done on other pages, but IMO mentioning the number of stages (in this case "seven") opens us up for easy mistakes. For example, if we add/remove stages in the future, it would be extremely easy to miss/forget to update this.

Could this be automatically determined/omitted?

modules/install/pages/upgrade-cluster-online-full-capacity.adoc
```diff
@@ -0,0 +1,236 @@
...
+----
+cbbackupmgr config --archive ${ABS_PATH_TO_ARCHIVE} --repo ${REPO_NAME}
+----
+
+Here, `ABS_PATH_TO_ARCHIVE` is an absolute path to a filesystem location that will serve as the archive within which the backed up data will reside.
+The `REPO_NAME` is the name of the repository that will be associated with the location.
+
+Once the archive and repository have been created, a command such as the following performs a full backup:
+
+[source,bash]
```
jamesl33, line 59: Does this need to be put above each `----` block (if so, I think there might be some missing)?

modules/install/pages/upgrade-cluster-online-full-capacity.adoc
```diff
@@ -0,0 +1,236 @@
...
+
+[#prepare-the-cluster]
+== Stage Two: Prepare the Cluster
+
+Back up all cluster-data.
+This can be performed either with xref:backup-restore:enterprise-backup-restore.adoc[cbbackupmgr] or with the xref:learn:services-and-indexes/services/backup-service.adoc[Backup Service]; and should be a _full_ (rather than an incremental) backup.
+
+For example, to use `cbbackupmgr` to configure an archive and repository for the backup, a command of the following form should be entered:
+
+[source,bash]
```
jamesl33, line 49: Should we extract the backup procedure into a separate ADOC template and import it in each document?

Should we also document the method for creating a full backup for versions of 'cbbackupmgr' which don't have the `--full-backup` flag (i.e. create a new repo and the first backup will be full)?

modules/install/pages/upgrade-cluster-online-full-capacity.adoc
```diff
@@ -0,0 +1,236 @@
...
+--cluster ${CLUSTER_ADDRESS} --username ${USERNAME} --password ${PASSWORD} \
+--full-backup
+----
+
+Here, the `CLUSTER_ADDRESS` is the IP address or domain name of the cluster that is being backed up.
+The `full-backup` flag ensures that the backup is indeed a _full_ backup.
+
+For the equivalent procedure as performed by the Backup Service, see xref:manage:manage-backup-and-restore/manage-backup-and-restore.adoc#run-an-immediate-backup[Run an Immediate Backup].
+
+In the event of node-upgrade failing, the backed-up data can be restored by means of xref:backup-restore:cbbackupmgr-restore.adoc[cbbackupmgr restore] or xref:cli:cbtools/cbrecovery.adoc[cbrecovery]; or by means of the Backup Service (see xref:manage:manage-backup-and-restore/manage-backup-and-restore.adoc#restore-backups[Restore Backups]).
```
jamesl33, line 71: I see that we've mentioned cbrecovery, from memory this only applies when using XDCR, however, I may be wrong (we should note this).

The notes that I supplied referenced `cbdatarecovery` which is a Go version of `cbtransfer`. Maybe we should mention both of these tools?

modules/install/pages/upgrade-cluster-online-full-capacity.adoc
```diff
@@ -0,0 +1,236 @@
...
+
+[#add-and-remove-a-node]
+== Stage Three: Add the Spare Node, and Remove a Node
+
+_Add_ the upgraded, removed node back into the cluster, and _remove_ a node that is currently part of the cluster.
+The node that is being added should be configured to run the same service as that running on the node that is to be removed.
```
jamesl33, line 77: We should say "service(s)" here, since it's supported (but not recommended) to run multiple services on one node.

modules/install/pages/upgrade-cluster-online-full-capacity.adoc
```diff
@@ -0,0 +1,236 @@
...
+For example, if the node to be removed is running the Data Service, configure the node to be added to run the Data Service.
+Couchbase Server will execute the rebalance as a _swap rebalance_, to maximize efficiency.
+
+For an overview of node-removal, see xref:learn:clusters-and-availability/removal.adoc[Removal]; and for practical examples of performing removal, see xref:manage:manage-nodes/remove-node-and-rebalance.adoc.
+For an overview of node-addition, see xref:learn:clusters-and-availability/nodes.html#clusters[Clusters]; and for practical examples of node-addition, see xref:manage:manage-nodes/add-node-and-rebalance.adoc[Add a Node and Rebalance].
+
+Note that for the CLI and REST API, the staging of a swap rebalance requires _two_ separate commands.
+The first specifies that one or more nodes be _added_ to the cluster: however, this command requires a subsequent rebalance to be performed, to complete the process.
+The second command indeed specifies that subsequent rebalance, but also includes an instruction to _remove_ one or more nodes from the cluster: therefore, as the rebalance occurs, it finalizes both node-addition and node-removal.
+
```
jamesl33, line 87: Are we missing an example here (you mention "two separate" commands, but I don't see any, maybe they're in a linked document)?

modules/install/pages/upgrade-cluster-online-full-capacity.adoc
```diff
@@ -0,0 +1,236 @@
...
+[#upgrade-the-removed-node]
+== Stage Four: Upgrade the Removed Node
+
+Upgrade the node that has just been removed, by proceeding as follows:
+
+. Stop the `couchbase-server.service` service, on the node.
+Enter the following command:
++
+----
+systemctl stop couchbase-server.service
```
jamesl33, line 97: I think this section has been repeated, could it be extracted into a template document?

(continues all the way past the install procedure. If there's specifics, we could avoid duplication by using conditionals in the template)

modules/install/pages/upgrade-cluster-online-full-capacity.adoc
```diff
@@ -0,0 +1,236 @@
...
+
+Upgrade the newly removed node as described above, in xref:install:upgrade-cluster-online-full-capacity.adoc#upgrade-the-removed-node[Stage Four: Upgrade the Removed Node].
+Then, add the upgraded node back into the cluster, and remove another node to be upgraded: as described above, in xref:install:upgrade-cluster-online-full-capacity.adoc#add-back-the-upgraded-node-and-remove-another-node[Stage Five: Add Back the Upgraded Node, and Remove Another Node].
+
+Continue to remove, upgrade, and restore nodes in this way; until the cluster is fully upgraded, and only the spare node remains outside the cluster.
+
+[#add-back-the-last-node]
+== Stage Seven: Repurpose the Spare Node
+
+One the cluster has been fully upgraded, the spare node can be repurposed.
```
jamesl33, line 235: Once

jamesl33 (commented):
Apologies for reviewing after this was merged; some of my comments are nitpicking so feel free to ignore them.

jamesl33:
> I've added the note on cloud. I will put some of the procedure-steps into asciidoc partials under a new ticket.

@tonyjhillman I missed this when I was reviewing, so some of my comments may not be valid.

modules/install/pages/upgrade-cluster-offline.adoc
tonyjhillman: Ok, done.

modules/install/pages/upgrade-cluster-offline.adoc
tonyjhillman: I have clarified that the linked page has explanations of all ways of doing it.

modules/install/pages/upgrade-cluster-offline.adoc
tonyjhillman: I'll try to clarify this.

modules/install/pages/upgrade-cluster-offline.adoc
tonyjhillman: Although the plural is indeed often used, "none" more properly governs a singular. ("None" = "no one".)

modules/install/pages/upgrade-cluster-offline.adoc
tonyjhillman: I guess we should. I would say we can open a ticket and do for .Next.

modules/install/pages/upgrade-cluster-online-full-capacity.adoc
tonyjhillman: Looking at this, I think it would be better to remove this paragraph (which seems partly incorrect and partly incomplete), and open a ticket for a fuller subsection, entitled something like Data Recovery After Upgrade-Failure. 
Then we can go into more detail about each of these, and how it should be used. What do you think?

modules/install/pages/upgrade-cluster-online-full-capacity.adoc
tonyjhillman: Have fixed.

modules/install/pages/upgrade-cluster-online-full-capacity.adoc
tonyjhillman: I have added links to sections showing how it is done.

modules/install/pages/upgrade-cluster-online-full-capacity.adoc
tonyjhillman: Yes, once we have the procedures settled, I will pull stuff out into asciidoc partials. We may want to handle non-root before that, and I think we'll also need to look at commands for some additional platforms, including Windows.

modules/install/pages/upgrade-cluster-online-full-capacity.adoc
tonyjhillman: Thanks. Fixed.

modules/install/pages/upgrade-cluster-offline.adoc
tonyjhillman: I've added --now to the enable command on each of the pages.

modules/install/pages/upgrade-cluster-online-full-capacity.adoc
tonyjhillman: No, maintained is right. But I've changed the punctuation to make it clearer. ("Can be" is intended to govern both "upgraded" and "maintained"._

modules/install/pages/upgrade-cluster-online-full-capacity.adoc
tonyjhillman: I've changed to "It also explains how, if additional nodes are available, the cluster can be maintained at full capacity during the upgrade."

modules/install/pages/upgrade-cluster-online-full-capacity.adoc
tonyjhillman: Yes I think it's okay here, because we are stating the assumptions for this particular example-procedure. We've mentioned "one or more" on the overview pages.

modules/install/pages/upgrade-cluster-online-full-capacity.adoc
tonyjhillman: Not really, that's the problem. I completely agree it's a liability. But right now, asciidoc limits what we can do. We can only automate the numbering of paragraphs. I'd must sooner divide the paras into sections, rather than have nested numbered-paras (which are hard to read, and which asciidoc is buggy in its handling of). We should really be able to number sections, but can't yet. What I will do is open a ticket to get this looked at (we've actually talked about this, but haven't scheduled the contractor's time).

modules/install/pages/upgrade-cluster-online-full-capacity.adoc
tonyjhillman: Yes. I'll fix. Thanks.

modules/install/pages/upgrade-cluster-online-full-capacity.adoc
tonyjhillman: Yes, I think that is something we should do. We can do it on a subsequent ticket. Regarding the versions of cbbackupmgr without the flag, those are pre-7.0, right? If so, we would do that when updating the upgrade docs pre-7.0 - is that correct?

modules/install/pages/upgrade-cluster-online-full-capacity.adoc
jamesl33: I agree, I think it would be better for us to cover this more comprehensively since there's a fair few options (all with pros/cons along with versioning to be aware of).

PR #2031, opened 2021-07-05
DOC-8736: Update architecture diagrams for Cheshire Cat

simon-dew:
Updates after PM review of Cheshire Cat documentation:

* Update [Index Service Architecture](https://docs-staging.couchbase.com/server/7.0/learn/services-and-indexes/services/index-service.html#index-service-architecture) — add scopes and collections
* Update [Query Sequence](https://docs-staging.couchbase.com/server/7.0/learn/services-and-indexes/services/query-service.html#query-execution) — add Search Service
* Update [Query Service Architecture](https://docs-staging.couchbase.com/server/7.0/learn/services-and-indexes/services/query-service.html#query-service-architecture) — GSI → Indexes
* Add updated diagram source files to docs-server repo in to `modules/learn/assets/source`

✅ Checked that docs site builds without error and source files are not included in output

Docs issue: [DOC-8736](https://issues.couchbase.com/browse/DOC-8736)

PR #2033, opened 2021-07-06
DOC-8736: Update architecture diagrams for Cheshire Cat

simon-dew:
Update description of query execution sequence to include Search Service.


simon-dew:
Further update following #2031

PR #2042, opened 2021-07-09
Fix image sizes

amarantha-k:


amarantha-k:
Removed the 220px specified for images on this page.

PR #2046, opened 2021-07-12
Added externalized files of queries

sixthcodebrewer:
- [X] added queries extracted from the document.
- [X] checked if the queries are in the correct format document.

- Files Added
1. subqueries.adoc

### DIFF
--- a/modules/n1ql/examples/n1ql-language-reference/subquery-1.n1ql
+++ b/modules/n1ql/examples/n1ql-language-reference/subquery-1.n1ql
@@ -1,0 +1,1 @@
+  SELECT t1.city FROM `travel-sample`.inventory.landmark t1 WHERE t1.city IN (SELECT RAW city FROM `travel-sample`.inventory.airport );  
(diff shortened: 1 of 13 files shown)


maria-robobug:
Nice work getting these extracted! 
Just to double check are these going to be referenced in the .adoc file (i.e subqueries.adoc) in a separate PR?

amarantha-k:
@maria-robobug Yes, we'll update the .adoc file to use includes in a separate PR.

maria-robobug (approved):
👍 LGTM

simon-dew:
Good start on this. I've got a couple of comments — @sixthcodebrewer  please can you look at these issues in future pull requests?

* Can you preserve line breaks and indendation in the code? This is very important for readability.
* Naming conventions — can the file names please preserve the name of the file that they're taken from? In this case `subqueries`. We may want to change them later, but this will help maintenance in the initial stages.
* Looking at `subqueries.adoc`, it appears some examples are missing from this pull request — examples Q4A, Q5B, Q5C, Q5D, maybe others. Do you need to refine the process to make sure files aren't missed

PR #2047, opened 2021-07-13
DOC-8789: Changed expiry pager default time to 10 minutes

tonyjhillman:


tonyjhillman:
DOC-8789 corresponds to MB-38741, which is for reducing the expiry pager default from 60 to 10 minutes. If this is okay, I will then do for previous releases (which I think means all since 6.0.x, is that right?) Thanks.

modules/learn/pages/views/views-operation.adoc
```diff
@@ -84,7 +84,7 @@ For instance, if you have ten expired scores still on disk for an average perfor
 However, if you have ten expired scores on disk for an average performed over 20 players, you would get very different result than the average you would expect.
 +
 Run the expiry pager process more frequently to ensure that items that have expired are not included in calculations used in the reduce function.
-We recommend an interval of 10 minutes for the expiry pager on each node of a cluster.
+We recommend the default interval of 10 minutes for the expiry pager on each node of a cluster.
```
owend74: given the interval has been reduced to 10 minutes - I think this should be removed.

modules/learn/pages/views/views-operation.adoc
```diff
@@ -84,7 +84,7 @@ For instance, if you have ten expired scores still on disk for an average perfor
...
 Do note that this interval will have some slight impact on node performance as it will be performing cleanup more frequently on the node.
```
owend74, line 87: Given the reduced default I think that can be removed. As we now run at 10 minutes.

tonyjhillman: Thanks. I've removed this sentence. I'll keep the ticket open for previous versions.

PR #2050, opened 2021-07-13
DOC-8701: Added cbbs rest/http ports

tonyjhillman:


modules/install/pages/install-ports.adoc
```diff
@@ -162,6 +162,13 @@ The following table contains a detailed description of each port used by Couchba
 | Yes
 | No
 
+| `cbbs_http_port` / `cbbs_ssl_port`
```
cg14823: could we change this to `backup_http_port / backup_ssl_port` if possible

tonyjhillman: Okay. I will also change cbbs_backup_grpc_port to backup_grpc_port, if that works.

cg14823 (approved):

PR #2085, opened 2021-07-21
removed AWS ID

RichardSmedley:


simon-dew (approved):
👍 LGTM. Confirmed link still works

PR #2093, opened 2021-07-25
System requirements updates

amarantha-k:
* DOC-6485
* DOC-1758
* DOC-8351

### DIFF
--- a/modules/install/pages/install-production-deployment.adoc
+++ b/modules/install/pages/install-production-deployment.adoc
@@ -61,11 +61,16 @@ xref:deployment-considerations-lt-3nodes.adoc[About Deploying Clusters With Less
 
 == General Guidelines
 
-Use RAID 1+0 (or RAID 1) if Couchbase bucket replication is 1 or lower:: If bucket replicas are set to 2 or higher, the administrator might rely on database-level replication for data durability.
+Disk Configuration:: 
+To reduce the likelihood of disk failure affecting data integrity, consider using a disk configuration such as RAID with mirroring (RAID 1, RAID10) or parity (RAID 5, 6) to ensure that a single disk failure does not lose data.
+
+If bucket replicas are set to 2 or higher, the administrator might rely on database-level replication for data durability.
 
 Linux OS::
 * When deploying Couchbase Server on production Linux, you should use either the XFS or ext4 file system.
-XFS is preferred because it provides better performance when multiple IO streams are accessing the file system, such as during Data Service compaction or when the working set exceeds the memory quota and the data is read from disk.
++
+XFS is preferred because it provides more consistent latency (fairer scheduling) when multiple IO streams are accessing the file system, such as during Data Service compaction or when the working set exceeds the memory quota and the data is read from disk.
+
 * The Couchbase Linux installer requires the ability to create a normal local Unix user with the name `couchbase`, if such a user doesn't already exist.
 This can be particularly important if you are using a directory service for user management, such as LDAP.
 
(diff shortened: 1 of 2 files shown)


modules/install/pages/install-production-deployment.adoc
```diff
@@ -58,3 +58,16 @@ However, you might find a need to have a smaller deployment for test and develop
 
 xref:deployment-considerations-lt-3nodes.adoc[About Deploying Clusters With Less Than Three Nodes]
 |===
+
+== General Guidelines
+
+Use RAID 1+0 (or RAID 1) if Couchbase bucket replication is 1 or lower:: If bucket replicas are set to 2 or higher, the administrator might rely on database-level replication for data durability.
```
daverigby: Not sure what the origin of this recommendation is. While RAID 1 (mirroring) is certainly a reasonable choice, other RAID levels such as 5 or 6 are perfectly valid choices.

Personally I think we should shy away from specific RAID levels here, instead saying something along the lines of:

> "To reduce the likelihood of disk failure affecting data integrity, consider using a disk configuration such as RAID with mirroring (RAID 1, RAID10) or parity (RAID 5, 6) to ensure that a single disk failure does not loose data.

You can keep the second sentence ("If bucket replicas are 2 or higher...").

modules/install/pages/install-production-deployment.adoc
```diff
@@ -58,3 +58,16 @@ However, you might find a need to have a smaller deployment for test and develop
...
+
+Linux OS::
+* When deploying Couchbase Server on production Linux, you should use either the XFS or ext4 file system.
+XFS is preferred because it provides better performance when multiple IO streams are accessing the file system, such as during Data Service compaction or when the working set exceeds the memory quota and the data is read from disk.
```
daverigby: "performance" is a somewhat vague term, and our users might pull us up on it. I think we can be more specific here - something like:

> "XFS is preferred because it provides more consistent latency (fairer scheduling) when multiple IO streams are accessing the file system, such as during Data Service compaction or when the working set exceeds the memory quota and the data is read from disk.

modules/install/pages/pre-install.adoc
```diff
@@ -9,10 +9,10 @@
 | | Minimum Specifications* | Recommended Specifications**
 
 | *CPU*
-| 2 GHz dual core x86_64
-| 3 GHz quad core x86_64 and above
+| 2 GHz dual core 86_64 CPU supporting SSE4.2
```
daverigby: Typo: "86_64" should be "x86_64"

daverigby (changes requested):

modules/install/pages/pre-install.adoc
```diff
@@ -35,3 +35,22 @@ However, you must adhere to the Minimum Specifications for production._
...
+----
+cat /sys/devices/system/clocksource/clocksource0/current_clocksource
+----
++
+Change the clock source using the following commands:
+[source,bash]
+----
+echo tsc > /sys/devices/system/clocksource/clocksource0/current_clocksource
+
+cat /sys/devices/system/clocksource/clocksource0/current_clocksource tsc
```
marcogrecopriolo: 'tsc' is the output of the cat command, so it should be on the next line.

Even better, you could say

Change the clock source using the following command:

echo tsc > /sys/devices/system/clocksource/clocksource0/current_clocksource

To verify the current setting of the clock source, use:

 cat /sys/devices/system/clocksource/clocksource0/current_clocksource

The output should read 'tsc'

amarantha-k: Updated.

modules/install/pages/pre-install.adoc
amarantha-k: Thanks for catching this! Fixed.

modules/install/pages/install-production-deployment.adoc
amarantha-k: Updated.

modules/install/pages/install-production-deployment.adoc
amarantha-k: Updated.

PR #2099, opened 2021-07-27
Eventing updates after initial 7.0 merge

jon-strabala:
Included:

- Lots of tweaks to eventing-language-constructs
- eventing-curl-spec - Removed bullet for "etc" & text " (introduced in 6.5)"
- In eventing-examples-rest-via-curl-get - removed repeated word "implementation"

No need for any changes:

- NO CHANGE (eventing-Terminologies - **the highest is 6.6.2**)
- NO CHANGE (eventing-language-constructs - Note that the pre-6.6.0 argument syntax is still fully supported, but you will not be able to differentiate deletion from expiration. – **this is still valid**)


amarantha-k (approved):
thanks for the updates

PR #2100, opened 2021-07-27
DOC-7657 Migrating data to collections-based model

amarantha-k:


amarantha-k:
In the interest of time, merging this PR so the content can be reviewed on stage.

PR #2113, opened 2021-07-29
Fixed content and hierarchy after the previous merge

ghost:
Fixed content and hierarchy after the previous merge


amarantha-k (approved):
Looks good at a quick glance. Thanks for getting these in.

PR #2135, opened 2021-08-11
DOC-8540: Added information on strict value for cluster encryption-level

tonyjhillman:


modules/learn/pages/security/on-the-wire-security.adoc
```diff
@@ -13,9 +13,17 @@ The interfaces provided by Couchbase Server for configuring on-the-wire security
...
 * _Establishing the Cluster Encryption-Level_.
-The level of encryption imposed on inter-node communications within the cluster can be configured: either as _control_ (meaning that only server-management information is passed in encrypted form) or _all_ (meaning that all information, including data handled by services, is passed in encrypted form).
+The level of encryption imposed on inter-node communications within the cluster can be configured.
+This can be:
+
+** `control`, meaning that server-management information passed between nodes is passed in encrypted form.
+
+** `all`, meaning that all information passed between nodes, including data handled by services, is passed in encrypted form.
+
+** `strict`, meaning the same as `all`, but meaning additionally that only encrypted communication is permitted between nodes and between the cluster and external clients.
```
anuthan: Can we not have "meaning the same as `all`" ? Same as all provides a different understanding. 
Instead it is `all` plus only encrypted communication is permitted. 

Same for others below

modules/manage/pages/manage-nodes/apply-node-to-node-encryption.adoc
```diff
@@ -74,7 +74,10 @@ The output indicates that encryption has been successfully enabled for each node
 
 . Establish an appropriate encryption-level for the cluster.
 This is achieved by the `setting-security` CLI command, specifying the `--cluster-encryption-level` parameter.
-Its value can be either control (meaning that only server-management information is passed in encrypted form) or all (meaning that all information, including data handled by services, is passed in encrypted form):
+Its value can be `control`, meaning that server-management information passed between nodes is passed in encrypted form; `all`, meaning that all information passed between nodes, including data handled by services, is passed in encrypted form; or `strict`, meaning the same as `all`, but meaning additionally that only encrypted communication is permitted between nodes and between the cluster and external clients.
```
anuthan: Same comment as above.

modules/learn/pages/security/on-the-wire-security.adoc
tonyjhillman: Thanks - although I may not quite understand the best rewording. So, perhaps: " `strict`, meaning `all`, but with only encrypted communication permitted...etc" .. something like that?

anuthan: I would leave the "but" out. 
" strict, meaning all with only encrypted communication permitted...etc"

modules/manage/pages/manage-nodes/apply-node-to-node-encryption.adoc
tonyjhillman: I've changed for all to:
`strict`, meaning `all` with only encrypted communication permitted between nodes and between the cluster and external clients.

anuthan (approved):
LGTM

amarantha-k (approved):

RayOffiah (approved):

PR #2165, opened 2021-08-27
DOC-9021: Added statement on effect of revEpoch

tonyjhillman:


tonyjhillman:
Should this be more specific as to exactly which SDK versions are covered?

dave-finlay (approved):

RayOffiah (approved):

PR #2184, opened 2021-09-02
DOC-9078: Added port 19102

tonyjhillman:


amithk:
When node 2 node encryption is enabled, ports 9100, 9101, 9103, 9104 and 9105 become encrypted ports (Note that 9102 still remains unencrypted). I am not sure if this information is relevant in this context.

tonyjhillman:
I have opened DOC-9083 to cover the issue of node-to-node and ports 9100, 9101, 9103, 9104 and 9105. So we can update the port list with this PR, then add the node-to-node info with a subsequent PR. We can then add to other versions, and also cover the 'strict' flag with 7.0.2+.

amarantha-k (approved):

tonyjhillman:
Approved by Shashank.

PR #2201, opened 2021-09-17
DOC-9017: Standard Index Storage for Ephemeral Buckets

tonyjhillman:


ksaikrishnateja (approved):
Should we mention about ForestDB too ? Ephemeral buckets are CE but MOI and Plasma aren't so only option will be ForestDB.

tonyjhillman:
I have added: This applies both to Enterprise and to Community Edition. 

Since the storage is described further below, I think this statement should cover it.

RayOffiah (approved):

PR #2206, opened 2021-09-20
Updated the incorrect Java sdk release notes link

Shweta0489:
Updated the incorrect Java SDK release notes link to https://docs.couchbase.com/java-sdk/current/project-docs/sdk-release-notes.html


simon-dew:
Assigning this to @RichardSmedley to review, but as it seems to be for Couchbase Server 5.1, and there's no equivalent page in the current Couchbase Server docs, you might want to just close as won't fix.

RichardSmedley:
Oh dear.
No point in merging at this stage, but no point in not merging either.
I'm going to put it through, but I doubt anyone finding the page in the archive will use the link with any expectation of finding the current docs tell them the certificates are not supported.

PR #2216, opened 2021-10-01
Doc 9018 document charts on query workbench

RayOffiah:


tonyjhillman (approved):

modules/tools/pages/query-workbench.adoc
```diff
@@ -443,6 +443,50 @@ This simply shows the text output of the {explain}[EXPLAIN] command.
...
+{empty}
+You can change the type of chart, labels and data set by selecting from the three drop-downs marked above.
+
+[horizontal]
+
+1. You can render the chart in one of several formats: `X-Y`, `Connected Point`, `Line`, `Area`, `Bar`, `Group Bar`, `Pie`, and `Donut`.
+
+2. Select a label for your axis. This must be one of the items returned in your query results.
+
+3. The data series to be rendered on the chart.
```
simon-dew: I've used unicode circled numbers as a hack to fake the style of callout lists for images: ➀➁➂. See lines 410-413 above. Not really an issue, It would be good to get the two callout lists to look consistent.

There is an open AsciiDoc issue to be able to create callout lists for images, but it's still in discussion.

simon-dew (approved):
LGTM, just one comment

modules/tools/pages/query-workbench.adoc
RayOffiah: Yes, that's much better. Thanks. I might change the callouts on the diagram so they look similar.

PR #2217, opened 2021-10-01
DOC-9240: remove reference to query tree from query-workbench docs

RayOffiah:


tonyjhillman (approved):

simon-dew (approved):
LGTM, do you want to remove the `query-workbench-result-tree.png` image file also? I can't find any other places where it's used

PR #2220, opened 2021-10-04
DOC-8818: Added note on Eventing Service not covered by node-to-node all

tonyjhillman:


jon-strabala:
LGTM

modules/learn/pages/clusters-and-availability/node-to-node-encryption.adoc
```diff
@@ -17,6 +17,9 @@ When node-to-node encryption is first enabled, this is the default.
 
 * _All_. Both cluster-management information and data is passed in encrypted form.
 Node-to-node encryption can be supported by means of either the Couchbase-provided certificates with which nodes are protected by default; or certificates installed by the administrator.
++
+Note, however, that node-to-node encryption has _not_ been implemented for the Eventing Service in versions of Couchbase Server prior to 7.0.2.
+Therefore, on such earlier versions, when _all_ is specified, whereas the data passed to and from all other Couchbase Services is passed in encrypted form, the data passed to and from the Eventing Service continues to be passed in _unencrypted_ form.
```
RayOffiah:
```suggestion
Therefore, on earlier versions, when _all_ is specified, the data passed to and from all other Couchbase Services is passed in encrypted form, whereas the data passed to and from the Eventing Service continues to be passed in _unencrypted_ form.
```

RayOffiah (changes requested):

modules/learn/pages/clusters-and-availability/node-to-node-encryption.adoc
jon-strabala: Sorry for the late update here ...  as such there will be a bit more rewrite.
1. Eventing supports node-to-node encryption for starting in version 6.6.5 and also starting in version 7.02
2. Note that all Eventing Functions must be paused or undeployed when enabling TLS to prevent possible mutation loss.

Item 1 is due to the back porting node-to-node encryption for Eventing to 6.6.5 _Refer to MB-48048. [B.P. to 6.6.4] Backport of Enforce TLS - Eventing._  So 7.0.0 and 7.0.1 don't support encryption as such we need to discuss both version trains.

Item 2 occurs because on an upgrade to TLS KV will close and restart the DCP streams to enable TLS thus any buffered events/mutations would be dropped.

tonyjhillman:
I'm closing this and restarting under DOC-9619.

PR #2225, opened 2021-10-06
Doc 9058 and 9181: document support for auto failover of index service; updates for graceful and concurrent failover

RayOffiah:


menih:
* We need to cover co-located data/index services. data service state always takes precedence over indexer.
* Indexer failover safety will be best effort based on the health state reported at the time we decide to failover.

menih:
@vzasade please review and add your comments

RayOffiah:
> * We need to cover co-located data/index services. data service state always takes precedence over indexer.
> * Indexer failover safety will be best effort based on the health state reported at the time we decide to failover.

I think this might have been covered, but I'll add a note to make it clearer. How about this:

image

RayOffiah:
Would be great if @jeelanp2003 and/or @menih  could have another quick look at the changes.

RayOffiah:
Awaiting further reviews.

cherkauer-couchbase (commented):
AFAIK this is correct but really ns_server needs to review because that is the component making the actual Autofailover decisions.

@vzasade would be the correct person for that.

cherkauer-couchbase (commented):
"Disk read/write failure. Attempts to read from or write to disk on a particular node have resulted in a significant rate of failure, for longer than a specified time-period. The node is removed by auto-failover, even though the node continues to be contactable."

This part is not yet implemented for GSI.

modules/learn/pages/clusters-and-availability/automatic-failover.adoc
```diff
@@ -83,9 +83,9 @@ The node-minimum for each service is provided in the following table:
 
 | Query
 | 2
-
+// TODO: Check the number for the index.
 | Index
-| Not supported
+| 3
```
vzasade: 2

modules/learn/pages/clusters-and-availability/automatic-failover.adoc
```diff
@@ -108,10 +108,10 @@ This is illustrated by the following examples:
 | Node | Services Hosted
 
 | #1
-| Data
+| Data & Index
```
vzasade: I don't think including Index in this example is needed

modules/learn/pages/clusters-and-availability/automatic-failover.adoc
```diff
@@ -171,7 +171,7 @@ In this case, even though the Query and Search Services were both running on onl
 If node #1, #2, or #3 becomes unresponsive, auto-failover can be triggered.
 In each case, this is due to _Data Service Preference_, which applies auto-failover based on the policy for the Data Service, irrespective of other services on the unresponsive node.
 Note that in the case of node #2, this allows an Index Service node to be automatically failed over.
-However, if node #4 becomes unresponsive, auto-failover is not triggered; since auto-failover is not supported for the Index Service, unless Data Service Preference applies.
+If node #4 becomes unresponsive, auto-failover is triggered as there are four nodes (The threshold for the index service is 3.)
```
vzasade: threshold for index service is 2, as for all other services

vzasade:
>> Group failure. An administrator-defined group
We removed that feature in Neo. Instead the autofailover can fail over multiple nodes at the same time regardless of them belonging to configured group

vzasade:
>> Only on the occurrence of one event at a time

Not true anymore. If multiple nodes are down at the same time, we will fail over multiple nodes, till the maximum of allowed failed over nodes is reached.

vzasade:
>> The highest permitted maximum is 3

We are removing this limit

modules/learn/pages/clusters-and-availability/automatic-failover.adoc
RayOffiah: Removed & Index

cherkauer-couchbase (commented):
"Note that in the case of node #2, this allows an Index Service node to be automatically failed over."

This verbiage that implies this is the only way for Index to be automatically failed over is leftover from when that was the case. Maybe should be tweaked or just removed now that Index can fail over on its own.

modules/learn/pages/clusters-and-availability/automatic-failover.adoc
```diff
@@ -1,7 +1,7 @@
 = Automatic Failover
 :description: A node or group can be failed over automatically when it either becomes unresponsive or experiences continuous disk-access problems.
```
shivaniguptacb: Replace with: One or more nodes, up to a maximum of less than half the cluster (i.e. (nodes +1)/2 -1 ), can be failed over automatically when the nodes become unresponsive or experience continuous disk problems.

shivaniguptacb (commented):
Overall the changes look good. Just have one comment for the overall description of Automatic Failover.

shivaniguptacb (changes requested):
See the change for the headline of Automatic Failoverr.

tonyjhillman (approved):

modules/learn/pages/clusters-and-availability/automatic-failover.adoc
```diff
@@ -19,26 +20,16 @@ For information on managing auto-failover, see the information provided for Couc
...
 
 [#auto-failover-constraints]
 == Auto-Failover Constraints
 
 Auto-failover is triggered:
 
-* Only on the occurrence of one event at a time.
-If multiple events are concurrent, auto-failover is not triggered.
+* If the majority of nodes (stem:[(nodes + 1)/2]) in the cluster can form a quorum to initiate failover following the failure of one the nodes.
 * Sequentially, only up to an administrator-specified maximum number of events.
```
shivaniguptacb, line 32: Change to 'Sequentially, only up to an administrator-specified maximum number of nodes.'

modules/learn/pages/clusters-and-availability/automatic-failover.adoc
```diff
@@ -19,26 +20,16 @@ For information on managing auto-failover, see the information provided for Couc
...
-The highest permitted maximum is 3.
```
shivaniguptacb, line 41: This line should be removed. We don't have this limit anymore.

modules/learn/pages/clusters-and-availability/automatic-failover.adoc
```diff
@@ -19,26 +20,16 @@ For information on managing auto-failover, see the information provided for Couc
...
+The autofailover can fail over multiple nodes at the same time.
 
 [#auto-failover-constraints]
 == Auto-Failover Constraints
 
 Auto-failover is triggered:
 
-* Only on the occurrence of one event at a time.
-If multiple events are concurrent, auto-failover is not triggered.
+* If the majority of nodes (stem:[(nodes + 1)/2]) in the cluster can form a quorum to initiate failover following the failure of one the nodes.
```
shivaniguptacb, line 31: Modify to:
If the majority of nodes (stem:[(nodes + 1)/2]) in the cluster can form a quorum to initiate failover following the failure of one or more nodes. Multiple concurrently failing nodes will be auto-failed over as long as the administrator specified limit is not exceeded.

modules/learn/pages/clusters-and-availability/automatic-failover.adoc
shivaniguptacb: One or more nodes can be failed over automatically when the nodes become unresponsive or experience continuous disk problems. This is done as long as data safety is maintained i.e. no data will be lost.

RayOffiah:
This one's a bit of a mess. Closing without merging

PR #2263, opened 2021-11-02
page-alias warning

RichardSmedley:
This alias should occur only once in the docs - preferably in the latest branch


simon-dew (approved):
Fantastic 👍

PR #2272, opened 2021-11-08
DOC-9145: Query for callsign on airport collection doesn't work

simon-dew:
* Fix query for callsign on airport collection → should be airline
* Also fix source language markup

Docs issue: [DOC-9145](https://issues.couchbase.com/browse/DOC-9145)

### DIFF
--- a/modules/getting-started/pages/try-a-query.adoc
+++ b/modules/getting-started/pages/try-a-query.adoc
@@ -28,7 +28,7 @@ Here's an example of a basic N1QL query and the JSON document it returns.
 
 With the query_context set to `travel-sample.inventory`, the following query asks for the country that is associated with the airline _Excel Airways_:
 
-[source,sql]
+[source,n1ql]
 ----
 SELECT a.country FROM airline a WHERE a.name = "Excel Airways";
 ----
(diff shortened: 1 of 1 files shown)


modules/getting-started/pages/try-a-query.adoc
```diff
@@ -57,30 +57,30 @@ The country is thus specified as `United Kingdom`.
...
 
 This displays the `cbq` shell prompt, against which you can enter N1QL commands, specifying your currently installed buckets.
-For example, the following query returns the different values that are used by the documents in the `airport` collection for the `callsign` field, limiting the number of results to five:
+For example, the following query returns the different values that are used by the documents in the `airline` collection for the `callsign` field, limiting the number of results to five:
 
-[source,sql]
+[source,shell]
 ----
-cbq> SELECT a.callsign FROM default:`travel-sample`.inventory.airport  a LIMIT 5;
+cbq> SELECT a.callsign FROM default:`travel-sample`.inventory.airline  a LIMIT 5;
```
osfameron, line 78: the current formatting as "sql" actually looks OK.
Would it make more sense to format as `n1ql` given that *most* of the line could be usable as n1ql?
(Just as the command *isn't* actually valid out-of-context as "shell", so this change doesn't seem to make it more correct?)

osfameron (approved):
Minor niggle inline, can ignore :D
Looks good 👍

modules/getting-started/pages/try-a-query.adoc
simon-dew: Thanks for the review! I _thought_ that when the source language is set to shell, the **Copy** button would ignore the `cbq>` prompt, which users wouldn't actually want. But further investigation seems to show that this isn't actually the case. I'll leave the source language as shell for now and look into this further.

simon-dew: Inspecting the source... highlight.js correctly identifies the prompt, and wraps it in a span with class `hljs-meta`. My understanding was that the Copy button is supposed to ignore spans with class `hljs-meta`, so I'm not sure why this isn't working.

PR #2273, opened 2021-11-09
Added missing release note detail for MB-32497

RayOffiah:
Added missing release note detail

### DIFF
--- a/modules/release-notes/pages/relnotes.adoc
+++ b/modules/release-notes/pages/relnotes.adoc
@@ -416,7 +416,7 @@ This section highlights some of the known issues in this release.
 | *Summary*: When a synchronous durability request is made unrelated operations pipelined on the same socket may be serviced with higher latency, due to https://issues.couchbase.com/browse/MB-10291[MB-10291].
 
 | https://issues.couchbase.com/browse/MB-32497[MB-32497^]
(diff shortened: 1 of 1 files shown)


rdemellow (commented):
Think we need to add more explanation to what the user might see e.g.
"Memcached connections may hang due to Connection::signalIfIdle triggering a deadlock when used from a different thread" Think adding the words connection and hang will help to make it easier for people when searching though the release notes :)

rdemellow (approved):

modules/release-notes/pages/relnotes.adoc
```diff
@@ -1,7 +1,7 @@
 = Release Notes for Couchbase Server 6.5
 
 Couchbase Server 6.5 brings many sought-after relational capabilities to the NoSQL world.
-This release seeks to improve developer agility, reduce total cost of ownership, enhance manageability, and provide more enterprise-grade security and performance.
```
osfameron, line 4: lots of whitespace changes, may be worth turning this feature off in VSCode, to avoid making the diffs noisy.

osfameron (approved):
Looks good 👍 
(just a niggle about whitespace, only because the actual change was hard to find, 😄 )

As this looks like it's a Known Issue, the wording looks great to me.

PR #2279, opened 2021-11-15
DOC-9133: New Sync Gateway Roles

tonyjhillman:


tonyjhillman:
The descriptions and tables I've drafted are based on the information 
in https://github.com/couchbase/ns_server/blob/master/src/menelaus_roles.erl. I have
simplified as seems appropriate, because some of the distinctions appear very 
low-level. 

In each case, I've said that the role gives no access to the web console. But I've
not been able to verify, because in the latest build, most of the roles are missing
from the security panel - as if they are being redeveloped.

In some cases, the description says that 'users' or 'users and roles' can be managed.
In each case, I've assumed this means 'users, groups, and roles'.

Sync Gateway Architect: The description is that this role 'can manage Sync Gateway 
databases'. But it's not clear to me what a 'Sync Gateway database' is - the actual 
roles reference 'collection' (not 'bucket'). That said, each reference to 'collection' 
seems to specify 'none' as access; which might mean there's actually no access to data,

If I can get a reference to a suitable section in the Sync Gateway docs, I can
include a link.

modules/learn/pages/security/roles.adoc
```diff
@@ -1097,6 +1097,191 @@ The user can, by means of Sync Gateway, read and write data, manage indexes and
...
+^| image:introduction/yes.png[]
+^| image:introduction/no.png[]
+^| image:introduction/no.png[]
+^| image:introduction/no.png[]
+|===
+
+[#sync-gateway-application]
+== Sync Gateway Application
+
+The *Sync Gateway Application* role (which is a _Mobile_ role) allows management of Sync Gateway users, groups, and roles; and allows application data to be read and written through Sync Gateway.
```
rajagp: @tonyjhillman  General (applies to all roles) : Remove "Groups" from "Sync Gateway users, groups, and roles".

modules/learn/pages/security/roles.adoc
```diff
@@ -1097,6 +1097,191 @@ The user can, by means of Sync Gateway, read and write data, manage indexes and
...
+The role does not allow access to Couchbase Web Console.
```
rajagp, line 1147: @tonyjhillman  Have to check on this one (for all roles). Not quite sure if our implementation of roles enforces this

rajagp (approved):

PR #2286, opened 2021-11-18
DOC-8564#7.0 Removed note saying analytics port is not in use.

RayOffiah:


b33f (approved):

simon-dew (approved):
Seems like a convoluted way of doing footnotes, but it works! It is correct that the encrypted Analytics port is currently in use, so LGTM 👍

RayOffiah:
> Seems like a convoluted way of doing footnotes, but it works! It is correct that the encrypted Analytics port is currently in use, so LGTM 👍

Yes, footnotes a little bit weird. Folk set them up as attributes to make docs easier to read, otherwise you get this sort of thing:

this is the textfootnote:[This is the footnote for text]

PR #2291, opened 2021-11-24
DOC-9444: Release notes showing Debian 11.x support

RayOffiah:
Reverted changed line. We're not supporting Debian 11.x until 7.1


RichardSmedley (approved):
:shipit:

PR #2293, opened 2021-11-24
DOC-9445: Multiple Root CA Overview

tonyjhillman:


modules/learn/pages/security/using-multiple-cas.adoc
```diff
@@ -0,0 +1,72 @@
...
+:description: Couchbase Server supports use of multiple CA (or 'root') certificates, for a single cluster.
+
+[abstract]
+{description}
+
+== Benefits
+
+Prior to Couchbase Server Version 7.1, a Couchbase-Server cluster could maintain at most _one_ root certificate: all nodes and all clients were obliged to rely on this certificate's authority.
+From version 7.1, Couchbase Server permits multiple root certificates to maintained in a _trust store_ for the cluster.
+This:
```
timofey-barmin: Please make sure the following use-cases are covered here.
Couchbase-server uses "trusted certificate list" to:

- verify client certificates (when client certificates authentication is enabled)
- verify identities of cluster nodes when n2n encryption is enabled
- verify identities of nodes that join cluster(when cluster is provisioned with certificates; when self-generated certificates are used we don't verify new-node's identity);
- verify identity of LDAP servers (when TLS is turned on in LDAP settings)

please let me know if you have questions.

modules/learn/pages/security/using-multiple-cas.adoc
```diff
@@ -0,0 +1,72 @@
...
+
+Multiple CAs are supported by Couchbase Server Version 7.1 and later.
+Typically, when a cluster is upgraded, it temporarily includes nodes of two versions; one being the version _to_ which the cluster is being upgraded, the other the version _from_ which it is being upgraded.
+
+Once the cluster's nodes are all upgraded to version 7.1 or later, multiple CAs are fully supported.
+During the upgrade period, only _some_ of the supportive APIs can be used: for information, see xref:rest-api:rest-certificate-management.adoc[Certificate Management API].
+
+== Adding Nodes
+
+When a new node is added to a Couchbase-Server cluster that is version 7.1 or later, the new node may be added with its existing root certificate and corresponding certificate chain, provided that the root certificate is already included in the cluster's trust store: if the node's certificate has not been added to the trust store, an error is flagged.
```
timofey-barmin, line 75: We can also mention that new node's trust store is cleaned when the node is joining (not merged). To be more precise: new node's trusted store can be used for initial connection establishment (right before join) to the cluster (might be needed when join is initiated from the new node). But a bit later, when the node is actually joining the cluster, its trust store is overwritten with cluster's trust store. 

I'll try to explain in example:
Given: Node-to-be-added joins Cluster

Cluster's trust store: ca1, ca2, ca3
Cluster node1: cert1 signed by ca1

Node-to-be-added's trusted store: ca1, ca2, ca4
Node-to-be-added's node cert: cert2 signed by ca2

After join new cluster's trust store will contain: ca1, ca2, ca3 (stays the same basically).
ca4 will not be added.

modules/learn/pages/security/using-multiple-cas.adoc
tonyjhillman: I have added these cases.

modules/learn/pages/security/using-multiple-cas.adoc
tonyjhillman: I have added.

A couple of questions. First, what if cert2 had been signed by CA 4?

Secondly (and this relates to a comment you have made elsewhere): there is a distinction between 'adding a node to a cluster' and 'joining a node to a cluster'. As a cluster, I appoint an external node that is to be added. As an external node, I appoint a cluster to which I join. But I believe that a join can only occur before provisioning - which would mean, before the addition of certificates. Therefore, a node with uploaded certs can only be added, not joined. Is that correct?

timofey-barmin: > First, what if cert2 had been signed by CA 4?

Then addition will not work, because cluster doesn't trust CA4


> adding a node to a cluster' and 'joining a node to a cluster'. 

I think you mean that the "join" can be initiated either by node-to-be-added or by the cluster.
In other words you can start the process either from cluster, or from node-to-be-added. 
Technically both of them are almost the same. The only difference is one extra http call for the case when the "join" is initiated from node-to-be-added.


> But I believe that a join can only occur before provisioning - which would mean, before the addition of certificates.

Not sure what you call provisioning here.
There are two options:
1. If your cluster uses self-generated certs, then nothing needs to be done before joining.
2. If your cluster uses custom certs, you only need to load certificates to the node-to-be-added before joining (note that this is not equal to initializing the cluster or initializing node). Upload CA and node cert to the inbox dir and call /loadCAcerts and /reloadCertificate (CLI should be able to do that as well).

Those things are true with no regards if the addition is initiated from cluster side or from node-to-be-added side.

modules/learn/pages/security/using-multiple-cas.adoc
```diff
@@ -8,15 +8,18 @@
 
 Prior to Couchbase Server Version 7.1, a Couchbase-Server cluster could maintain at most _one_ root certificate: all nodes and all clients were obliged to rely on this certificate's authority.
 From version 7.1, Couchbase Server permits multiple root certificates to maintained in a _trust store_ for the cluster.
-This:
-
-* Allows an individual node either to use a CA that is also used by one or more other nodes; or to use a CA that is used by no other node.
+This allows an individual node either to use a CA that is also used by one or more other nodes; or to use a CA that is used by no other node.
 This may be used to increase security: ensuring that a node can be successfully authenticated by some applications, but not by others.
```
timofey-barmin: This may be used during CA certificate rotation I would say.
You upload New CA, then change nodes certificates one by one, then remove old CA.

modules/learn/pages/security/using-multiple-cas.adoc
tonyjhillman: Thank you. I have added a note on addition not working when cert 2 is signed by CA 4:

Note also that if the node certificate for _node x_, _Cert 2_, had been signed by _CA 4_ (instead of _CA 2_), the addition of _node x_ would fail; because the cluster does not trust _CA 4_.

modules/learn/pages/security/using-multiple-cas.adoc
tonyjhillman: I have changed to:

This may be used during CA certificate rotation: a new CA is uploaded, node certificates are changed one by one, and finally, the old CA is removed.

timofey-barmin (approved):

timofey-barmin (approved):

PR #2296, opened 2021-11-29
Doc 9221 6.5 information on alternate address should include note about port forwarding

RayOffiah:


RayOffiah:
I think I needed to pull this branch first!

PR #2297, opened 2021-11-29
DOC-9221 information on alternate address should include note about port forwarding

RayOffiah:
Added note about port forwarding

### DIFF
--- a/modules/learn/pages/clusters-and-availability/connectivity.adoc
+++ b/modules/learn/pages/clusters-and-availability/connectivity.adoc
@@ -59,6 +59,13 @@ For information on using the CLI to change the address family for an existing cl
 Couchbase Server allows an _alternate address_ to be assigned to any individual cluster-node, and an _alternate port number_ to be assigned to any service running on that node.
 For a list of services, and the standard ports they occupy, see xref:install:install-ports.adoc[Network and Firewall Requirements].
 
+Before assigning an alternative address, port forwarding must be enabled on the host machine. For example:
+[source, console]
+----
+iptables -A PREROUTING -t nat -i eth0 -p tcp --dport 9000 -j REDIRECT --to-port 8093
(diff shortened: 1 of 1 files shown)


RayOffiah:
I'm just doing the one for 6.5 initially. Once that's okay, I'll roll the same change into 6.6 and 7.0

modules/learn/pages/clusters-and-availability/connectivity.adoc
```diff
@@ -59,6 +59,13 @@ For information on using the CLI to change the address family for an existing cl
 Couchbase Server allows an _alternate address_ to be assigned to any individual cluster-node, and an _alternate port number_ to be assigned to any service running on that node.
 For a list of services, and the standard ports they occupy, see xref:install:install-ports.adoc[Network and Firewall Requirements].
 
+Before assigning an alternative address, port forwarding must be enabled on the host machine. For example:
```
hyunjuV: Since port forwarding isn't necessary unless you're using different port numbers with the alternate address, and since when you do the port forwarding doesn't matter, it might be better to tweak the wording on line 62 a bit.  Also, since the example in the blog specifically stated that the command was used on a virtual machine, we should include that clause since it may be different in other scenarios.   So, my suggested change for line 62 would be:

When assigning an alternative address and using different port numbers than the expected, port forwarding must be enabled on the host machine.  For example, on a virtual machine:

hyunjuV (changes requested):
Just the suggested change for line 62.

modules/learn/pages/clusters-and-availability/connectivity.adoc
RayOffiah: Will do! Thanks for the look over.

tonyjhillman (approved):

PR #2303, opened 2021-11-30
DOC-9221-7.0 information on alternate address should include note about port forwarding

RayOffiah:
Adding info on alternate addresses


tonyjhillman (approved):
Only thing I'd suggest is that you have used 'alternative' in line 70, so you might wish to you 'alternate' instead, since that is standard (although I have always preferred 'alternative' myself).

PR #2306, opened 2021-12-01
DOC-9319a: New certs API pages

tonyjhillman:


modules/rest-api/pages/delete-trusted-cas.adoc
```diff
@@ -0,0 +1,67 @@
...
+
+----
+DELETE /pools/default/trustedCAs/<id>
+----
+
+[#description]
+== Description
+
+Removes from the list of trusted CA certificates for the cluster the certificate with the specified id.
+The id can be retrieved by means of the `GET /node/controller/trustedCAs` method and URI: see xref:rest-api:get-trusted-cas.adoc[Get Root Certificates].
```
timofey-barmin: `GET /node/controller/trustedCAs` and xref:rest-api:get-trusted-cas.adoc are the same, no?

modules/rest-api/pages/load-trusted-cas.adoc
```diff
@@ -0,0 +1,104 @@
...
+
+Loads trusted CA (or _root_) certificates that reside on the accessed node into the Couchbase-Server cluster.
+All loaded CA certificates can be accessed by all nodes.
+Loaded certificates can be used to provide authority to the cluster's nodes, and can be used to authenticate clients' access-attempts.
+Either the Full Admin or the Security Admin role is required.
+
+Note that use of multiple root certificates is not possible prior to Couchbase Server Version 7.1.
+The `POST /node/controller/loadTrustedCAs` method and URI can therefore only be used on clusters all of whose nodes are running at least version 7.1: it cannot be used if _any_ cluster node is running a version prior to 7.1.
+
+Prior to being loaded, the trusted CA certificates must be copied into a folder named `inbox/CA`; which has been manually created by the administrator on a given node, in the appropriate platform-specific location.
```
timofey-barmin: Please mention the following:
- it loads CA certificates from all the files in inbox/CA
- certificates must be in pem format
- each file may contains several certs
- files that start from "." are ignored.

We can also mention 'for security reasons' as a reason for why it has to be put into inbox/CA, rather than attaching it to an http request.

modules/rest-api/pages/load-trusted-cas.adoc
```diff
@@ -0,0 +1,104 @@
...
+An incorrect method returns `HTTP/1.1 405 Method Not Allowed`.
+
+[#example]
+== Example
+
+The following example shows the creation of the `inbox/CA` folders, the copying of a trusted CA certificate into the destination `CA` folder, and the loading of the certificate by means of the REST API.
+
+----
+sudo mkdir -p /opt/couchbase/var/lib/couchbase/inbox/CA
+sudo cp ca.pem /opt/couchbase/var/lib/couchbase/inbox/CA/.
```
timofey-barmin: Dot at the end doesn't make any sense

modules/rest-api/pages/rest-certificate-management.adoc
```diff
@@ -9,33 +9,30 @@
...
+From a management perspective, server certificates can be considered to be of two kinds:
 
 * _Root_ certificates.
-A single root certificate exists for each cluster.
-This certificate, which is sometimes referred to as the _cluster_ certificate, contains the public key of a Certificate Authority (CA).
-Programs that wish to interact securely with Couchbase Server must elect to trust this CA.
+At least one root certificate exists for each cluster.
+Any number of root certificates can be uploaded: together, these constitute the cluster's _trust store_.
+Each root certificate contains the public key of a Certificate Authority (CA).
+Programs that wish to interact securely with Couchbase Server must elect to trust the CA referenced by one of its root certificates.
```
timofey-barmin: > Programs that wish to interact securely with Couchbase Server must elect to trust the CA referenced by one of its root certificates.

I would say it's not fully correct or at least confusing.
I suggest we elaborate a little bit here because the difference is not obvious for many people.
The meaning is rather the following: 
Couchbase-server uses "trusted certificate list"  to:
- verify client certificates (when client certificates authentication is enabled)
- verify identities of cluster nodes when n2n encryption is enabled
- verify identities of nodes that join cluster cluster (when server is provisioned with certificates)
- verify identity of LDAP servers (when TLS is turned on in LDAP settings)

Please let me know if you need help with it.

modules/rest-api/pages/rest-certificate-management.adoc
```diff
@@ -9,33 +9,30 @@
...
-See xref:rest-api:upload-retrieve-root-cert.adoc[Upload and Retrieve the Root Certificate].
+* Root certificates can be uploaded, retrieved, and deleted.
+See xref:rest-api:load-trusted-cas.adoc[Load Root Certificates], xref:rest-api:get-trusted-cas.adoc[Get Root Certificates], and xref:rest-api:delete-trusted-cas.adoc[Delete Root Certificates].
 
-* The current certificate for a specific node can be retrieved and examined.
-A new node certificate can be uploaded.
+* The current certificate for a specific node can be uploaded and retrieved.
 See xref:rest-api:upload-retrieve-node-cert.adoc[Upload and Retrieve a Node Certificate].
-
-* All certificates &#8212; root and node &#8212; can be _regenerated_ (that is, restored to their automatically provided default values).
```
timofey-barmin, line 40: Not sure why "regeneration" is removed here.

modules/rest-api/pages/upload-retrieve-node-cert.adoc
```diff
@@ -12,19 +12,24 @@
...
-Note that such retrieval can only be performed with an administrator-configured node certificate: it cannot be performed with the default node certificate that is automatically provided to a node by Couchbase Server on initial configuration.
+The `GET` method and `/pools/default/certificate/node/<ip-address-or-domain-name>` URI allow an administrator-configured node-certificate to be retrieved.
+Note that such retrieval can only be performed with an administrator-configured node-certificate: it cannot be performed with the default node-certificate that is automatically generated for a node by Couchbase Server on initial configuration.
+
+Note that the `POST` API _can_ be used on clusters one or more of whose nodes is running a version of Couchbase Server prior to 7.1.
+The `GET` API can likewise be used: however, node-certificates for pre-7.1 nodes are not returned.
 
 Both calls require either the Full Admin or the Security Admin role.
+For the loading of the node-certificate to succeed, the private key must be owned by user `couchbase`.
```
timofey-barmin: > For the loading of the node-certificate to succeed, the private key must be owned by user `couchbase`.

This is not fully correct. Rather "user 'couchbase' should have 'read' access to chain and private key files". 
The thing is it can be achieved by different ways, so it's not correct to say that the file must be owned by user couchbase.

modules/rest-api/pages/upload-retrieve-node-cert.adoc
```diff
@@ -34,24 +39,95 @@ The curl syntax for these calls is as follows:
...
-The copy of the node certificate returned by the `GET` method should be saved appropriately: for example, by means of `> <node-certificate-copy>`, on the local filesystem.
+=== JSON Passphrase Registration
+
+If a node-certificate to be loaded is associated with an encrypted private key, a procedure can be defined to allow Couchbase Server to access and use the key's passphrase, when use of the key is required: the passphrase can be _registered_, by means of a JSON object, specified as the `json-passphrase-registration` argument.
+
+Note that if the node-certificate has already been loaded, in order to register the passphrase for its key, the certificate can be _reloaded_, with the `json-passphrase-registration` specified: for this to be successful, the corresponding `pem` and `key` files must both still be within the created `inbox`, or must be recopied into it.
+
+The following examples show how to register a passphrase in association with an encrypted private key.
+
+==== Plain
```
timofey-barmin: Since this option is unsecure should we make it the last option in the list?

modules/rest-api/pages/upload-retrieve-node-cert.adoc
```diff
@@ -34,24 +39,95 @@ The curl syntax for these calls is as follows:
...
+----
+{"privateKeyPassphrase": {"type": "script",
+                          "path": "<path to script>",
+                          "args": ["arg1", "arg2"],
+                          "trim": false,
+                          "timeout": 5000}}
+----
+
+The specified `type` is `script`.
+The value of `path` is the path to the script on the current node: this must be in the `data/script` directory.
```
timofey-barmin: We can add 'for security reasons':

"for security reasons the script must be in `data/script` directory."

because this limitation might raise questions, so it would be cool to give some answers right away

modules/rest-api/pages/upload-retrieve-node-cert.adoc
```diff
@@ -34,24 +39,95 @@ The curl syntax for these calls is as follows:
...
 An incorrectly specified IP address or domain name causes the attempted connection to time out, with a `Failed to connect` notification.
 
 If the `POST` is called with either no node certificate or no node-private key in the `inbox` directory, the call fails with `400 Bad Request`, and either the notification `"Unable to read private key file /opt/couchbase/var/lib/couchbase/inbox/pkey.key. The file does not exist."`; or the notification `"Unable to read certificate chain file /opt/couchbase/var/lib/couchbase/inbox/chain.pem. The file does not exist."`.
 
-If the `GET` method is used with no administrator-configured node certificate having been uploaded, the call fails with `404 Object Not Found`, and the notification `Certificate is not set up on this node`.
+If the `POST` is called with the private key for the node-certificate not owned by user `couchbase`, the call fails with an error message such as the following: `"Unable to read private key file /opt/couchbase/var/lib/couchbase/inbox/pkey.key. Missing permission for reading the file, or for searching one of the parent directories."`
```
timofey-barmin: See my comment above.
Rather something like:
if server doesn't have 'read' permission for chain file or private key file...

modules/rest-api/pages/upload-retrieve-node-cert.adoc
```diff
@@ -34,24 +39,95 @@ The curl syntax for these calls is as follows:
...
+In such a case, ownership of the private key should be changed to user `couchbase`.
```
timofey-barmin: See my comment above about 'read' permission

modules/rest-api/pages/upload-retrieve-node-cert.adoc
```diff
@@ -34,24 +39,95 @@ The curl syntax for these calls is as follows:
...
+
+If the `GET` method is used with no administrator-configured node-certificate having been uploaded, the call fails with `404 Object Not Found`, and the notification `Certificate is not set up on this node`.
```
timofey-barmin: This is actually the pre-7.1 behavior. In 7.1 the GET will show the "generated" certificate instead ("type": "generated").
Please let me know if you need more information.

modules/rest-api/pages/upload-retrieve-node-cert.adoc
```diff
@@ -72,18 +148,22 @@ curl -v -X GET http://10.143.201.101:8091/pools/default/certificate/node/10.143.
 -u Administrator:password
 ----
 
-If successful, the call returns `200 OK`, and an object whose fields specify the Subject Common Name, expiry date, and pem-encoded contents of the node certificate:
+If successful, the call returns `200 OK`, and an object whose fields specify warnings, the Subject Common Name, the expiry date, the type (whether _generated_ by Couchbase Server, or _uploaded_ by the administrator), the pem-encoded contents of the node certificate, and the passphrase for the private key (securely displayed as asterisks), if one has been established:
```
timofey-barmin: securely displayed as asterisks - only when plain password is used

modules/rest-api/pages/upload-retrieve-root-cert.adoc
```diff
@@ -21,15 +26,18 @@ GET /pools/default
 [#description]
 == Description
 
-The `POST` method and `/controller/uploadClusterCA` URI upload an appropriately configured certificate to the cluster, so that it becomes the root certificate for the cluster.
+The `POST` method and `/controller/uploadClusterCA` URI (which are deprecated in 7.1) upload an appropriately configured certificate to the cluster, so that it becomes the root certificate for the cluster.
 Either the Full Admin or the Security Admin role is required.
 The certificate must be specified as binary data, in a location on the filesystem, for successful upload to occur.
+In versions 7.1 and after, by default, this method and URI can be used on _localhost_ only.
```
timofey-barmin: For security reasons

modules/rest-api/pages/upload-retrieve-root-cert.adoc
```diff
@@ -21,15 +26,18 @@ GET /pools/default
...
+However, this default setting can be changed, by means of the `POST` method and `/settings/security/allowNonLocalCACertUpload` URI, with the `-d false` flag.
+The status of the current setting can be retrieved with the `GET` method and `/settings/security/allowNonLocalCACertUpload` URI.
 
-The `GET` method and `/pools/default/certificate` URI make a copy of the cluster's current root certificate, and returns the copy.
+The `GET` method and `/pools/default/certificate` URI (which are deprecated in 7.1) make a copy of the cluster's current root certificate, and return the copy.
```
timofey-barmin: It's a bit tricky with this endpoint.
It's supposed to return one certificate, but in 7.1 it's possible to have several trusted certificates, so it's not clear what to return for backward compatibility.
Currently it works the following way in 7.1: after upgrade it continues to return the certificate that was used before the upgrade, but once at least one cert is uploaded it starts to return "400 Bad Request"  "deprecated, please use /pools/default/trustedCAs"

modules/rest-api/pages/upload-retrieve-root-cert.adoc
```diff
@@ -21,15 +26,18 @@ GET /pools/default
 [#description]
 == Description
 
-The `POST` method and `/controller/uploadClusterCA` URI upload an appropriately configured certificate to the cluster, so that it becomes the root certificate for the cluster.
+The `POST` method and `/controller/uploadClusterCA` URI (which are deprecated in 7.1) upload an appropriately configured certificate to the cluster, so that it becomes the root certificate for the cluster.
 Either the Full Admin or the Security Admin role is required.
 The certificate must be specified as binary data, in a location on the filesystem, for successful upload to occur.
+In versions 7.1 and after, by default, this method and URI can be used on _localhost_ only.
+However, this default setting can be changed, by means of the `POST` method and `/settings/security/allowNonLocalCACertUpload` URI, with the `-d false` flag.
```
timofey-barmin: It's also makes sense to mention POST /settings/security/allowNonLocalCACertUpload` can be called from localhost only (for security reasons)

modules/rest-api/pages/upload-retrieve-root-cert.adoc
```diff
@@ -42,6 +50,13 @@ curl -X POST
   http://<ip-address-or-host-name>:8091/controller/uploadClusterCA
   -u <username>:<password>
 
+curl -X POST /settings/security/allowNonLocalCACertUpload
```
timofey-barmin: Both curls are missing http://<ip-address-or-domain-name>:8091

modules/rest-api/pages/upload-retrieve-root-cert.adoc
```diff
@@ -52,10 +67,10 @@ curl -X GET
   -u <username>:<password>
```
timofey-barmin, line 60: --cacert only makes sense when https is used. Please remove --cacert from the examples where http is used.
I also suggest we remove --cacert from description below.

modules/rest-api/pages/delete-trusted-cas.adoc
tonyjhillman: Yes, it's just a link. I've clarified in the text.

modules/rest-api/pages/load-trusted-cas.adoc
tonyjhillman: I've added: 
Note the following:

* Files to be loaded must reside in the `inbox/CA` directory.
(Note that the loading process uses this directory in order to ensure security.)

* Each file may contains several certificates.

* Certificates must be in _pem_ format.

* Any file whose name starts with `.` is ignored.

modules/rest-api/pages/load-trusted-cas.adoc
tonyjhillman: I've removed.

modules/rest-api/pages/rest-certificate-management.adoc
tonyjhillman: I have added this.

modules/rest-api/pages/rest-certificate-management.adoc
tonyjhillman: I have re-added.

modules/rest-api/pages/upload-retrieve-root-cert.adoc
tonyjhillman: I have modified.

modules/rest-api/pages/upload-retrieve-root-cert.adoc
tonyjhillman: I have added this.

modules/rest-api/pages/upload-retrieve-root-cert.adoc
tonyjhillman: Thanks, I have fixed.

modules/rest-api/pages/upload-retrieve-root-cert.adoc
tonyjhillman: I may not be quite sure what to do here. So, I think my error was that 'http' rather than 'https' was specified in the curl syntax section (this conflicted with the description, which specified secure access only). So, I have changed 'http' to 'https' - does that in itself fix the problem? Or should I do more? Thanks.

modules/rest-api/pages/upload-retrieve-node-cert.adoc
tonyjhillman: I've changed to: For the loading of the node certificate to succeed, the private key and chain file must be readable by user `couchbase`.

modules/rest-api/pages/upload-retrieve-node-cert.adoc
tonyjhillman: Done.

modules/rest-api/pages/upload-retrieve-node-cert.adoc
tonyjhillman: Done.

modules/rest-api/pages/upload-retrieve-node-cert.adoc
tonyjhillman: I have modified accordingly.

modules/rest-api/pages/upload-retrieve-node-cert.adoc
tonyjhillman: I have modified.

modules/rest-api/pages/upload-retrieve-node-cert.adoc
tonyjhillman: I have changed to 'and, when the `plain` password-type has been used for an encrypted private key, the passphrase for the private key (securely displayed as asterisks):'

modules/rest-api/pages/upload-retrieve-root-cert.adoc
tonyjhillman: I have added.

modules/rest-api/pages/load-trusted-cas.adoc
timofey-barmin: Thanks, I guess contains -> contain

modules/rest-api/pages/upload-retrieve-node-cert.adoc
timofey-barmin: Thanks, also please fix data/script -> data/scripts

modules/rest-api/pages/upload-retrieve-root-cert.adoc
timofey-barmin: I think we have bigger problem here.
First of all, it's not clear to me what GET /pools/default has to do with "Upload and Retrieve the Root Certificate"
Also since all other examples use http it's not clear to me why this particular one suddenly uses https.
Also description of --cacert arg belongs to curl manual. It has nothing to do with couchbase.

I suggest we remove all mentions of GET /pools/default from this page. This will also remove the example in question and the description of --cacert.
If you have a reason to keep GET /pools/default, than I suggest to change https -> http (to be consistent with other examples), and remove --cacert from example and from description below.

modules/rest-api/pages/load-trusted-cas.adoc
tonyjhillman: Thanks, done.

modules/rest-api/pages/upload-retrieve-root-cert.adoc
tonyjhillman: I have removed all references to GET /pools/default from this page.

modules/rest-api/pages/upload-retrieve-node-cert.adoc
tonyjhillman: I've changed data/script to data/scripts.

timofey-barmin:
LGTM

timofey-barmin (approved):

PR #2308, opened 2021-12-01
DOC-9461: Modified management procedure for certs

tonyjhillman:


modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -305,35 +309,56 @@ sudo cp ./public/chain.pem /opt/couchbase/var/lib/couchbase/inbox/chain.pem
 sudo cp ./private/pkey.key /opt/couchbase/var/lib/couchbase/inbox/pkey.key
 ----
 
+. Change ownership of the private key to `couchbase`.
```
timofey-barmin: Strictly speaking we don't need the private key to be owned by user couchbase, we only need it to be 'read' by used couchbase.
Also the same is true for node chain file and all files in inbox/CA.
We can say something like the following here:
"Make sure all certificate and private key files can be read by user 'couchbase'. It can be achieved by, for example, changing ownership of the files to 'couchbase' and setting 0600 permissions"

modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -566,21 +591,40 @@ cp private/couchbase.node.svc.key pkey.key
...
 sudo mkdir /opt/couchbase/var/lib/couchbase/inbox/  # if needed
 
+
 sudo cp ./chain.pem /opt/couchbase/var/lib/couchbase/inbox/chain.pem
 sudo cp ./pkey.key /opt/couchbase/var/lib/couchbase/inbox/pkey.key
 ----
 
-. Upload the root certificate, thereby activating it for the entire cluster.
-Then, reload the node certificate from disk, for the current node.
+. Change the ownership of the private key to `couchbase`.
```
timofey-barmin: See my comment above

modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -640,18 +672,25 @@ Consequently:
 This is because a node can be joined to a cluster only prior to the node's provisioning; while certificate-deployment can only occur _subsequent to_ the node's provisioning.
```
timofey-barmin: >  No node can be _joined_ to a cluster.
> This is because a node can be joined to a cluster only prior to the node's provisioning; while certificate-deployment can only occur _subsequent to_ the node's provisioning.

I can't understand this sentence. "No node can be joined to a cluster" sounds like it's impossible to add a node to a cluster. I suggest we modify it.

modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -640,18 +672,25 @@ Consequently:
...
 
 * A node can be _added_ to a cluster only after the appropriate certificates have been deployed on it, as a provisioned, one-node cluster.
```
timofey-barmin: >  as a provisioned, one-node cluster.

This doesn't seem to be accurate and a bit confusing (confused at least one engineer).
There is no need to initialize a one-node cluster before join.

Here is what needs to be done exactly on the node before adding it to a cluster (assuming the cluster uses "custom", not self-generated certificates):
1) Load CA of the "cluster node" to the trust store of the node-to-be-added (load cluster CA basically)
2) Load CA of this node to the trust store of the node-to-be-added (most likely can be skipped because this CA is usually the same as in 1)
3) Make sure the cluster trusts this node's CA (it should already be true if CA from 1 and CA from 2 are the same);
4) Load node certificate

modules/manage/pages/manage-security/configure-server-certificates.adoc
tonyjhillman: I have changed accordingly.

modules/manage/pages/manage-security/configure-server-certificates.adoc
tonyjhillman: Regarding step 1 - I'm not sure I understand 'CA of the "cluster node"'. Does this mean 'a CA from the cluster's trust store that is used to sign the node certificate of at least one node in the cluster'? I ask because the cluster, as I understand, may be using more than one CA to do its node-signing..

tonyjhillman: Or, could it mean 'any CA in the cluster's trust store'. In which case, I wonder, would an alternative procedure be to upload the node's CA to the cluster's trust store, and then just add the node 'as is'?

modules/manage/pages/manage-security/configure-server-certificates.adoc
tonyjhillman: I've changed as advised above.

modules/manage/pages/manage-security/configure-server-certificates.adoc
timofey-barmin: When node-to-be-added is joining the cluster, it joins via some cluster node, right? This is what we call "cluster node" in the context of node addition. So basically there are two actors in node addition: "cluster node" and "node-to-be-added". During node addition both nodes must be able to connect to each other. Which means both ends should trust each other's CA. I think in most cases both ends will use the same CA's so it will work automatically, but it's possible that they use different CA's. Admin will needs to add "cluster node"'s CA to node-to-be-added then, and vise versa: add node-to-be-added's CA to cluster before joining.

Regarding existing doc: I suggest we remove "as a provisioned, one-node cluster" because there is no need to initialize anything before adding the node to the cluster. We only need to load certs.
So the doc can say: "A node can be _added_ to a cluster only after the appropriate certificates have been deployed on it."

If you want to elaborate you might add something like "if node-to-be-added's certificate and cluster node's certificate are issued by different CA's, make sure to load cluster node's CA to node-to-be-added and vise versa before adding the node"

tonyjhillman: Thanks for clarifying. I have written the following:

When a new node is added to a cluster, the operation is performed with reference to a particular node that is already a member of the cluster.
Since this _cluster node_ and this _new node_ must be able to connect with one another, each must trust the CA of the other.
Therefore:

* If the two nodes already have the same CA, connection and node-addition can occur.

* If the two nodes do _not_ have the same CA, the CA of each must be loaded onto the other; after which, connection and node-addition can occur.

timofey-barmin: Thanks. I think it makes sense. It seems like we are missing two things here:
1) This needs to be done only when cluster uses custom certificates (not self-generated).
2) Admin should add chain and pkey ("node certificates") to the node-to-be-added.


Probably we can have something like this in the very beginning of the "== Adding New Nodes":

> If the cluster uses self-generated certs nothing needs to be done before adding a node to the cluster.
> Otherwise node to be added has to be provisioned with node certificates before joining the cluster. 
> 
> When a new node is added to a cluster ...
> 

Note that by "provisioned with node certificates" I mean that certificates should be uploaded to inbox and corresponding API is called.
I don't mean node initialization, or one node cluster initialization.

tonyjhillman: I have now written this section as the following:

If a cluster uses _system-generated_ certificates, which Couchbase Server provides by default, no certificate-related work needs to be performed in order to add a new node to the cluster.
However, once a cluster is using _uploaded_ certificates, a node that is to be added must itself be provisioned with conformant certificates, before addition can be successfully performed: this means that the appropriate _chain_ file (containing the node certificate) and _private key_ must have been placed in the node's `inbox`, and the appropriate REST API then called.

Note also that when a cluster is using uploaded certificates, and a new node is added to a cluster, the operation is performed with reference to a particular node that is already a member of the cluster.
Since this _cluster node_ and this _new node_ must be able to connect with one another, each must trust the CA of the other.
Therefore:

* If the two nodes already have the same CA, connection and node-addition can occur.

* If the two nodes do _not_ have the same CA, the CA of each must be loaded onto the other; after which, connection and node-addition can occur.

timofey-barmin (approved):

PR #2316, opened 2021-12-06
DOC-9405a: Release notes for 6.6.4

tonyjhillman:


modules/release-notes/pages/relnotes.adoc
```diff
@@ -5,6 +5,83 @@
...
+|===
+
+==== Eventing Service
+
+[#table_fixedissues_v664-eventing-service,cols="25,66"]
+|===
+| Issue | Description
+
+| https://issues.couchbase.com/browse/MB-48289[MB-48289^]
+| *Summary*: Investigate eventing-producer RSS bloat when eventing-consumers are re-spawned continuously
```
jon-strabala, line 56: Perhaps drop the word "Investigate" or change it to "Fix" the current text reads like we did anything.

tonyjhillman: I will fix in the 6.6.5 release notes, which this content now becomes.

PR #2275, opened 2021-11-10
DOC-9205 - KV Operations dev-reorg

maria-robobug:
https://issues.couchbase.com/browse/DOC-9205

Will squash commits when it eventually gets merged.

**Ready:**
- [X] [Overview landing page](https://maria-robobug.github.io/DOC-9205/server/7.0-reorg/guides/intro.html)
- [x] [Key Value Operations page](https://maria-robobug.github.io/DOC-9205/server/7.0-reorg/guides/kv-operations.html)
- [x] [Create Data page](https://maria-robobug.github.io/DOC-9205/server/7.0-reorg/guides/creating-data.html)
- [x] [Read Data page](https://maria-robobug.github.io/DOC-9205/server/7.0-reorg/guides/reading-data.html)
- [x] [Update Data page](https://maria-robobug.github.io/DOC-9205/server/7.0-reorg/guides/updating-data.html)
- [x] [Delete Data page](https://maria-robobug.github.io/DOC-9205/server/7.0-reorg/guides/deleting-data.html)
- [x] [Bulk Operations page](https://maria-robobug.github.io/DOC-9205/server/7.0-reorg/guides/bulk-operations.html)

PR #2317, opened 2021-12-06
Doc 9216 import data

osfameron:
PR for import developer guide

Docs issue: [DOC-9216](https://issues.couchbase.com/browse/DOC-9216)

### DIFF
--- a/modules/guides/pages/import.adoc
+++ b/modules/guides/pages/import.adoc
@@ -28,8 +28,9 @@ Data load essentially consists of the following steps:
 . Connect to the appropriate bucket, scope, and collection.
 
 . Decide on the key for this document (could be an ID, a sequence number, or some combination of fields).
+
 . Do any additional processing if required.
-  
+
 . Insert the document.
 
 // tag::clients_section[]
@@ -128,9 +129,9 @@ include::nodejs-sdk:howtos:example$import.jsonl[]
 --
 ====
 
-== Use `cbimport`
+== Using `cbimport`
 
-Using `cbimport` is straight-forward.
+Using xref:tools:cbimport.adoc[cbimport] is straight-forward.
 Ensure you have the path to the command line clients in Couchbase Server in your path.
 
 You can import all of the data formats described above:
@@ -207,22 +208,24 @@ cbimport json \
 ====
 
 
-== Import using an SDK
+== Importing Using an SDK
 
 While `cbimport` accomplishes all the necessary steps in a single command, as above,
 using an SDK gives you more flexibility and control.
 However all the same considerations apply, so let us look at those in turn.
 
-=== Parse the import into an array or stream of records
+=== Parsing the Import into an Array or Stream of Records
 
-The details of how to parse the import data vary depending on the most appropriate library for your SDK:
+The details of how to parse the import data vary depending on the chosen input format, and the most appropriate library for your SDK.
+
+==== Parsing CSV and TSV data
 
 [{tabs}]
 ====
 pass:[.NET]::
 +
 --
-For *CSV* and *TSV*, use the https://joshclose.github.io/CsvHelper/[CsvHelper^] library: 
+Use the https://joshclose.github.io/CsvHelper/[CsvHelper^] library: 
 
 [source,csharp]
 ----
@@ -239,29 +242,13 @@ include::dotnet-sdk:howtos:example$Import.csx[tags=importCSV,indent=0]
 include::dotnet-sdk:howtos:example$Import.csx[tags=importTSV,indent=0]
 ----
 
-For *JSON* or *JSONL* format, use https://www.newtonsoft.com/json[Newtonsoft^]: 
-
-[source,csharp]
-----
-include::dotnet-sdk:howtos:example$Import.csx[tags=json-jsonl-import,indent=0]
-----
-[source,csharp]
-----
-include::dotnet-sdk:howtos:example$Import.csx[tags=importJSON,indent=0]
-----
-
-[source,csharp]
-----
-include::dotnet-sdk:howtos:example$Import.csx[tags=importJSONL,indent=0]
-----
-
 NOTE: {github}
 --
 
 Java::
 +
 --
-For *CSV* and *TSV*, use the http://opencsv.sourceforge.net/[opencsv^] library: 
+Use the http://opencsv.sourceforge.net/[opencsv^] library: 
 
 [source,java]
 ----
@@ -289,6 +276,88 @@ include::java-sdk:howtos:example$Import.java[tags=importCSV_batch;!omit,indent=0
 ----
 include::java-sdk:howtos:example$Import.java[tags=importTSV_batch;!omit,indent=0]
 ----
+
+NOTE: {github}
+--
+
+Node.js::
++
+--
+Use the https://csv.js.org/parse/[csv-parse^] library:
+
+[source,nodejs]
+----
+include::nodejs-sdk:howtos:example$import.js[tags=csv-tsv-import,indent=0]
+----
+
+[source,nodejs]
+----
+include::nodejs-sdk:howtos:example$import.js[tags=csvStream,indent=0]
+----
+
+[source,nodejs]
+----
+include::nodejs-sdk:howtos:example$import.js[tags=tsvStream,indent=0]
+----
+
+NOTE: {github}
+--
+
+Python::
++
+--
+Use the https://docs.python.org/3/library/csv.html[csv^] library:
+
+[source,python]
+----
+include::python-sdk:howtos:example$import.py[tags=csv-tsv-import,indent=0]
+----
+
+[source,python]
+----
+include::python-sdk:howtos:example$import.py[tags=csvImport,indent=0]
+----
+
+[source,python]
+----
+include::python-sdk:howtos:example$import.py[tags=tsvImport,indent=0]
+----
+
+
+NOTE: {github}
+--
+====
+
+==== Parsing JSON and JSONL data
+
+[{tabs}]
+====
+pass:[.NET]::
++
+--
+
+Use https://www.newtonsoft.com/json[Newtonsoft^]: 
+
+[source,csharp]
+----
+include::dotnet-sdk:howtos:example$Import.csx[tags=json-jsonl-import,indent=0]
+----
+[source,csharp]
+----
+include::dotnet-sdk:howtos:example$Import.csx[tags=importJSON,indent=0]
+----
+
+[source,csharp]
+----
+include::dotnet-sdk:howtos:example$Import.csx[tags=importJSONL,indent=0]
+----
+
+NOTE: {github}
+--
+
+Java::
++
+--
 
 For *JSON*, simply read the file as a string, then use our built-in xref:java-sdk:howtos:json.adoc[JSON] handling to parse the result into an array of JSON objects.
 
@@ -328,24 +397,7 @@ NOTE: {github}
 Node.js::
 +
 --
-For *CSV* and *TSV*, use the https://csv.js.org/parse/[csv-parse^] library:
-
-[source,nodejs]
-----
-include::nodejs-sdk:howtos:example$import.js[tags=csv-tsv-import,indent=0]
-----
-
-[source,nodejs]
-----
-include::nodejs-sdk:howtos:example$import.js[tags=csvStream,indent=0]
-----
-
-[source,nodejs]
-----
-include::nodejs-sdk:howtos:example$import.js[tags=tsvStream,indent=0]
-----
-
-For *JSON* or *JSONL* format, use the https://github.com/uhop/stream-json[stream-json^] library.
+Use the https://github.com/uhop/stream-json[stream-json^] library.
 
 NOTE: stream-json formats its output with a `{ key: ..., value: ...}` wrapper, so we need to map the stream into the expected format.
 
@@ -370,24 +422,7 @@ NOTE: {github}
 Python::
 +
 --
-For *CSV* and *TSV*, use the https://docs.python.org/3/library/csv.html[csv^] library:
-
-[source,python]
-----
-include::python-sdk:howtos:example$import.py[tags=csv-tsv-import,indent=0]
-----
-
-[source,python]
-----
-include::python-sdk:howtos:example$import.py[tags=csvImport,indent=0]
-----
-
-[source,python]
-----
-include::python-sdk:howtos:example$import.py[tags=tsvImport,indent=0]
-----
-
-For *JSON* and *JSONL*, use the https://docs.python.org/3/library/json.html[json^] library:
+Use the https://docs.python.org/3/library/json.html[json^] library:
 
 [source,python]
 ----
@@ -408,7 +443,7 @@ NOTE: {github}
 --
 ====
 
-=== Connect to the Couchbase Server
+=== Connecting to the Couchbase Server
 
 First, you need the connection details for the Couchbase server.
 
@@ -472,7 +507,8 @@ NOTE: {github}
 --
 ====
 
-=== Inserting the documents
+
+=== Inserting the Documents
 
 Having processed each imported document, you can insert it into the keyspace.
 
@@ -494,7 +530,7 @@ This means that in the case of any error, it is easy to make any required tweaks
 pass:[.NET]::
 +
 --
-Hook the prepared data into an `upsertDocument` routine:
+Hook the prepared data into an `upsert` routine:
 
 NOTE: as CsvHelper and Newtonsoft generate different outputs, we've provided some overloaded options that work for either.
 
@@ -599,9 +635,14 @@ NOTE: {github}
 
 * The xref:guides:kv-operations.adoc[] guide shows how to read and update the data you have imported. 
 
+Key-Value Operations with SDKs:
 
-////
-See also:
-
-NOTE: Can you add links to the kv-operations page for every SDK? Copy from @maria-robobug 's Creating Data how-to guide.
-////
+* xref:c-sdk:howtos:kv-operations.adoc[C]
+| xref:dotnet-sdk:howtos:kv-operations.adoc[.NET]
+| xref:go-sdk:howtos:kv-operations.adoc[Go]
+| xref:java-sdk:howtos:kv-operations.adoc[Java]
+| xref:nodejs-sdk:howtos:kv-operations.adoc[Node.js]
+| xref:php-sdk:howtos:kv-operations.adoc[PHP]
+| xref:python-sdk:howtos:kv-operations.adoc[Python]
+| xref:ruby-sdk:howtos:kv-operations.adoc[Ruby]
+| xref:scala-sdk:howtos:kv-operations.adoc[Scala]
--- a/modules/guides/pages/import.adoc
+++ b/modules/guides/pages/import.adoc
@@ -1,5 +1,5 @@
-= Import Data
-// Import and Stream Data
+= Importing Data
+// Importing and Streaming Data
 :page-topic-type: guide
 :imagesdir: ../assets/images
 :tabs:
@@ -17,17 +17,18 @@ Importing data can be done from the Couchbase Server UI, via the
 xref:tools:cbimport.adoc[cbimport] command-line tool shipped with Couchbase Server,
 or using the SDK to script the process.
 
-Data load essentially consists of:
+Data load essentially consists of the following steps:
 
 . Prepare data in some well known format such as Comma Separated Values or JSON documents. 
-  * Parse this data, and iterate over each document.
+
+. Parse this data, and iterate over each document.
 
 . Connect to your Couchbase instance.
 
 . Connect to the appropriate bucket, scope, and collection.
 
 . Decide on the key for this document (could be an ID, a sequence number, or some combination of fields).
-  * Do any additional processing if required.
+. Do any additional processing if required.
   
 . Insert the document.
 
@@ -119,7 +120,8 @@ JSONL::
 --
 JSON Lines (.json) also known as NDJSON is a common format for streaming JSON, with one JSON object per line of text.
 
-[source,jsonl]
+// "jsonl" not currently supported by highlighter
+[source,json]
 ----
 include::nodejs-sdk:howtos:example$import.jsonl[]
 ----
@@ -146,10 +148,10 @@ cbimport csv \
   --bucket travel-sample --scope-collection-exp inventory.airline \ <.>
   --generate-key %type%_%id% <.>
 ----
-<.> import a local CSV file
-<.> replace these values with your connection details
-<.> replace the bucket, scope, and collection as required
-<.> you need to choose the ID to import as. This generator will generate IDs such as `airline_1234`
+<.> Import a local CSV file.
+<.> Replace these values with your connection details.
+<.> Replace the bucket, scope, and collection as required.
+<.> Specify an ID for the imported documents. This generator will generate IDs such as `airline_1234`.
 --
 
 TSV::
@@ -163,10 +165,10 @@ cbimport csv \
   --bucket travel-sample --scope-collection-exp inventory.airline \ <.>
   --generate-key %type%_%id% <.>
 ----
-<.> import a local TSV file
-<.> replace these values with your connection details
-<.> replace the bucket, scope, and collection as required
-<.> you need to choose the ID to import as. This generator will generate IDs such as `airline_1234`
+<.> Import a local TSV file.
+<.> Replace these values with your connection details.
+<.> Replace the bucket, scope, and collection as required.
+<.> Specify an ID for the imported documents. This generator will generate IDs such as `airline_1234`.
 --
 
 JSON::
@@ -180,10 +182,10 @@ cbimport json \
   --bucket travel-sample --scope-collection-exp inventory.airline \ <.>
   --generate-key %type%_%id% <.>
 ----
-<.> import a local JSON file
-<.> replace these values with your connection details
-<.> replace the bucket, scope, and collection as required
-<.> you need to choose the ID to import as. This generator will generate IDs such as `airline_1234`
+<.> Import a local JSON file.
+<.> Replace these values with your connection details.
+<.> Replace the bucket, scope, and collection as required.
+<.> Specify an ID for the imported documents. This generator will generate IDs such as `airline_1234`.
 --
 
 JSONL::
@@ -197,10 +199,10 @@ cbimport json \
   --bucket travel-sample --scope-collection-exp inventory.airline \ <.>
   --generate-key %type%_%id% <.>
 ----
-<.> import a local JSONL file
-<.> replace these values with your connection details
-<.> replace the bucket, scope, and collection as required
-<.> you need to choose the ID to import as. This generator will generate IDs such as `airline_1234`
+<.> Import a local JSONL file.
+<.> Replace these values with your connection details.
+<.> Replace the bucket, scope, and collection as required.
+<.> Specify an ID for the imported documents. This generator will generate IDs such as `airline_1234`.
 --
 ====
 
@@ -220,7 +222,7 @@ The details of how to parse the import data vary depending on the most appropria
 pass:[.NET]::
 +
 --
-For *CSV* and *TSV*, use the xref:https://joshclose.github.io/CsvHelper/[CsvHelper] library: 
+For *CSV* and *TSV*, use the https://joshclose.github.io/CsvHelper/[CsvHelper^] library: 
 
 [source,csharp]
 ----
@@ -237,7 +239,7 @@ include::dotnet-sdk:howtos:example$Import.csx[tags=importCSV,indent=0]
 include::dotnet-sdk:howtos:example$Import.csx[tags=importTSV,indent=0]
 ----
 
-For *JSON* or *JSONL* format, use xref:https://www.newtonsoft.com/json[Newtonsoft]: 
+For *JSON* or *JSONL* format, use https://www.newtonsoft.com/json[Newtonsoft^]: 
 
 [source,csharp]
 ----
@@ -259,7 +261,7 @@ NOTE: {github}
 Java::
 +
 --
-For *CSV* and *TSV*, use the xref:http://opencsv.sourceforge.net/[opencsv] library: 
+For *CSV* and *TSV*, use the http://opencsv.sourceforge.net/[opencsv^] library: 
 
 [source,java]
 ----
@@ -275,7 +277,8 @@ include::java-sdk:howtos:example$Import.java[tags=importCSV,indent=0]
 include::java-sdk:howtos:example$Import.java[tags=importTSV;!omit,indent=0]
 ----
 
-If you are using the *Reactor API* then, as OpenCSV doesn't have a built-in converter, use the Flux::generate method to convert the CSV or TSV file into a stream:
+If you are using the *Reactor API* then, as OpenCSV doesn't have a built-in converter, 
+use the https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#generate-java.util.concurrent.Callable-java.util.function.BiFunction-java.util.function.Consumer-[`Flux::generate`] method to convert the CSV or TSV file into a stream:
 
 [source,java]
 ----
@@ -298,7 +301,7 @@ include::java-sdk:howtos:example$Import.java[tags=json-jsonl-import,indent=0]
 include::java-sdk:howtos:example$Import.java[tags=importJSON,indent=0]
 ----
 
-If you are using the *Reactor API* then, once you've read the JSON array, use the Flux::fromIterable method to convert it into streams:
+If you are using the *Reactor API* then, once you've read the JSON array, use the https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#fromIterable-java.lang.Iterable-[`Flux::fromIterable`] method to convert it into streams:
 
 [source,java]
 ----
@@ -312,7 +315,7 @@ For *JSONL*, do the same, but reading the file line-by-line:
 include::java-sdk:howtos:example$Import.java[tags=importJSONL;!omit,indent=0]
 ----
 
-If you are using the *Reactor API* then open the JSONL file as a stream using the Flux::using method:
+If you are using the *Reactor API* then open the JSONL file as a stream using the https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#using-java.util.concurrent.Callable-java.util.function.Function-java.util.function.Consumer-[`Flux::using`] method:
 
 [source,java]
 ----
@@ -325,7 +328,7 @@ NOTE: {github}
 Node.js::
 +
 --
-For *CSV* and *TSV*, use the xref:https://csv.js.org/parse/[csv-parse] library:
+For *CSV* and *TSV*, use the https://csv.js.org/parse/[csv-parse^] library:
 
 [source,nodejs]
 ----
@@ -342,7 +345,7 @@ include::nodejs-sdk:howtos:example$import.js[tags=csvStream,indent=0]
 include::nodejs-sdk:howtos:example$import.js[tags=tsvStream,indent=0]
 ----
 
-For *JSON* or *JSONL* format, use the xref:https://github.com/uhop/stream-json[stream-json] library.
+For *JSON* or *JSONL* format, use the https://github.com/uhop/stream-json[stream-json^] library.
 
 NOTE: stream-json formats its output with a `{ key: ..., value: ...}` wrapper, so we need to map the stream into the expected format.
 
@@ -367,7 +370,7 @@ NOTE: {github}
 Python::
 +
 --
-For *CSV* and *TSV*, use the xref:https://docs.python.org/3/library/csv.html[csv] library:
+For *CSV* and *TSV*, use the https://docs.python.org/3/library/csv.html[csv^] library:
 
 [source,python]
 ----
@@ -384,7 +387,7 @@ include::python-sdk:howtos:example$import.py[tags=csvImport,indent=0]
 include::python-sdk:howtos:example$import.py[tags=tsvImport,indent=0]
 ----
 
-For *JSON* and *JSONL*, use the xref:https://docs.python.org/3/library/json.html[json] library:
+For *JSON* and *JSONL*, use the https://docs.python.org/3/library/json.html[json^] library:
 
 [source,python]
 ----
@@ -409,7 +412,7 @@ NOTE: {github}
 
 First, you need the connection details for the Couchbase server.
 
-Now decide which bucket and xref:learn:data:scopes-and-collections.adoc[scope and collection] you want to import to (and create them if they don't already exist.)
+Now decide which bucket and xref:learn:data/scopes-and-collections.adoc[scope and collection] you want to import to (and create them if they don't already exist.)
 
 [{tabs}]
 ====
@@ -433,7 +436,7 @@ Java::
 include::java-sdk:howtos:example$Import.java[tags=connect,indent=0]
 ----
 
-If you are using the xref:java:howtos:concurrent-async-apis.adoc#reactive-programming-with-reactor[Reactive API], then use the reactive collection instead:
+If you are using the xref:java-sdk:howtos:concurrent-async-apis.adoc#reactive-programming-with-reactor[Reactive API], then use the reactive collection instead:
 
 [source,java]
 ----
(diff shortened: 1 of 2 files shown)


maria-robobug:
Shouldn't this PR be for the `dev-reorg` branch instead of `release/7.0`?

osfameron:
> Shouldn't this PR be for the `dev-reorg` branch instead of `release/7.0`?

Yes indeed! Still draft of course, lemme see if I can repoint it, otherwise will recreate...

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
+[abstract]
+{description}
+
+== Introduction
+
+Importing data can be done from the Couchbase Server UI,
+via the `cbimport` command-line tool shipped with Couchbase Server,
+or using the SDK to script the process.
+
+XXX server/current/tools/cbimport.html
```
simon-dew: The Antora path is xref:tools:cbimport.adoc[cbimport].

Is it possible to include the `cbimport` examples in each of the tab sets below, or does it work too differently from the way the SDKs do? (Either is fine)

Refer to #2275 to see how the `cbc` examples are included alongside the SDK examples.

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
+----
+include::python-sdk:howtos:example$import.py[tag=connect,indent=0]
+----
+--
+====
+
+
+== Parse the import into an array or stream of records
+
+We now need to parse the input data, either by reading the whole document into memory (if that's possible) or streaming each record in turn.
```
simon-dew: This wording reads more like a tutorial. For a how-to guide it should be more like a procedure, with a title identifying the goal of the procedure, and a list of steps, each of which contains a "do" (imperative statement). And definitely second-person.

So something like this — although it's a bit hard to tell if this is the correct suggestion without having seen the code.

```suggestion
To parse the input data, do one of the following:

* Read the whole document into memory, if possible.

* Stream each record in turn.
```

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
+----
+--
+
+////
+
+Node.js::
++
+--
+
+We can easily stream a *CSV* file with an external library.
```
simon-dew:
```suggestion
To stream a *CSV* file, use the XXX external library.
```

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
+----
+include::nodejs-sdk:howtos:example$import.js[tags=csv-tsv-import,indent=0]
+----
+
+[source,nodejs]
+----
+include::nodejs-sdk:howtos:example$import.js[tags=csvStream,indent=0]
+----
+
+*TSV* files can be processed with the same library:
```
simon-dew:
```suggestion
Similarly, to process a *TSV* file, use the XXX external library.
```

I'll allow the "similarly", but it's not really necessary.

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
+----
+include::nodejs-sdk:howtos:example$import.js[tags=csv-tsv-import,indent=0]
+----
+
+[source,nodejs]
+----
+include::nodejs-sdk:howtos:example$import.js[tags=tsvStream,indent=0]
+----
+
+*JSON* requires a different library, which streams its output in a `{ key: ..., value: ...}` wrapper, so we need to map the stream into the expected format:
```
simon-dew:
```suggestion
To stream a *JSON* file, use the YYY library, and map the stream into the expected `{ key: ..., value: ...}` format.
```

Again, without seeing the code, my suggestion may not be exactly right, but hopefully this conveys the right tone and voice.

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
+
+[source,nodejs]
+----
+include::nodejs-sdk:howtos:example$import.js[tags=json-jsonl-import;json-import,indent=0]
+----
+----
+include::nodejs-sdk:howtos:example$import.js[tags=jsonStream,indent=0]
+----
+
+The *JSON Lines* format requires the same treatment:
```
simon-dew:
```suggestion
Similary, to stream the *JSON Lines* format, use the YYY library, and map the stream into the expected `{ key: ..., value: ...}` format.
```

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
+include::nodejs-sdk:howtos:example$import.js[tags=json-jsonl-import;jsonl-import,indent=0]
+----
+----
+include::nodejs-sdk:howtos:example$import.js[tags=jsonlStream,indent=0]
+----
+--
+Python::
++
+--
+We can easily parse a *CSV* file with an external library.
```
simon-dew:
```suggestion
To parse a *CSV* file, use the XXX.py external library.
```

Tighten up the terminology: is it parse, stream, or process? (I don't know the right answer here, but use the correct term for the SDK and the process you are describing, and be consistent across steps and procedures.)

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
+
+[source,python]
+----
+include::python-sdk:howtos:example$import.py[tags=csv-tsv-import;csvImport,indent=0]
+----
+
+*TSV* files can be processed with the same library:
```
simon-dew:
```suggestion
Similary, to process a *TSV* file, use the XXX.py external library.
```

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
+
+[source,python]
+----
+include::python-sdk:howtos:example$import.py[tags=csv-tsv-import;tsvImport,indent=0]
+----
+
+
+*JSON* requires a different library:
```
simon-dew:
```suggestion
To process a *JSON* file, use the YYY.py external library.
```

Also tighten up terminology around "file" or "stream".

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
+
+[source,python]
+----
+include::python-sdk:howtos:example$import.py[tags=json-jsonl-import;jsonImport,indent=0]
+----
+
+The *JSON Lines* format requires the same treatment, with a little processing to get the line-by-line behaviour:
```
simon-dew:
```suggestion
Similary, to stream the *JSON Lines* format, use the YYY.py library, and process the stream to get the line-by-line behavior.
```

🇺🇸 spelling

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
+
+
+[{tabs}]
+====
+CSV::
++
+--
+Comma Separated Values (.csv) are easily exported from many spreadsheet and database applications.
+
+Typically, the first row is a header row containing the names of the columns within the document.
```
simon-dew: Jus to tighten up the wording for a how-to guide: rather than describing the situations they might encounter in a discursive way, _tell them what to do_ (in so many words). So this might be something like

```suggestion
If required, include a header row containing the names of the columns within the document.
```

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
+
+[source,csv]
+----
+include::nodejs-sdk:howtos:example$import.csv[]
+----
+
+{github}
```
simon-dew: Don't think you need this here? It's a complete example rather than an extract?

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
++
+--
+Tab Separated Values (.csv) are a common variant of CSV files.
+
+[source,tsv]
+----
+include::nodejs-sdk:howtos:example$import.tsv[]
+----
+
+{github}
```
simon-dew: Don't think you need this

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
++
+--
+JSON (.json) files are especially well suited to import into Couchbase, as it is the default native datatype. As we want to import a number of objects, and a .json file contains only one single value, we will typically provide a .json with an *array* of the values to import.
+
+[source,json]
+----
+include::nodejs-sdk:howtos:example$import.json[]
+----
+
+{github}
```
simon-dew: Don't think you need this

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
++
+--
+JSON Lines (.json) also known as NDJSON is a common format for streaming JSON, with one JSON object per line of text.
+
+[source,jsonl]
+----
+include::nodejs-sdk:howtos:example$import.jsonl[]
+----
+
+{github}
```
simon-dew: Don't think you need this here

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
+== Connect to the Couchbase Server
+
+First you'll need the connection details for the Couchbase server.
+You should also decide which bucket, scope, and collection you want to import to (and create them if they don't already exist.)
+
+XXX learn/data/scopes-and-collections.html
+
+XXX cloud get-started cluster-and-date #import page
+
+[{tabs}]
```
simon-dew, line 139: I think you do need the `{github}` link for each example in this tab set, as you are including extracts.

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
+====
+////
+pass:[.NET]::
++
+--
+[source,csharp]
+----
+include::dotnet-sdk:howtos:example$Import.cs[tag=connect,indent=0]
+----
+--
```
simon-dew, line 156: Include a link to the relevant .NET reference page. See the SDK examples [on this page](https://simon-dew.github.io/docs-site/DOC-9196/server/7.0-reorg/guides/create-index.html).

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
+Java::
++
+--
+[source,java]
+----
+include::java-sdk:howtos:example$Import.java[tag=connect,indent=0]
+----
+--
```
simon-dew, line 173: Include a link to the relevant Java reference page.

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
+////
+
+Node.js::
++
+--
+[source,nodejs]
+----
+include::nodejs-sdk:howtos:example$import.js[tag=connect,indent=0]
+----
+--
```
simon-dew, line 190: Include a link to the relevant Node.js reference page.

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
+Python::
++
+--
+[source,python]
+----
+include::python-sdk:howtos:example$import.py[tag=connect,indent=0]
+----
+--
```
simon-dew, line 207: Include a link to the relevant Python reference page.

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
+====
+
+
+== Parse the import into an array or stream of records
+
+We now need to parse the input data, either by reading the whole document into memory (if that's possible) or streaming each record in turn.
+
+[{tabs}]
```
simon-dew, line 223: In this tab set, include the `{github}` link for each example, and include a link to the relevant SDK reference page for each tab.

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
+But of course, Couchbase is a "Key-Value" store, and the document is the *value*... but what's the key?
+
+So the first step is to determine the key. This could be as simple as extracting the `id` field from the document, or using an incrementing sequence number.
+
+Then we may want to do additional processing, for example calculating fields, or adding metadata about the importer.
+
+Finally we do an insert (or upsert, to update the document with that key, if it exists.)
+
+
+[{tabs}]
```
simon-dew, line 528: In this tab set, include the `{github}` link for each example, and include a link to the relevant SDK reference page for each tab.

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
+====
+Node.js::
++
+--
+
+Node lets us iterate a stream with a simple `for` loop in the same way as an array:
```
simon-dew:
```suggestion
To iterate the stream, use a simple `for` loop in the same way as an array.
```

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
+
+[source,nodejs]
+----
+include::nodejs-sdk:howtos:example$import.js[tag=importStream,indent=0]
+----
+
+The processing and upsert are then done in this routine:
```
simon-dew: Rewrite this as an imperative instruction rather than a description. (Sorry I don't know the details, you'll have to supply them.)

```suggestion
Use the PPPP function to process and the UUUU function to upsert each record.
```

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
+
+[source,nodejs]
+----
+include::nodejs-sdk:howtos:example$import.js[tag=upsertDocument,indent=0]
+----
+--
+Python::
++
+--
+For convenience, let's define functions to select the key, and process the value:
```
simon-dew:
```suggestion
Define functions to select the key and process the value.
```

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
+[source,python]
+----
+include::python-sdk:howtos:example$import.py[tag=key,indent=0]
+----
+[source,python]
+----
+include::python-sdk:howtos:example$import.py[tag=process,indent=0]
+----
+
+Now the upsert function can simply call those:
```
simon-dew:
```suggestion
Use the upsert function to call the functions you defined.
```

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
+
+[source,python]
+----
+include::python-sdk:howtos:example$import.py[tag=upsertDocument,indent=0]
+----
+
+NOTE: The Python SDK offers a set of batch operations which are marked as _volatile_ as of SDK 3.2.3, which may be more efficient. Here's a brief example for CSV:
```
simon-dew, line 620: This is a note, so fine as description rather than instruction.

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
+== Inserting the documents
+
+So we have a document, and we know where to put it, let's go!
+But of course, Couchbase is a "Key-Value" store, and the document is the *value*... but what's the key?
+
+So the first step is to determine the key. This could be as simple as extracting the `id` field from the document, or using an incrementing sequence number.
+
+Then we may want to do additional processing, for example calculating fields, or adding metadata about the importer.
+
+Finally we do an insert (or upsert, to update the document with that key, if it exists.)
```
simon-dew: Again, rewrite to be more of a how-to guide than a tutorial. Should be less hand-holdy, and definitely second person ("you do this!" not "we can do this".)

```suggestion
Having processed each imported document, you can insert it into the keyspace.
Couchbase is a key-value store, and the document is the value, so before you can insert the document, you need to determine the key.

To insert an imported document into the keyspace:

. Specify the key. This could be as simple as extracting the `id` field from the document, or using an incrementing sequence number.

. Do any additional processing, for example calculating fields, or adding metadata about the importer.

. Finally, use an insert operation to the insert the document.
Alternatively, if you want to update an existing document with the same key, use an upsert operation to update the existing document.
```

simon-dew (changes requested):
Had a quick look at this to make sure it's going in the right direction. I haven't checked out the request and built the document so I have not commented on the technical accuracy. Most changes requested around style and tone to make it more of a how-to guide.

osfameron:
Thanks @simon-dew!
cbcimport is next up after the SDK things, I'll try to incorporate if it makes sense, otherwise will precede the SDK ones.
Other notes look good, will finish code samples then get to that polish / approach tweak.

modules/guides/pages/import.adoc
osfameron: Ah. CSV lets you do either, but the *code examples* use the header row, because I've made decision that's the best approach for us.

Just checking we can be *opinionated* here? And I don't have to pepper with `NOTE:`s pointing out that you could do it differently...

modules/guides/pages/import.adoc
osfameron: I'm unsure what this *does*... (I'd copy/pasted it above, it renders as a literal string.)

modules/guides/pages/import.adoc
simon-dew: You haven't defined the `{github}` attribute in this file anyway. See https://raw.githubusercontent.com/couchbase/docs-server/dev-reorg/modules/guides/pages/select.adoc#L6.

modules/guides/pages/import.adoc
simon-dew: You haven't defined the `{github}` attribute in this file

modules/guides/pages/import.adoc
simon-dew: You haven't defined the `{github}` attribute in this file

modules/guides/pages/import.adoc
simon-dew: You haven't defined the `{github}` attribute in this file

modules/guides/pages/import.adoc
simon-dew: You haven't defined the `{github}` attribute in this file

modules/guides/pages/import.adoc
simon-dew: Hmmm, it's an arguable question at this specific point in the document — are you describing the layout of a typical CSV file here, or instructing users how to lay out a CSV file for import?

Generally the example files should illustrate the instructions, rather than the other way around.

It's fine to be opinionated — in fact you have to be, because we have to decide what procedures we're going to cover, and what steps each procedure should include. The instructions should be clear and concise, so we can only cover the most likely use cases, referring to the reference material for everything else.

The point to bear in mind is that in a how-to guide, we don't know exactly what the user is trying to do. They have a problem they're trying to solve: import some data. So (IMHO) we can't say to them, "our file looks like this, we want to do this". We don't know what their file looks like. We have to give them instructions that will cover as much of their problem as is reasonably possible. The examples to illustrate the procedure and confirm that they're on the right lines. Then refer them to the reference pages for edge cases.

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,335 @@
...
+include::nodejs-sdk:howtos:example$import.tsv[]
+----
+
+{github}
+--
+
+JSON::
++
+--
+JSON (.json) files are especially well suited to import into Couchbase, as it is the default native datatype. As we want to import a number of objects, and a .json file contains only one single value, we will typically provide a .json with an *array* of the values to import.
```
simon-dew: This is a good example of the point I was trying to make above (L46). We don't know whether the user wants to import a number of objects, although it's a good guess that they do.

If you decide that this section is instructional (how to lay out a file for import), you could reword like this:

```suggestion
JSON (.json) files are especially well suited to import into Couchbase, as it is the default native datatype.
A JSON file can only contain one single value.
If you want to import a number of objects, create a JSON file with an *array* of the values to import.
```

simon-dew: Alternatively, if you decide that this section is descriptive (of the layout of a typical JSON file), you could reword like this:

```suggestion
JSON (.json) files are especially well suited to import into Couchbase, as it is the default native datatype.
A JSON file can contains only one single value.
If you need to import a number of objects,  the JSON file will typically contain an *array* of the values to import.
```

I'll leave you to decide which approach to take for this particular section, consistently of course 😀 , but for the most part I think you should default to the instructional approach for the how-to guides.

osfameron: Though giving an `if` suggests that users may later ask for an `else`... I can say 

> To let you import one or any number of objects as needed, create an array...

Also, it's possible (though unlikely) that they may want to store arrays, strings, etc.
Which is fine (but the "process by adding an import key" example later would then break)

Um, just thinking out loud, I think, will play around and submit again when ready for review!

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+[abstract]
+{description}
+
+== Introduction
+
+Importing data can be done from the Couchbase Server UI, via the  
+xref:tools:cbimport.adoc[cbimport] command-line tool shipped with Couchbase Server,
+or using the SDK to script the process.
+
+Data load essentially consists of:
```
simon-dew: When a list of bullets or numbered points is introduced by an incomplete lead-in clause like this, I think the syntax of the points should follow naturally from the lead-in clause. As it is, it reads a bit awkwardly: "Data load essentially consists of prepare data..."

Either update the numbered points to start with a present participle ("Preparing...", "Parsing...", "Connecting...", etc.), or change the lead-in paragraph to a complete sentence:

```suggestion
Data load essentially consists of the following steps:
```

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+
+. Prepare data in some well known format such as Comma Separated Values or JSON documents. 
+  * Parse this data, and iterate over each document.
```
simon-dew: Not sure why this is a sub-point? Seems equally important

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+
+. Connect to your Couchbase instance.
+
+. Connect to the appropriate bucket, scope, and collection.
+
+. Decide on the key for this document (could be an ID, a sequence number, or some combination of fields).
+  * Do any additional processing if required.
```
simon-dew: Again, not sure why this is a sub-point and not a top-level list item?

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+include::nodejs-sdk:howtos:example$import.json[]
+----
+--
+
+JSONL::
++
+--
+JSON Lines (.json) also known as NDJSON is a common format for streaming JSON, with one JSON object per line of text.
+
+[source,jsonl]
```
simon-dew: `jsonl` isn't currently supported by our syntax highlighter, so I would change this to json

```suggestion
[source,json]
```

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+cbimport csv \
+  --dataset -d file://./import.csv \ <.>
+  --cluster localhost --username Administrator --password password \ <.>
+  --bucket travel-sample --scope-collection-exp inventory.airline \ <.>
+  --generate-key %type%_%id% <.>
+----
+<.> import a local CSV file
+<.> replace these values with your connection details
+<.> replace the bucket, scope, and collection as required
+<.> you need to choose the ID to import as. This generator will generate IDs such as `airline_1234`
```
simon-dew: Use sentence case and full stop for callout list items. Otherwise it looks weird when a sentence break occurs in the middle of the list item, as in point 4 here.

```suggestion
<.> Import a local CSV file.
<.> Replace these values with your connection details.
<.> Replace the bucket, scope, and collection as required.
<.> Specify an ID for the imported documents. This generator will generate IDs such as `airline_1234`.
```

Also tightened up the wording slightly for point four.

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+=== Parse the import into an array or stream of records
+
+The details of how to parse the import data vary depending on the most appropriate library for your SDK:
+
+[{tabs}]
+====
+pass:[.NET]::
++
+--
+For *CSV* and *TSV*, use the xref:https://joshclose.github.io/CsvHelper/[CsvHelper] library:
```
simon-dew: This has to be a link, not a cross-reference, because it's an external URL. I tend to use `^` for external links so that they open in a new tab.

```suggestion
For *CSV* and *TSV*, use the https://joshclose.github.io/CsvHelper/[CsvHelper^] library: 
```

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+cbimport json \
+  --dataset -d file://./import.json --format list \ <.>
+  --cluster localhost --username Administrator --password password \ <.>
+  --bucket travel-sample --scope-collection-exp inventory.airline \ <.>
+  --generate-key %type%_%id% <.>
+----
+<.> import a local JSON file
+<.> replace these values with your connection details
+<.> replace the bucket, scope, and collection as required
+<.> you need to choose the ID to import as. This generator will generate IDs such as `airline_1234`
```
simon-dew: Use sentence case and full stop for callout list items. Otherwise it looks weird when a sentence break occurs in the middle of the list item, as in point 4 here.

```suggestion
<.> Import a local JSON file.
<.> Replace these values with your connection details.
<.> Replace the bucket, scope, and collection as required.
<.> Specify an ID for the imported documents. This generator will generate IDs such as `airline_1234`.
```

Also tightened up the wording slightly for point four.

simon-dew: To be clear, I think it's fine to have list items that are _not_ capitalized and do not end with a full stop, if they are all short phrases. But if any list item contains a sentence break, then I think _all_ the list items should be punctuated as complete sentences. The key thing is consistency.

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+----
+include::dotnet-sdk:howtos:example$Import.csx[tags=importCSV,indent=0]
+----
+
+[source,csharp]
+----
+include::dotnet-sdk:howtos:example$Import.csx[tags=importTSV,indent=0]
+----
+
+For *JSON* or *JSONL* format, use xref:https://www.newtonsoft.com/json[Newtonsoft]:
```
simon-dew: This has to be a link, not a cross-reference, because it's an external URL. I tend to use `^` for external links so that they open in a new tab.

```suggestion
For *JSON* or *JSONL* format, use https://www.newtonsoft.com/json[Newtonsoft^]: 
```

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+include::dotnet-sdk:howtos:example$Import.csx[tags=importJSONL,indent=0]
+----
+
+NOTE: {github}
+--
+
+Java::
++
+--
+For *CSV* and *TSV*, use the xref:http://opencsv.sourceforge.net/[opencsv] library:
```
simon-dew: Likewise, this has to be a link, not a cross-reference.

```suggestion
For *CSV* and *TSV*, use the http://opencsv.sourceforge.net/[opencsv^] library: 
```

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+----
+include::java-sdk:howtos:example$Import.java[tags=importCSV,indent=0]
+----
+
+[source,java]
+----
+include::java-sdk:howtos:example$Import.java[tags=importTSV;!omit,indent=0]
+----
+
+If you are using the *Reactor API* then, as OpenCSV doesn't have a built-in converter, use the Flux::generate method to convert the CSV or TSV file into a stream:
```
simon-dew: Add link to documentation for the `Flux::generate` method

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+====
+
+
+== Import using an SDK
+
+While `cbimport` accomplishes all the necessary steps in a single command, as above,
+using an SDK gives you more flexibility and control.
+However all the same considerations apply, so let us look at those in turn.
+
+=== Parse the import into an array or stream of records
```
simon-dew: The procedures and examples in this section are quite long. For better readability I would split this section into three:

* Parse a CSV or TSV file
* Parse a JSON file
* Parse a JSONL file

With tab sets for each of the four SDKs in each section.

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+[source,java]
+----
+include::java-sdk:howtos:example$Import.java[tags=json-jsonl-import,indent=0]
+----
+[source,java]
+----
+include::java-sdk:howtos:example$Import.java[tags=importJSON,indent=0]
+----
+
+If you are using the *Reactor API* then, once you've read the JSON array, use the Flux::fromIterable method to convert it into streams:
```
simon-dew: Add link to documentation for the `Flux::fromIterable` method

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+----
+
+For *JSONL*, do the same, but reading the file line-by-line:
+
+[source,java]
+----
+include::java-sdk:howtos:example$Import.java[tags=importJSONL;!omit,indent=0]
+----
+
+If you are using the *Reactor API* then open the JSONL file as a stream using the Flux::using method:
```
simon-dew: Add link to documentation for the `flux:using` method

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+include::java-sdk:howtos:example$Import.java[tags=importJSONL_batch;!omit,indent=0]
+----
+
+NOTE: {github}
+--
+
+Node.js::
++
+--
+For *CSV* and *TSV*, use the xref:https://csv.js.org/parse/[csv-parse] library:
```
simon-dew: This should be a link, not an xref

```suggestion
For *CSV* and *TSV*, use the https://csv.js.org/parse/[csv-parse^] library:
```

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+----
+include::nodejs-sdk:howtos:example$import.js[tags=csvStream,indent=0]
+----
+
+[source,nodejs]
+----
+include::nodejs-sdk:howtos:example$import.js[tags=tsvStream,indent=0]
+----
+
+For *JSON* or *JSONL* format, use the xref:https://github.com/uhop/stream-json[stream-json] library.
```
simon-dew: This should be a link, not an xref

```suggestion
For *JSON* or *JSONL* format, use the https://github.com/uhop/stream-json[stream-json] library.
```

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+include::nodejs-sdk:howtos:example$import.js[tags=jsonlStream,indent=0]
+----
+
+NOTE: {github}
+--
+
+Python::
++
+--
+For *CSV* and *TSV*, use the xref:https://docs.python.org/3/library/csv.html[csv] library:
```
simon-dew: Link not xref

```suggestion
For *CSV* and *TSV*, use the https://docs.python.org/3/library/csv.html[csv^] library:
```

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+----
+include::python-sdk:howtos:example$import.py[tags=csvImport,indent=0]
+----
+
+[source,python]
+----
+include::python-sdk:howtos:example$import.py[tags=tsvImport,indent=0]
+----
+
+For *JSON* and *JSONL*, use the xref:https://docs.python.org/3/library/json.html[json] library:
```
simon-dew: Link not xref

```suggestion
For *JSON* and *JSONL*, use the https://docs.python.org/3/library/json.html[json^] library:
```

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+
+NOTE: {github}
+--
+====
+
+=== Connect to the Couchbase Server
+
+First, you need the connection details for the Couchbase server.
+
+Now decide which bucket and xref:learn:data:scopes-and-collections.adoc[scope and collection] you want to import to (and create them if they don't already exist.)
```
simon-dew: Broken xref

```suggestion
Now decide which bucket and xref:learn:data/scopes-and-collections.adoc[scope and collection] you want to import to (and create them if they don't already exist.)
```

Hint: always have `nav.adoc` open in your editor. You can copy and paste xrefs from there to make sure you get the Antora xref co-ordinates right.

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+--
+Java::
++
+--
+[source,java]
+----
+include::java-sdk:howtos:example$Import.java[tags=connect,indent=0]
+----
+
+If you are using the xref:java:howtos:concurrent-async-apis.adoc#reactive-programming-with-reactor[Reactive API], then use the reactive collection instead:
```
simon-dew: Broken xref:

```suggestion
If you are using the xref:java-sdk:howtos:concurrent-async-apis.adoc#reactive-programming-with-reactor[Reactive API], then use the reactive collection instead:
```

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+----
+
+Learn more about xref:dotnet-sdk:howtos:kv-operations.adoc[].
+
+NOTE: {github}
+--
+Java::
++
+--
+Hook the data prepared above into an `upsert` routine. For the blocking API, uses the method below:
```
simon-dew: Grammatical fix

```suggestion
Hook the imported data into an `upsert` routine. For the blocking API, use the method below:
```

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+----
+
+Now hook the imports prepared earlier into an `upsertDocument` routine which uses these functions.
+
+[source,python]
+----
+include::python-sdk:howtos:example$import.py[tags=upsertDocument,indent=0]
+----
+
+Learn more about xref:java-sdk:howtos:kv-operations.adoc[].
```
simon-dew: Should this be python?

```suggestion
Learn more about xref:python-sdk:howtos:kv-operations.adoc[].
```

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+Hook the data prepared above into an `upsert` routine. For the blocking API, uses the method below:
+
+[source,java]
+----
+include::java-sdk:howtos:example$Import.java[tags=upsertDocument,indent=0]
+----
+
+The *Reactive API* examples above already include a call to `ReactiveCollection::upsert`.
+
+In both cases, you must provide a preprocess routine which returns a key/value tuple object:
```
simon-dew: Style guide says [key-value](https://github.com/couchbase/docs-site/blob/master/home/modules/contribute/pages/word-list.adoc#k)

```suggestion
In both cases, you must provide a preprocess routine which returns a key-value tuple object:
```

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
++
+--
+Hook the data prepared above into an `upsert` routine. For the blocking API, uses the method below:
+
+[source,java]
+----
+include::java-sdk:howtos:example$Import.java[tags=upsertDocument,indent=0]
+----
+
+The *Reactive API* examples above already include a call to `ReactiveCollection::upsert`.
```
simon-dew, line 557: I won't keep adding comments on all of these, but as a general rule, when you tell users to use a particular Couchbase method, please add a link to the reference documentation for that method. (I know, some SDKs have better / more complete reference documentation than others...)

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+
+NOTE: we use `upsert` rather than `insert` to upload the document even if the target key already has a value.
+This means that in the case of any error, it is easy to make any required tweaks to the import file and re-run the whole import.
+
+[{tabs}]
+====
+pass:[.NET]::
++
+--
+Hook the imports above into an `upsertDocument` routine:
```
simon-dew: Vague. Maybe something like "Hook the imported data ..."

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+----
+
+Learn more about xref:java-sdk:howtos:kv-operations.adoc[].
+
+NOTE: {github}
+--
+Node.js::
++
+--
+To iterate the stream you created earlier, use a simple `for` loop in the same way as an array.
```
simon-dew: Vague. "To iterate the imported data stream..."

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+[source,python]
+----
+include::python-sdk:howtos:example$import.py[tags=key,indent=0]
+----
+[source,python]
+----
+include::python-sdk:howtos:example$import.py[tags=process,indent=0]
+----
+
+Now hook the imports prepared earlier into an `upsertDocument` routine which uses these functions.
```
simon-dew: "Now hook the imported data...."

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+----
+include::python-sdk:howtos:example$import.py[tags=csvImportMulti,indent=0]
+----
+
+NOTE: {github}
+--
+====
+
+
+== Next Steps
```
simon-dew: This is a self-contained how-to guide so there are no next steps.

```suggestion
== Related Links
```

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+
+* xref:n1ql:n1ql-language-reference/join.adoc[JOIN Clause]
+
+Tutorials:
+
+* https://query-tutorial.couchbase.com/tutorial/#1[N1QL Query Language Tutorial^]
+
+See also:
+
+* XDCR
```
simon-dew: Can you add links to the `kv-operations` page for every SDK? Copy from @maria-robobug 's Creating Data how-to guide.

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+
+. Decide on the key for this document (could be an ID, a sequence number, or some combination of fields).
+  * Do any additional processing if required.
+  
+. Insert the document.
+
+// tag::clients_section[]
+// include::partial$clients.adoc[]
+// following section copied/tweaked from above
+// differences: cbc vs cbimport, topic, etc.
```
simon-dew, line 39: If necessary you can tweak the partial so it applies to your use-case as well as existing use-cases?

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+
+* xref:tools:cbimport.adoc[cbimport]
+
+* xref:home::sdk.adoc[SDK Clients]
+
+* xref:manage:import-documents/import-documents.adoc[Couchbase Server UI]
+
+// end::clients_section[]
+
+== Preparing the Data
```
simon-dew, line 71: There's a lot of inconsistency within this page about headings. So this heading says "Preparing" with a participle, whereas the next heading is "Use cbimport" with a bare imperative verb. Our style guide is silent on this, but think we should at least be consistent within the page. So either change all the headings to "-ing" forms or change this to "Prepare the Data"

modules/guides/pages/import.adoc
```diff
@@ -0,0 +1,613 @@
...
+include::python-sdk:howtos:example$import.py[tags=connect,indent=0]
+----
+
+Learn more about xref:python-sdk:howtos:managing-connections.adoc[].
+
+NOTE: {github}
+--
+====
+
+=== Inserting the documents
```
simon-dew: Either change all the headings to "-ing" forms or change this to "Insert the documents". Also be consistent with casing! Personally I would change all headings to use title case, whereas you have a mixture of title case and sentence case.

simon-dew (changes requested):
Thanks for the updates @osfameron ! Most of my comments are to do with consistency of style and tone. I've also suggested breaking down the Parse section as the procedures / examples are quite long.

One more comment to add, outside of the changes you have made:

- [x] Add a link to `import.adoc` in the navigation — there's a dummy entry for this page already

modules/guides/pages/import.adoc
osfameron: It's not "imported" at this point, but rewording throughout for more consistency.. "prepared data".

modules/guides/pages/import.adoc
osfameron: It needed a little more thought to unpick (copy/tweak absolutely isn't ideal!) but can look at this afterwards.

modules/guides/pages/import.adoc
osfameron: Ah, the Flux ones earlier were an external library, but yes, good point!

modules/guides/pages/import.adoc
osfameron: OK, CSV/TSV and JSON/JSONL (e.g. 2 sections) is a much simpler edit, and makes more sense to me.

simon-dew (approved):

simon-dew:
LGTM! 👍

PR #2321, opened 2021-12-08
Re-Purpose for New Developers -  createindex.adoc

robashcom:
CREATE INDEX is logically one of the first pages a new developer will see. With no intro to Couchbase, that user would be confused by this page. Too much complicated stuff right at the top. Too many warnings and caveats. This developer needs the basics first; then more details. I removed a LOT of content from the top of the page and made the language more consistent and succinct.


simon-dew:
Thanks for the contribution! This PR addresses a real issue. However, since Couchbase Server 7.0 we have introduced simple developer guides, including a [Create Indexes](https://docs.couchbase.com/server/current/guides/create-index.html) page, which covers the basics of this topic in a much simpler way for new developers. We've kept the CREATE INDEX page as a complete in-depth reference.

PR #2328, opened 2021-12-14
DOC-9491: Ports 11280, 9123 and 9124 missing from install docs

RayOffiah:
Add missing points 11280 and 9123. Move note on port 9124 to the right location.


b33f (approved):

RichardSmedley (approved):
:+1:

PR #2338, opened 2021-12-22
[DOC-9520] Internal document is given as external link on release notes page

RayOffiah:
See https://issues.couchbase.com/browse/DOC-9520

Corrected link on a page that was:

1.  Pointing to a non-existent ID.
1. Was given as a full link to the global docs site rather than pointing to an external document.


osfameron (approved):
👍

PR #2345, opened 2022-01-04
Doc 9482

tonyjhillman:


modules/rest-api/pages/rest-get-cluster-events.adoc
```diff
@@ -1 +1,143 @@
...
+  http://<ip-address-or-domain-name>:8091/eventsStreaming
+----
+
+The `sinceTime` path-parameter specifies the time (inclusive) from which previously logged events are to be returned.
+The associated `time-value` must be specified in _ISO 8601 UTC_ format &#8212; which is `YYYY-MM-DDThh:mm:ssZ`.
+If this path-parameter is not used, logs starting from the oldest are returned, .
+
+The `limit` path-parameter specifies the maximum number of logged events to be returned.
+The default `limit-value` is `250`.
+If `limit-value` is specified as `1`, no limit is imposed; and all events logged since the specified `sinceTime`, or from the oldest logged event, are therefore returned.
```
hareen-kancharla: ~~Typo: "is specified as '1'" should be "is specified as '-1'".~~
Looks like a subsequent commit took care of this. Thanks.

hareen-kancharla: I think we should included another example for the following case:
if 'sinceTime' is not passed in the query params then the most recent 'limit-value' number of event logs are returned.
For example, GET /events?limit=300 will send the most recent 300 event log.

tonyjhillman: Okay, I've tried to clarify this a bit. I've added the extra example, and have also added some words in the main description of the params. Let me know if this seems right. Thanks.

hareen-kancharla: The changes look good @tonyjhillman. Thanks.

tonyjhillman: Thanks. I'll merge.

PR #2346, opened 2022-01-05
DOC-9538: New reference section for system events

tonyjhillman:


tonyjhillman:
I've modified the user and group security events - however, the group deleted one says '(no additional information)', but seems to have the hashed group-name in the extras field. So I'm not sure - I have put 'Hash of group name': is that correct?

hareen-kancharla:
Yup sorry my bad - it has to be the "hash of the group name".

hareen-kancharla (approved):
All the cluster_manager, data and security related event logs look good. If the services have approved their specific sections - this looks good to commit.

PR #2347, opened 2022-01-05
[DOC-9527]

RayOffiah:
The description for Distributed ACID transactions has been reworked to make it a little clearer.
Note that this change will also need to be made for version 6.6 of the server docs.

https://issues.couchbase.com/browse/DOC-9527


RayOffiah:
Not sure what happened here. Duff merge request

PR #2348, opened 2022-01-05
[DOC-9527] Improved description of distributed ACID transactions

RayOffiah:
In response to the JIRA request:

https://issues.couchbase.com/browse/DOC-9527

The description of distributed ACID transactions has been rewritten.

NOTE: This change will also be required for version 6.6

### DIFF
--- a/modules/learn/pages/data/transactions.adoc
+++ b/modules/learn/pages/data/transactions.adoc
@@ -7,7 +7,7 @@ _Transactions_ guarantee that multiple documents can be updated atomically.
 [#understanding-transactions]
 == Understanding Transactions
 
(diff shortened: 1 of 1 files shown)


RichardSmedley (approved):
I think this is okay to merge -
it does fulfil the JIRA request -
but the https://docs.couchbase.com/server/current/learn/data/transactions.html doc would be better if we could justify the time removing the later features from it (we can't - 6.5 is not where most people will be going for txns advice).

iow Thanks @RayOffiah   LGTM

RayOffiah:
Okay, so there's no need to make the change for 6.6 then?

RichardSmedley:
> Okay, so there's no need to make the change for 6.6 then?

It would be a good idea to make the same improvement at 6.5 in 6.6,
at the very least...
There's little leap in upgrading from 6.5 to 6.6, so we'd expect many more people to be on 6.6 than 6.5.

RayOffiah:
No problem. I'll just copy the text over.

PR #2349, opened 2022-01-05
DOC-9515 "Couchbase Client Deployment Strategies" link goes to the generic SDK intro.

RayOffiah:
The client-server installation page has been removed, so the deployment guide has been reworded to reflect the new location.


RichardSmedley (approved):
I changed the landing page - the other one is no longer maintained, and should not be in the nav.

RichardSmedley:
> I changed the landing page - the other one is no longer maintained, and should not be in the nav.

Needless to say that the why and the what-to-do-about-it are non-trivial ;)

PR #2318, opened 2021-12-07
Doc 4141 server side moxi deprecated

RayOffiah:

PR #2351, opened 2022-01-06
DOC-9521 Inconsistency between Table 1 and Table 2 sections

RayOffiah:
Corrected inconsistencies. Note that the port 11211 is no longer in use:
https://issues.couchbase.com/browse/DOC-4141

Related:

https://github.com/couchbase/docs-server/pull/2318

### DIFF
--- a/modules/install/pages/install-ports.adoc
+++ b/modules/install/pages/install-ports.adoc
@@ -79,7 +79,7 @@ The following table lists all port numbers, grouped by category of communication
 *Encrypted*: 9999, 11206, 11207, 18091-18094, 19102, 19130, 21150
 
 | _Client-to-node_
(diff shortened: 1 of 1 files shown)


dave-finlay (approved):
Hi Ray - I'm not the complete authority on this stuff, but I think I know enough to review. Port 11211 (the old moxi port) has been removed and so it's correct to not have it anywhere in the port list. I'm pretty sure the FTS ports are used node-to-node as you can run a N1QL query on one node and it's possible that it will make a request against an FTS port (either encrypted or unencrypted on another node.) In short, looks good.

RayOffiah:
That's great. Thanks very much for doing this.

simon-dew (approved):
Happy to approve following @dave-finlay 's review :+1:

PR #2358, opened 2022-01-12
Doc 9551 - System Events Overview

tonyjhillman:


modules/learn/pages/clusters-and-availability/system-events.adoc
```diff
@@ -7,38 +7,44 @@
...
 
 System events are recorded in a _system log_, which is maintained by the Cluster Manager.
-Each node in the cluster has a copy of the system log; and each log is kept up to date by means of _gossip-replication_, between the nodes.
-(See xref:learn:clusters-and-availability/metadata-management.adoc[Metadata Management], for information on gossip-replication.)
+Each node in the cluster has a copy of the system log; and each log is kept up to date by means of _gossip replication_ between the nodes.
+(See xref:learn:clusters-and-availability/metadata-management.adoc[Metadata Management], for information on gossip replication.)
 
-System events can be retrieved from the system log by means of the REST API.
-Two methods of retrieval are supported:
+System events do not need to be explicitly enabled by the administrator: they are collected by default, and can be retrieved by the administrator as required.
```
hareen-kancharla, line 29: I think we should also include the default number of cluster-wide events we store (10K) and the info that it is possible to configure the value between 3K to 20K. 
The eventLogsLimit can be changed via the internalSettings endpoint as follow: 
 curl -v -X POST http://Administrator:password@127.0.0.1:9000/internalSettings -d eventLogsLimit=3000

The current eventLogsLimit can be retrived via the internalSettings endpoint as follow:

curl -v  http://Administrator:password@127.0.0.1:9000/internalSettings | jq -c .

{"enforceLimits":false,"indexAwareRebalanceDisabled":false,"rebalanceIndexWaitingDisabled":false,"rebalanceIndexPausingDisabled":false,"rebalanceIgnoreViewCompactions":false,"rebalanceMovesPerNode":4,"rebalanceMovesBeforeCompaction":64,"maxParallelIndexers":4,"maxParallelReplicaIndexers":2,"maxBucketCount":30,"eventLogsLimit":3000,"gotraceback":"single","indexAutoFailoverDisabled":true,"certUseSha1":false,"httpNodeAddition":false} 

In a mixed-mode cluster trying to change eventLogsLimit will return an error reading "not supported in mixed version clusters".

tonyjhillman: I have added to this page, and also updated the internalSettings REST page.

hareen-kancharla:
I think there a lot of places where we are referring to the current limit for the event logs as the maximum. I think it would be less confusing if we just said the default limit for the event logs is 10K and once changed it could be called "the current limit" or something else.

tonyjhillman:
I've removed maximum and used current.

tonyjhillman:
I need to close this PR and open a new one with the same content.

PR #2361, opened 2022-01-14
DOC-9514 Seems to be missing the upgrade step:

RayOffiah:
Added missing text: the node needs to be upgraded before being returned to the cluster.


brentjanorske:
The change looks good and clarifies the process a bit more.

simon-dew (approved):
LGTM, would prefer new line for each separate sentence, but that's just a stylistic thing 👍

PR #2364, opened 2022-01-18
[DOC-9566] External users should use TLS when communicating with cluster

RayOffiah:
https://issues.couchbase.com/browse/DOC-9566

### DIFF
--- a/modules/manage/pages/manage-security/manage-users-and-roles.adoc
+++ b/modules/manage/pages/manage-security/manage-users-and-roles.adoc
@@ -284,6 +284,16 @@ image::manage-security/usersSecurityRows.png[,800,align=left]
 The externally authenticated user is now shown to have the username `externalUser`, and the *Query System Catalog* and *Cluster Admin* roles.
 Their *auth domain* is specified as `External`, indicating that they are authenticated on an external server.
 
+[NOTE]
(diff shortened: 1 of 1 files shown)


modules/manage/pages/manage-security/manage-users-and-roles.adoc
```diff
@@ -284,6 +284,16 @@ image::manage-security/usersSecurityRows.png[,800,align=left]
 The externally authenticated user is now shown to have the username `externalUser`, and the *Query System Catalog* and *Cluster Admin* roles.
 Their *auth domain* is specified as `External`, indicating that they are authenticated on an external server.
 
+[NOTE]
+====
+External users can only be authenticated using the `PLAIN` method. 
+By default, the Couchbase SDK will only default to PLAIN when connecting over TLS ports. +
+To ensure you cannot through the SDK, external users should use the TLS ports.
```
owend74: To ensure you cannot through the SDK - I found this part confusing.
Suggest something like: 
By default, the Couchbase SDK will only default to PLAIN when connecting over TLS ports therefore, external users should use the TLS ports.

tonyjhillman (approved):
Should be 'More details on authentication methods'

PR #2375, opened 2022-01-20
DOC-9577: Headers option for rest passphrase registration

tonyjhillman:


tonyjhillman:
Some questions, so we might add a little more info:

Is the address-family default inet4?

Is the default for httpsOpts" {"verifyPeer": true}?

Is the specifying of one or more headers optional? What are the keys - can they be anything, or must they be key1, key2, etc?

Is the default for timeout 5000?

Thanks!

timofey-barmin:
These are good questions.
Let's make sure the following info is reflected in the doc:
**For "script":**
path - required
trim - optional, true or false, true by default
args - optional, list of strings, empty list by default
timeout - optional, in ms, 5000ms by default

**For "rest":**
url - required
addressFamily - optional, "inet" or "inet6", "inet" by default
timeout - optional, in ms, 5000 by default
headers - optional, json object, {} (no headers) by default
httpsOpts - optional, json object, only valid when https is used in "url"; {"verifyPeer": true} by default; currently "verifyPeer" is the only supported https option

> What are the keys - can they be anything, or must they be key1, key2, etc?
Basically these are simply key-value pairs that can be attached to the rest request. Can be used for many purposes. For authentication, for example.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers

modules/rest-api/pages/upload-retrieve-node-cert.adoc
```diff
@@ -85,14 +85,16 @@ A JSON object of the following kind registers a passphrase of type `rest`:
...
 
 The specified `type` is `rest`.
-The value of `url` must be the endpoint to be called with `GET`.
-The `addressFamily` specifies the address family to be used, and can be either `inet4` or `inet6`.
-The value of `httpsOpts` must be a JSON key-value pair whose key is `verifyPeer`, and whose value is either `true` or `false`: this determines whether peer verification is performed if Https is used.
-The value of `timeout` specifies the number of milliseconds that must elapse before the call is timed out.
+The value of `url`, which is a _required_ argument, must be the endpoint to be called with `GET`: the endpoint can use either `http` or `https`.
+The `addressFamily`, which is an _optional_ argument, specifies the address family to be used, and can be either `inet` (meaning IPv4) or `inet6` (meaning IPv6): the default value is `inet`.
+The value of `httpsOpts`, which is an _optional_ argument that is valid only when `https` is used in the value specified for `url`, must be a JSON key-value pair whose key is `verifyPeer`, and whose value is `true`: this determines that peer verification is performed when Https is used &#8212; `true` is the only value supported, and is the default.
```
timofey-barmin: verifyPeer is the only option supported in httpsOpts
verifyPeer can be either true or false, true by default

tonyjhillman: Thanks, I have rewritten.

timofey-barmin (approved):
Thanks Tony! LGTM

PR #2377, opened 2022-01-20
DOC-9585: Rename fts-creating-index.adoc to fts-creating-indexes.adoc

abhinavdangeti:


abhinavdangeti:
@RayOffiah would you review this for me?

RayOffiah:
Quick note. When you change and/or move a file, then put the original name and/or location in a page alias. That way, you won't need to change all the references to it in other files; they'll be automagically redirected for you. --> @abhinavdangeti

RayOffiah (approved):

modules/fts/pages/fts-creating-indexes.adoc
```diff
@@ -1,5 +1,6 @@
 [#Creating-Indexes]
 = Creating Indexes
+:page-aliases: fts-creating-index.adoc
```
abhinavdangeti, line 3: @RayOffiah But there are no other references to fts-creating-index.adoc.

Prior to this change I found references to fts-creating-index.adoc and fts-creating-indexes.adoc. I've fixed one of them as part of this commit.

RayOffiah: Yes, but having the page-aliases also means that if a user has bookmarked the old page name, then they'll be redirected to the new page name. --> @abhinavdangeti

PR #2389, opened 2022-01-25
[rayoffiah] tutorial fix for the new couchbase download page

RayOffiah:
The couchbase download server page looks a bit dfferent, so the images have been updated. 
Also corrected a problem with the incorrect version number being shown on the server setup screen.


maria-robobug:
Hmm, should `modules/tutorials/images/launch-couchbase-server-page.png` be an EE image also?

RayOffiah:
Good question. 

I left it as community version because I reckoned devs would be more comfortable with something they could play with over an extended period of time (possibly in their own time), rather than worrying about hitting the end of the trial. --> @maria-robobug

maria-robobug (approved):

PR #2394, opened 2022-01-26
[DOC-9611] Wrong output for OBJECT_INNER_VALUES

RayOffiah:
Created a new example. 
Let me know if you think of a better one!


RayOffiah:
On a closer look, this appears to be behaving correctly.

PR #2399, opened 2022-01-28
[DOC-9611] Wrong output for OBJECT_INNER_VALUES

RayOffiah:
Interesting one. 
The examples are correct, but if you come into the page from a search, then you'll miss the important details at the top instructing you to update the travel sample bucket.

Added admonitions in each section to remind the reader to follow the instructions at the top before attempting to run the examples.

Corrected a few inconsistencies and a minor bug in the JSON sample.


modules/n1ql/pages/n1ql-language-reference/objectfun.adoc
```diff
@@ -19,6 +20,9 @@ SET schedule[0] = {"day" : 7, "special_flights": [
 WHERE destinationairport = "CDG" AND sourceairport = "TLV";
 ----
 
+
+:load-example-warning: Make sure that you have run the `UPDATE` command described in <<load-example-data, this section>> before running the example(s).
```
simon-dew: I think you need to do something to make sure the backticks are passed through properly. They're being rendered literally in the output (rather than as monospace text). Maybe `pass:q[]` macro?

simon-dew (changes requested):
Does the job, should help out users who jump straight to an individual function 👍

Just one issue with the markup.

modules/n1ql/pages/n1ql-language-reference/objectfun.adoc
RayOffiah: @simon-dew <-- Fixed.

modules/n1ql/pages/n1ql-language-reference/objectfun.adoc
```diff
@@ -19,6 +20,9 @@ SET schedule[0] = {"day" : 7, "special_flights": [
 WHERE destinationairport = "CDG" AND sourceairport = "TLV";
 ----
 
+
+:load-example-warning: pass:q[Make sure that you have run the `UPDATE` command described in <<load-example-data, this section>> before running the example(s).]
```
simon-dew: Well, the backticks are now working properly, but the xref shorthand is now replaced by `<>`. See the [preview doc](https://github.com/couchbase/docs-server/blob/70ef31e48ec75e86f1482f5c4fbdf5489126f048/modules/n1ql/pages/n1ql-language-reference/objectfun.adoc).

To work around this, replace the xref shorthand with an xref macro. I've tested this offline and it now seems to work.

 ```suggestion
:load-example-warning: pass:q[Make sure that you have run the `UPDATE` command described in xref:load-example-data[this section] before running the example(s).]
```

simon-dew (changes requested):
More pass-through markup shenanigans. Sorry, I think I should have tested this more thoroughly last time.

simon-dew (approved):

PR #2400, opened 2022-01-28
DOC-9465for7.1: Added security checklist

tonyjhillman:


b33f:
Is there any way to make the website layout look similar to the PDF, with a table and checkboxes ?

PR #2405, opened 2022-01-31
DOC-9631 Overly complicated Windows installation instructions.

RayOffiah:
Removed instructions that directed reader to use the command line.


simon-dew (approved):
LGTM 👍 (I don't have Windows, so can't test the procedure, but the edit itself is fine.)

PR #2413, opened 2022-02-01
DOC-9132: Updated information on TLS 1.3 support

tonyjhillman:


tonyjhillman:
I have:
a) Used ::1 instead of localhost.
b) Stated that all services, plus cluster manager, xdcr, and views, are supported by TLS 1.3
c) Left unchanged the information that states that only Data and Analytics services can perform custom-configuration of cipher suites under 1.3 (this may have changed - possibly all can do it now, but I'm not sure)
Thanks.

modules/learn/pages/security/on-the-wire-security.adoc
```diff
@@ -63,9 +63,9 @@ The relationships between _global_ and _per service_ settings for TLS and cipher
 
 The minimum TLS version can be established either globally or per service: the server will not accept client connections with protocols below the level of the established, minimum version.
 
-The minimum version can be established globally and for all services as `tlsv1`, `tlsv1.1`, or `tlsv1.2`.
-A value of `tlsv1.3` can be established for the following services: _Data_, _Search_, _Index_, _Query_, _Eventing_, _Analytics_, and _Backup_; it _cannot_ be established either globally or for the Cluster Manager.
-See xref:learn:security/on-the-wire-security.adoc#cipher-suite-configuration-limitations[Cipher-Suite Configuration Limitations], below, for further information.
+The minimum version can be established globally and for all services as `tlsv1`, `tlsv1.1`, `tlsv1.2`.
```
bryandmc: Should there be a trailing period after this but before the .. "or `tlsv1.3`." part?

modules/learn/pages/security/on-the-wire-security.adoc
```diff
@@ -115,13 +115,12 @@ The following limitations apply to cipher-suite configuration.
 [#tls-3-cipher-suite-limitations]
 === TLS 1.3 Cipher-Suite Limitations
 
-As indicated above, only the _Data_, _Search_, _Index_, _Query_, _Eventing_, _Analytics_, and _Backup_ services can use TLS 1.3 cipher-suites.
-TLS 1.3 cipher-suites _cannot_ be used by the Cluster Manager.
+As indicated above, TLS 1.3 cipher-suites can by used by all services; and by the Cluster Manager, XDCR, and Views.
 
-Of the services that _can_ use TLS 1.3 ciphers, only the _Data_ and _Analytics_ services support the custom-configuration of TLS 1.3 ciphers &#8212; _custom-configuration_ of ciphers means the selection of a subset of available ciphers for a custom cipher-suite list, and the specifying of the selected ciphers in the subset in any order judged appropriate.
+However, only the _Data_ and _Analytics_ services support the custom-configuration of TLS 1.3 ciphers &#8212; _custom-configuration_ of ciphers means the selection of a subset of available ciphers for a custom cipher-suite list, and the specifying of the selected ciphers in the subset in any order judged appropriate.
```
bryandmc: I don't know what the fully correct statement here would be, but I do know that ns_server (cluster manager) also supports custom-configuration of tls 1.3 ciphers. So at the very least that needs to be added, but someone else will have to confirm whether others do as well.

anuthan: Agree that cluster manager should be included for custom-configuration of TLS 1.3. 
The golang services,(index, n1ql, backup, eventing) TLS1.3 ciphers suites aren't customizable IIRC. 
@bryandmc we should check with service endpoints too, similar to what was tried for ns_server endpoint, i.e., after configuring the cipher suites you can test what cipher that particular golang service endpoint offers using scripts/test-ciphers.sh in ns_server directory.

bryandmc: will do @anuthan -- will report back with results.

bryandmc: yeah @anuthan, Timofey was right in the metting. After testing, all the go based services seem to have the same limitation of not being able to set custom ciphers for tlsv1.3. So @tonyjhillman the correct version of this needs to add in cluster manager, but the rest is fine. That is to say: index, n1ql, backup, eventing .. etc still don't support setting tlsv1.3 cipher suites.

modules/learn/pages/security/on-the-wire-security.adoc
tonyjhillman: Yes, thank you.

modules/learn/pages/security/on-the-wire-security.adoc
tonyjhillman: Okay, thanks. I have changed to say: 
However, only _Cluster Manager_ and the _Data_ and _Analytics_ services support the custom-configuration of TLS 1.3 ciphers

modules/learn/pages/security/on-the-wire-security.adoc
```diff
@@ -117,7 +116,7 @@ The following limitations apply to cipher-suite configuration.
 
 As indicated above, TLS 1.3 cipher-suites can by used by all services; and by the Cluster Manager, XDCR, and Views.
 
-However, only the _Data_ and _Analytics_ services support the custom-configuration of TLS 1.3 ciphers &#8212; _custom-configuration_ of ciphers means the selection of a subset of available ciphers for a custom cipher-suite list, and the specifying of the selected ciphers in the subset in any order judged appropriate.
+However, only _Cluster Manager_ and the _Data_ and _Analytics_ services support the custom-configuration of TLS 1.3 ciphers &#8212; _custom-configuration_ of ciphers means the selection of a subset of available ciphers for a custom cipher-suite list, and the specifying of the selected ciphers in the subset in any order judged appropriate.
```
bryandmc: Shouldn't this be: `However, only _Cluster Manager_, _Data_ and _Analytics_ services support the custom-configuration of TLS 1.3 ciphers`

anuthan:
I'm fine with the changes, once Bryan's comments are fixed

modules/learn/pages/security/on-the-wire-security.adoc
tonyjhillman: I don't actually think so, because the Cluster Manager is not formally referred to as a 'service'.

tonyjhillman: But, I do think an article is required before 'Cluster Manager'. Thank you.

tonyjhillman: Okay, I've added. I hope that is clearer.

modules/learn/pages/security/on-the-wire-security.adoc
```diff
@@ -116,7 +116,7 @@ The following limitations apply to cipher-suite configuration.
 
 As indicated above, TLS 1.3 cipher-suites can by used by all services; and by the Cluster Manager, XDCR, and Views.
 
-However, only _Cluster Manager_ and the _Data_ and _Analytics_ services support the custom-configuration of TLS 1.3 ciphers &#8212; _custom-configuration_ of ciphers means the selection of a subset of available ciphers for a custom cipher-suite list, and the specifying of the selected ciphers in the subset in any order judged appropriate.
+However, only the _Cluster Manager_ and the _Data_ and _Analytics_ services support the custom-configuration of TLS 1.3 ciphers &#8212; _custom-configuration_ of ciphers means the selection of a subset of available ciphers for a custom cipher-suite list, and the specifying of the selected ciphers in the subset in any order judged appropriate.
```
bryandmc: I don't mean to nitpick at all but the change I was hoping to see was to use the form:
`only Thing1, Thing2, and FinalThing` ... instead of `only Thing1 and Thing2 and FinalThing`.. See how we have the multiple and's ? Generally we should just use `and` for the final item. The `the` isn't as important, though I guess is also fine.

bryandmc (changes requested):

modules/learn/pages/security/on-the-wire-security.adoc
bryandmc: Maybe something like:
```
However, only the _Cluster Manager_, _Data_ and _Analytics_ services support the custom-configuration of TLS 1.3 ciphers ...
```

tonyjhillman: You are not nit picking and I don't mind the discussion at all. But I do disagree.

I  understand (I think) what you are saying and its motivation. However, the use of 'and' is, I believe, required; in order to ensure that 'services' applies only to Thing2 and FinalThing, but not to Thing1. It seems to me that what you are proposing would be interpreted as  "However, only the Cluster Manager service, the Data service, and the Analytics service support the custom-configuration ...etc." This is incorrect, because the Cluster Manager is not a service. I recognise that the absence of a comma after "Data" might mitigate the misinterpretation, but only, I think, to leave the sentence ambiguous.

Regarding the use of 'and' in lists: I agree it is simpler to reserve it for separating the last two items; but all the same, it is very typically used to separate elements within an item. So: "The options for printing include color and black and white." That's two options, not three. 

A rewording that avoids such problems might be: "However, only the Cluster Manager, the Data Service, and the Analytics Service support ... etc". That removes ambiguity, and uses 'and' just once; at the expense of a more formal and necessarily extended naming of the services.

At any rate, I have rewritten the sentence according to the wording you proposed. Let me know if you can approve. Thanks again.

bryandmc (approved):

anuthan (approved):
Thanks @tonyjhillman and @bryandmc .

modules/learn/pages/security/on-the-wire-security.adoc
bryandmc: Wait so with this explanation maybe we really should do it the way it was before you made this change. I guess I didn't know we didn't consider cluster manager a service.. So @tonyjhillman I will gladly approve either. So if you feel strongly about it feel free to change it and i'll re-approve that as well.

PR #2418, opened 2022-02-02
DOC-9640 Minor bug on the Magma storage engine page

RayOffiah:
Changed (> 5TB) to "several terabytes"

### DIFF
--- a/modules/learn/pages/buckets-memory-and-storage/storage-engines.adoc
+++ b/modules/learn/pages/buckets-memory-and-storage/storage-engines.adoc
@@ -67,7 +67,7 @@ You should use the Couchstore backend if:
 You should use the Magma backend if:
 
 * your working set is much larger than the available memory and you need disk access speed only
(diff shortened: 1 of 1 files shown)


modules/learn/pages/buckets-memory-and-storage/storage-engines.adoc
```diff
@@ -67,7 +67,7 @@ You should use the Couchstore backend if:
 You should use the Magma backend if:
 
 * your working set is much larger than the available memory and you need disk access speed only
-* you need to store large amounts of data in a single node (>{nbsp}5{nbsp}TB) using lowest amount of memory
+* you need to store large amounts of data in a single node (several terabytes) using lowest amount of memory
```
shivaniguptacb: This is still not correct. Here is the line I had suggested in DOC-9640:
"you need to store and access large amounts of data (several TBs) using lowest amount of memory"

There is no need to mention 'in a single node'

shivaniguptacb (changes requested):

shivaniguptacb (approved):

PR #2421, opened 2022-02-03
DOC-9650 Note on eviction is in the wrong place.

RayOffiah:
Fixed (I think)

@shivaniguptacb, @stevewatanabe  <-- I just want someone to take a quick look at this. We seem to have  `ejection` and `eviction` used here. Are they the same thing?


shivaniguptacb (approved):
Yes, Ejection and Eviction are used to refer to the same thing. The new location of the note looks correct.

PR #2423, opened 2022-02-03
[rayoffiah] test_kroki_on_production

RayOffiah:
Changing diagram to PlantUML definition.


maria-robobug (approved):
Nice, tested it on https://www.planttext.com/, seems to render just fine 👍

PR #2426, opened 2022-02-04
DOC-8000 Remove Microsoft Windows Server 2016 in Neo

RayOffiah:


b33f (changes requested):
The staging docs page still says, 
2016 Standard Edition (Deprecated)
2016 DataCenter Edition (Deprecated)

PR #2442, opened 2022-02-10
DOC-9374 Direct upgrade from 6.0.x to 7.1.x is not supported

RayOffiah:
Added 7.x to the upgrade tables.

### DIFF
--- a/modules/install/pages/upgrade.adoc
+++ b/modules/install/pages/upgrade.adoc
@@ -33,14 +33,14 @@ All supported upgrades can be performed with the cluster either _offline_ or _on
 | Starting Version |  Path to Current Version
 
 | 5.0.x
-| 5.0.x and 5.1.x -> 6.0.4+ -> 6.6 -> 7.0
+| 5.0.x and 5.1.x -> 6.0.4+ -> 6.6 -> 7.0 -> 7.1x
 
-5.0.x and 5.1.x -> 6.6 -> 7.0
+5.0.x and 5.1.x -> 6.6 -> 7.0 -> 7.1x
 
-5.5.0+ -> 6.6 -> 7.0
+5.5.0+ -> 6.6 -> 7.0 -> 7.1x
 
 | 6.x
-| 6.x -> 7.0
+| 6.x -> 7.0 -> 7.1x
 
 |===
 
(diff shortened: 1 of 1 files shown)


dave-finlay (commented):
Hi Ray - I'm not sure exactly how you want to present this and if the idea of "upgrade path" is the best way to do it. Let me list the valid upgrades to help illustrate:

7.1.x: 6.5.x, 6.6.x
7.0.x: 6.6.x, 6.5.x, 6.0.x
6.6.x: 6.5.x, 6.0.x, 5.5.x, 5.1.x, 5.0.x
6.5.x: 6.0.x, 5.5.x, 5.1.x, 5.0.x
6.0.x: 5.5.x, 5.1.x, 5.0.x

(The way to read the above information is: you can upgrade to 7.1.x from any 6.5 or 6.6 release only; and you can't upgrade say from 6.0.x to 7.1.x). 

In any case, you can get to 7.1 from 5.0 using a large variety of paths. E.g.
5.0.x -> 5.1.x -> 5.5.x -> 6.0.x -> 6.5.x -> 6.6.x -> 7.0.x -> 7.1.x (longest)
5.0.x -> 6.6.x -> 7.1.x (shortest)

And *lots* of paths in between that I don't feel like comprehensively specifying. In particular I don't know why 6.0.4 is so specifically called out. Generally one should use the latest build on every release branch when upgrading.

I'm not sure if you want to exhaustively list all the upgrade paths, but I'm not sure if I want to ask you to rewrite this whole section and present upgrade differently. I guess I'll give you the information and let you decide how best to proceed. Let me know if you have questions.

RayOffiah:
@dave-finlay, @tonyjhillman  <--

Definitely see your point, but I don't think we'll have time for a complete rewrite before the Neo release unfortunately.

I would probably say the best solution is to make sure that we note that you can't go from 6 to 7.1 (which I think we do), and  then look at revamping the whole page when we have more time.

I'd like to look and see how other sites document upgrade paths. I agree that we should not be noting down every single path as it does seem a lot to read.

I wonder if it would be better to arrange it like the release notes page: A heading with the latest version, and which version you can upgrade from underneath?

dave-finlay:
> I would probably say the best solution is to make sure that we note that you can't go from 6 to 7.1 (which I think we do), and then look at revamping the whole page when we have more time.

Yes, I think this is the key point that may negatively surprise some users. 

>Definitely see your point, but I don't think we'll have time for a complete rewrite before the Neo release unfortunately.

Understood. Perhaps we should file a DOC ticket to track a larger change to this section.

> I wonder if it would be better to arrange it like the release notes page: A heading with the latest version, and which version you can upgrade from underneath?

Docs are much more discoverable and a much more user friendly way for users to find out important things like if they can upgrade directly. I would favor keeping this in the docs.

RayOffiah:
@dave-finlay  <-- Okay, I'm going to merge this PR and create a ticket to revamp the whole page after Neo. I'll have a look at some examples from other sites.

> Docs are much more discoverable and a much more user friendly way for users to find out important things like if they can upgrade directly. I would favor keeping this in the docs.

Yes, definitely keeping it in the same place; just need to decide on the new structure. I'll put it to the team when I create the ticket.

PR #2465, opened 2022-02-17
Bad permission syntax GET /api/pindex

jon-strabala:
Simple change on only the GET /api/pindex call (the first item) as per DOC-9337
update for 7.0 (this PR) and 7.1


jon-strabala:
Aborted the last PR, I think this is what we want - correct?

abhinavdangeti (approved):
Looks good.

PR #2478, opened 2022-02-22
DOC-9686: type mapping examples formatting changes

Thejas-bhat:
It looks like one of the hanging pages(fts-type-mappings-Docid-with-regexp) was updated instead of the page that's part of the fts ToC(fts-type-mappings). This is something to clean up as per ticket DOC-9722. So, this PR contains the corrections regarding the same.

### DIFF
--- a/modules/fts/pages/fts-type-mappings.adoc
+++ b/modules/fts/pages/fts-type-mappings.adoc
@@ -397,52 +397,55 @@ For example, while working with the `travel-sample` bucket,  set up docid_regexp
 * airport
 
 Below is a full index definition using it.
+[source, JSON]
+----
 {
-    "name": "airline-airport-index",
-    "type": "fulltext-index",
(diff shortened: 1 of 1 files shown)


modules/fts/pages/fts-type-mappings.adoc
```diff
@@ -397,6 +397,8 @@ For example, while working with the `travel-sample` bucket,  set up docid_regexp
 * airport
 
 Below is a full index definition using it.
+[source, JSON]
+----
 {
     "name": "airline-airport-index",
     "type": "fulltext-index",
```
abhinavdangeti: Would you also fix the indentation of this definition within this commit?

abhinavdangeti (commented):
Looks good, just one comment.

abhinavdangeti (approved):

PR #2484, opened 2022-02-23
Updating Server Group documentation

sreekanth-cb:


modules/fts/pages/fts-high-availability-for-search.adoc
```diff
@@ -15,7 +15,28 @@ High availability indexes help users prevent downtime caused by unplanned incide
...
+
+ 
+Following the group definition and rebalance, the primary partitions for any Search indexes are located on one group, while the corresponding replicas are located on another server group and this makes the Group recovery possible. Whenever an entire group/rack goes offline making the primary index partitions inaccessible, the replica index partitions that remain available on another group can be failed over and promoted to primary/active status. Such failure protection is optimal when the groups are assigned `equal numbers of nodes`, and index partition replicas are thereby distributed such that none ever occupies the same group along with its associated parimary partitions.
```
Thejas-bhat: A small typo-
```suggestion
Following the group definition and rebalance, the primary partitions for any Search indexes are located on one group, while the corresponding replicas are located on another server group and this makes the Group recovery possible. Whenever an entire group/rack goes offline making the primary index partitions inaccessible, the replica index partitions that remain available on another group can be failed over and promoted to primary/active status. Such failure protection is optimal when the groups are assigned `equal numbers of nodes`, and index partition replicas are thereby distributed such that none ever occupies the same group along with its associated primary partitions. 
```

modules/fts/pages/fts-high-availability-for-search.adoc
```diff
@@ -15,7 +15,28 @@ High availability indexes help users prevent downtime caused by unplanned incide
...
+
+Search service would place the index replicas on separate server groups as long as there is enough number of server groups (with Search service) equal to the number of replicas available in the cluster. If not, then it would end up positioning multiple replicas on the same server group/rack.
+
+Optimum distribution - when the number of index replicas created for a given index is at least one less than the total number of groups for the cluster, and each group contains sufficient nodes running the Search Service, automatic distribution ensures that each index and index replica resides on its own exclusive group. This configuration gives the High Availability for the Search service.
```
Thejas-bhat: Just a thought here, this particular section of optimum distribution looks rather important. Can we perhaps make it as a note (or a tip) which would make it pop up more, so that the user can quickly notice something important like this?

sreekanth-cb: done @Thejas-bhat ! thanks!

abhinavdangeti (approved):

jon-strabala (approved):
LGTM

PR #2491, opened 2022-02-23
DOC-9574: Add a note that N1QL can select any qualifying FTS index

abhinavdangeti:


abhinavdangeti:
A forward merge (https://github.com/couchbase/docs-server/pull/2490) should cover this.

PR #2497, opened 2022-02-25
Configure index options content update

Thejas-bhat:
This PR aims to update some content under 3 subsections in the configure index options page:

- Date/Time Parser - There were capella specific information which was repeated which have been dropped since this isn't a capella specific doc page.
- Custom Filters - Same as above and updated the content with a screenshot of the custom filters panel.
- Index Type - Outdated screenshot.


jon-strabala (approved):
Looks good

Note, at some point we might reintroduce capella blocks via a conditional macro if documentation becomes multitargeted (which it is not right now).

abhinavdangeti (approved):

PR #2501, opened 2022-02-25
DOC-9748: Remove cbbackup and cbrestore guides

yonsken:


modules/ROOT/nav.adoc
```diff
@@ -255,7 +255,7 @@
   *** xref:fts:fts-advanced-settings-CBFT-ENV-OPTIONS.adoc[CBFT_ENV_OPTIONS]
  ** xref:fts:fts-monitor.adoc[Statistics and Monitoring]
  ** xref:fts:fts-troubleshooting.adoc[Troubleshooting and FAQs]
- 
+
  * xref:eventing:eventing-overview.adoc[Eventing]
```
simon-dew: Eventing is part of the Develop section, so no need to have a blank line here. But the top-level entry for Eventing should be flush with left margin to keep the indentation consistent.

```suggestion
* xref:eventing:eventing-overview.adoc[Eventing]
```

simon-dew (approved):
LGTM :+1: just one minor markup suggestion.

PR #2523, opened 2022-03-09
DOC-9595: CBO statistics new system keyspaces

simon-dew:

* [Monitor Queries › system:dictionary](https://simon-dew.github.io/docs-site/DOC-9595/server/current/manage/monitor/monitoring-n1ql-query.html#sys-dictionary) — new system keyspace
* [Monitor Queries › system:dictionary](https://simon-dew.github.io/docs-site/DOC-9595/server/current/manage/monitor/monitoring-n1ql-query.html#sys-dictionary-cache) — new system keyspace
* [Monitor Queries › system:dictionary](https://simon-dew.github.io/docs-site/DOC-9595/server/current/manage/monitor/monitoring-n1ql-query.html#sys-functions) — new fields
* [Monitor Queries › system:dictionary](https://simon-dew.github.io/docs-site/DOC-9595/server/current/manage/monitor/monitoring-n1ql-query.html#sys-functions-cache) — new fields
* [UPDATE STATISTICS](https://simon-dew.github.io/docs-site/DOC-9595/server/current/n1ql/n1ql-language-reference/updatestatistics.html) — added sections on **Index Residency** and **Monitoring Optimizer Statistics**

Docs issues: [DOC-9595](https://issues.couchbase.com/browse/DOC-9595) and [DOC-9720](https://issues.couchbase.com/browse/DOC-9720)

### DIFF
--- a/modules/manage/pages/monitor/monitoring-n1ql-query.adoc
+++ b/modules/manage/pages/monitor/monitoring-n1ql-query.adoc
@@ -1999,17 +1999,19 @@ This will result in a list similar to:
   {
     "functions": {
       "definition": {
-        "#language": "inline", // <1>
-        "expression": "substr0(`vString`, (length(`vString`) - `vLen`), `vLen`)", // <2>
-        "parameters": [ // <3>
-          "vString",
-          "vLen"
-        ]
+        "#language": "inline", // <.>
+        "expression": "(((`fahrenheit` - 32) * 5) / 9)", // <.>
+        "parameters": [ // <.>
+          "fahrenheit"
+        ],
+        "text": "((fahrenheit - 32) * 5/9)" // <.>
       },
       "identity": {
-        "name": "rstr",
-        "namespace": "default",
-        "type": "global"
+        "bucket": "travel-sample", // <.>
+        "name": "celsius", // <.>
+        "namespace": "default", // <.>
+        "scope": "inventory", // <.>
+        "type": "scope" // <.>
       }
     }
   },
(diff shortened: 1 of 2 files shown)


simon-dew:
@miaobingjie — the spec doc states that `system:dictionary_cache` returns the same fields as `system:dictionary`,

> except for an extra node field that shows entries from different query nodes.

I spent a while playing around with multiple query nodes, but I couldn't get this extra `node` field to appear. Have you got a test case or an example I could see?

marcogrecopriolo:
You are so right! A bit too late for me, but I'll fix it tomorrow

miaobingjie:
With regard to system:dictionary and system:dictinoary_cache:
_#3: An array of indexes available in the keyspace_
this is not quite accurate. An index will only show up in this array if index statistics is available for the index. In other words, there could be indexes defined on the keyspace but not showing in this array.
_#4: An array of statistics for the index_
this is correct, but you might want to add: with one element for each partition of the index. That is to say, for a non-partitioned index, we will have one element in the array; for a partitioned index, we will have one element for each index partition in the array.

With regard to the "node" field when selecting from system:dictionary_cache, you will need to make sure there are entries in the dictionary cache for each query node, i.e., run a query on each of the query node referencing the keyspace. The dictionary cache is loaded on-demand for a keyspace, i.e. you'll need to reference the keyspace on that query node. Once you've done this, then try selecting from system:dictionary_cache from one of the query node, you should see entries from both query nodes, with the entry representing the remote node having the extra "node" field.

marcogrecopriolo:
@simon-dew the node information was missing from the entries from the local node.
It should be fixed in the next build.

simon-dew:
> I couldn't get this extra `node` field to appear

Okay, I understand this will appear in local nodes in later builds. For now, running queries from two separate query nodes fixed this for me.

PR #2526, opened 2022-03-10
Split cbas_path values for improved readability

mblow:


modules/rest-api/pages/rest-node-index-path.adoc
```diff
@@ -38,7 +38,10 @@ Curl request syntax:
 curl -X POST -u [admin]:[password]
     -d path=[data storage path]
     -d index_path=[index storage path]
-    -d cbas_path=[analytics storage paths]
+    -d cbas_path=[analytics storage path 1]
+    [-d cbas_path=[analytics storage path 2],
```
simon-dew, line 42: Pretty sure the comma here is wrong

```suggestion
    [-d cbas_path=[analytics storage path 2]
```

simon-dew (changes requested):

simon-dew:
Assigning this to @ggray-cb . As it comes from @mblow I think we can assume that the content is accurate 😉 

Just one requested change re: comma in the syntax.

PR #2544, opened 2022-03-17
DOC-8867 Index auto-failover

RayOffiah:
Noted dangers around possible loss of the index service.


tonyjhillman (approved):
Looks good, though the convention is to use initial caps on services: ie, Data Service and Index Service.

PR #2550, opened 2022-03-18
DOC-9810 Should not use 'dpkg -i' to install couchbase-release package

RayOffiah:
Fixed

### DIFF
--- a/modules/install/pages/install-uninstalling.adoc
+++ b/modules/install/pages/install-uninstalling.adoc
@@ -50,7 +50,7 @@ Ubuntu and Debian::
 --
 [source,console]
 ----
-sudo dpkg -r couchbase-server
+sudo apt purge couchbase-server
 ----
 --
 ====
--- a/modules/install/pages/ubuntu-debian-install.adoc
+++ b/modules/install/pages/ubuntu-debian-install.adoc
@@ -43,7 +43,7 @@ curl -O https://packages.couchbase.com/releases/couchbase-release/couchbase-rele
 +
 [source,console]
 ----
-sudo dpkg -i ./couchbase-release-1.0-amd64.deb
+sudo apt-get install -y ./couchbase-release-1.0-amd64.deb
 ----
 +
 The meta package installs the necessary information for `apt` to be able to retrieve all of the necessary Couchbase Server installation packages and dependencies.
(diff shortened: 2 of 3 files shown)


modules/install/pages/ubuntu-debian-install.adoc
```diff
@@ -161,7 +161,7 @@ sudo apt-get update
 +
 [source,console,subs=+quotes]
 ----
-sudo dpkg -i ./[.var]_package-name_.deb
+sudo apt-get install -y ./[.var]_package-name_.deb
 ----
 +
 If `dpkg` reports any errors about missing dependencies, issue the following command to download and install those dependencies from the internet:
```
ceejatec: This line and the following [source] block should be removed, since the instructions no longer say to invoke "dpkg". (apt-get will automatically download and install the required dependencies; that's the main reason I recommend using it instead of raw dpkg.)

ceejatec (changes requested):
(See code comment)

modules/install/pages/install-uninstalling.adoc
```diff
@@ -50,7 +50,7 @@ Ubuntu and Debian::
 --
 [source,console]
 ----
-sudo dpkg -r couchbase-server
+sudo apt purge couchbase-server
```
RayOffiah, line 53: @ceejatec <-- Not that familiar with Linux, so is this the correct way to remove the package, given that the intro in the instructions says that all the configuration files have to be removed as well?

RichardSmedley: Yes, that's fine - apt remove takes out an installed package, but apt purge will remove associated config files.

RayOffiah:
@ceejatec <-- Fixes done.

RichardSmedley:
We should update the version of Couchbase installed in the doc from 6.x to 7.x, on lines 82, 96, 120, and 134.

RayOffiah:
> We should update the version of Couchbase installed in the doc from 6.x to 7.x, on lines 82, 96, 120, and 134.

@RichardSmedley <-- Done

RichardSmedley (approved):
LGTM  :+1:

ceejatec (approved):

PR #2552, opened 2022-03-21
DOC-9352 Document how to obtain Capella host name

RayOffiah:
Added documentation explaining how to get the Capella instance name from the console.


simon-dew (commented):
This is great, but it really should be in the [Capella documentation](https://github.com/couchbasecloud/couchbase-cloud), not the Server docs. (We do need to think about de-coupling tutorials and guides, so that they apply equally across Capella and Server — but until we do that, this needs to be in the Capella documentation.)

simon-dew:
I think the initial DOC issue was created in the wrong place.

RayOffiah:
> I think the initial DOC issue was created in the wrong place.

Ah okay.

PR #2300, opened 2021-11-30
Doc 9184 document support for unbalanced server group

RayOffiah:

PR #2554, opened 2022-03-22
DOC-9850 Document the limitation of server group data placement with respect to replica count in 7.1

RayOffiah:
Moved changes from https://github.com/couchbase/docs-server/pull/2300 which were attached to the incorrect branch.


tonyjhillman (approved):

modules/learn/pages/clusters-and-availability/groups.adoc
```diff
@@ -59,16 +59,35 @@ Note that Group 2 contains all the replica vBuckets that correspond to active vB
 [#vbucket-distribution-across-unequal-groups]
 === Unequal Groups
 
+A number of constraints come into play when allocating active and replica vbuckets across server groups:
+
+[horizontal]
+
+*rack-zone*:: all replicas of a given vbucket must reside in separate server groups.
+
+*active balance*:: there should be approximately{empty}footnote:approximate-vbuckets[Note that by "approximately the same" we mean the number should vary by 5 vbuckets or less] the same number of active vbuckets on each node in the cluster.
```
dave-finlay: I think we shouldn't specify a hard number on the max difference of active vbucket counts on different nodes. Generally, the active vbucket counts will only vary by 1 across nodes, but we might loosen that in the future. Also I don't want customers with 100 nodes with an average of 10 vbuckets per node complaining that some nodes might have as many as 15 vbuckets and carry 50% greater load. I'd prefer to just say the the counts are approximately the same and leave it at that.

modules/learn/pages/clusters-and-availability/groups.adoc
```diff
@@ -59,16 +59,35 @@ Note that Group 2 contains all the replica vBuckets that correspond to active vB
...
+
+*replica balance*:: there should be approximately{empty}footnote:approximate-vbuckets[] the same number of replica vbuckets on each node in the cluster.
```
dave-finlay: Same comment here about approximate counts of vbuckets.

modules/learn/pages/clusters-and-availability/groups.adoc
```diff
@@ -59,16 +59,35 @@ Note that Group 2 contains all the replica vBuckets that correspond to active vB
...
+
+Not all the constraints can be satisfied when the buckets are allocated across  uneven group.
```
dave-finlay: extra space

modules/learn/pages/clusters-and-availability/groups.adoc
```diff
@@ -59,16 +59,35 @@ Note that Group 2 contains all the replica vBuckets that correspond to active vB
...
+
+*rack-zone*:: all replicas of a given vbucket must reside in separate server groups.
+
+*active balance*:: there should be approximately{empty}footnote:approximate-vbuckets[Note that by "approximately the same" we mean the number should vary by 5 vbuckets or less] the same number of active vbuckets on each node in the cluster.
+
+*replica balance*:: there should be approximately{empty}footnote:approximate-vbuckets[] the same number of replica vbuckets on each node in the cluster.
+
+Not all the constraints can be satisfied when the buckets are allocated across  uneven group. 
+In this scenario, the `active balance` and `rack-zone` constraints will take priority: 
+when the vbucket map is generated, the aim is to ensure that there are the same number of active vbuckets on each cluster and that replicas of a given vbucket must reside in separate groups.
```
dave-finlay: suggest "we will" instead of "the aim is"

dave-finlay: approximately the same

modules/learn/pages/clusters-and-availability/groups.adoc
```diff
@@ -59,16 +59,35 @@ Note that Group 2 contains all the replica vBuckets that correspond to active vB
...
+
+Group 2 contains all the replica vBuckets that correspond to active vBuckets in Group 1.
+
+However, in order to ensure that replicas of a VBucket must reside in a separate group, then we may have a situation where there are vBuckets from Server 9 replicated to Group 1, but there are no additional vBuckets in Group 1 to provide balance in Group 2.
 
-Group 1 contains all the replica vBuckets that correspond to active vBuckets on Group 2.
-However, since the groups contain unequal number of nodes, Group 2 not only contains all the replica vBuckets that correspond to active vBuckets on Group 1, but also contains all the replica vBuckets for its own additional node, Server 9 — the replicas for Server 9 being distributed across the other Group 2 nodes; which are Servers 5, 6, 7, and 8.
-Server 9 contains its own active vBuckets, plus replica vBuckets for Group 1.
+NOTE: Since smaller server groups will carry more replicas on each node, their performance may not be the same as the nodes in the bigger server groups. 
+For best performance, try to maintain equality of server groups.
```
dave-finlay: We need to make this recommendation more strongly. Smaller server groups will carry more replica vbuckets, which means there is greater memory pressure on memcached so more fetches go to disk which means higher worst case GET latencies. Additionally more replicas mean more writes to disk and greater compaction burden which will also affect latencies. Customers will notice this as the smaller server groups will "perform" worse than the later server groups. So for reasons of consistency of performance Couchbase strong recommends that customers balance their server groups.

dave-finlay (commented):
initial set of comments left

modules/learn/pages/clusters-and-availability/groups.adoc
RayOffiah: Good spot, though Asciidoc removes double spaces like this when the page is compiled.

RayOffiah:
@dave-finlay <-- Fixes pushed.

RayOffiah:
To make it easier to read, I've put up the page here:

https://rayoffiah.github.io/DOC-9850/server/7.1/learn/clusters-and-availability/groups.html

tonyjhillman (approved):
Line 85 'group' should be plural.

The spelling of vBucket varies, so should probably be made consistent.

Line 104 should probably be 'a greater compaction burden'.

Line 107 'strong' should be 'strongly'.

The statement that customers should 'balance' their server groups I think means that there should be the same number of nodes in each. But they might interpret it as 'rebalance', so I wonder if another phrase would be better.

modules/learn/pages/clusters-and-availability/groups.adoc
```diff
@@ -46,6 +47,18 @@ For information on the standard (non-Group-based) distribution of replica vBucke
 The distribution of vBuckets across groups is exemplified below.
 In each illustration, all servers are assumed to be running the Data Service.
 
+The number of replicas of each bucket in each group depends on the number of server groups in the rack.
```
dave-finlay: It would be good if we could do the following:

- It would be to highlight this as "Important Note" or "Pay Attention Here" -- however we generally do that in the docs.
- We should also tell the users that this is a behavior change from 7.0 and prior releases. 
- We should emphasize that this behavior change only affects situations where the user has configured non-trivial server groups (i.e. more than one server group with that contain KV nodes).

dave-finlay: Instead of the "The number of replicas of each bucket", we should say: "the number of replicas placed in server groups depends on the number of configured server groups." 

May be a good idea to introduce the idea of "non-trivial KV server groups" here. Non-trivial is a terrible name but it's the best I can come up with with little thought. Perhaps Shivani might have a better name. Worth getting a review from her, I would say.

RayOffiah:
Added the details in an admonition

image

@shivaniguptacb <-- How does this look?

modules/learn/pages/clusters-and-availability/groups.adoc
```diff
@@ -46,6 +47,23 @@ For information on the standard (non-Group-based) distribution of replica vBucke
 The distribution of vBuckets across groups is exemplified below.
 In each illustration, all servers are assumed to be running the Data Service.
 
+[IMPORTANT]
+====
+The number of replicas placed in server groups depends on the number of configured server groups..
```
dave-finlay: double period

modules/learn/pages/clusters-and-availability/groups.adoc
```diff
@@ -46,6 +47,23 @@ For information on the standard (non-Group-based) distribution of replica vBucke
...
+If there are at least two server groups, then we can broadly say that:
```
dave-finlay: Suggest: "In 7.1, when there are at least two server groups containing KV nodes, then it is the case that"

modules/learn/pages/clusters-and-availability/groups.adoc
```diff
@@ -46,6 +47,23 @@ For information on the standard (non-Group-based) distribution of replica vBucke
...
+
+[stem]
+++++
+number\:of\:replicas = number\:of\:server\:groups - 1
```
dave-finlay: Suggest "number of replicas placed" instead of just "number of replicas"

modules/learn/pages/clusters-and-availability/groups.adoc
```diff
@@ -46,6 +47,23 @@ For information on the standard (non-Group-based) distribution of replica vBucke
 The distribution of vBuckets across groups is exemplified below.
 In each illustration, all servers are assumed to be running the Data Service.
 
+[IMPORTANT]
```
shivaniguptacb: This whole section 'Important' is  super complicated. I suggest we simplify it to something as follows. Also place it after the discussion on equal and unequal groups (not right at the top of the section):

If you have fewer server groups than the total number of data copies (= 1 + number of replicas), only as many copies will be placed as the number of server groups. Essentially you may not get as many replicas as you set if you have fewer server groups. 
E.g. if you have 2 server groups, and set replicas = 2, you will only have one replica for each vbucket. But if you have 3 server groups you will have 2 replicas for each vbucket.
This behavior is a change from pre-7.1 behavior where we did not reduce the number of data copies to match the number of server groups.

shivaniguptacb (changes requested):

RayOffiah:
https://rayoffiah.github.io/DOC-9850/server/7.1/learn/clusters-and-availability/groups.html

image

modules/learn/pages/clusters-and-availability/groups.adoc
```diff
@@ -59,16 +60,60 @@ Note that Group 2 contains all the replica vBuckets that correspond to active vB
...
+
+Additionally, more replicas mean more writes to disk and a greater compaction burden which will also affect latencies. 
+
+Customers will notice this as the smaller server groups will "perform" worse than the later server groups. 
+So for reasons of consistency of performance Couchbase strongly recommends that customers endeavor to maintain an equal number of nodes across server groups.
+====
+
+For more information on optimizing your cluster configuration, consult the xref:install:sizing-general.adoc[].
+
+.IMPORTANT - Change in behavior from previous versions
```
dave-finlay, line 93: Looks good to me.

modules/learn/pages/clusters-and-availability/groups.adoc
```diff
@@ -1,6 +1,7 @@
 = Server Group Awareness
```
shivaniguptacb, line 1: There is a para here that should be removed as it is no longer true. It is the following one:

"Protection is optimal when groups are assigned equal numbers of nodes, and replica vBuckets and index replicas are thereby distributed such that none ever occupies the same group as its associated active vBucket or index. By contrast, when groups are not assigned equal numbers of nodes, rebalance can only produce a best effort redistribution of replicas: this may result in one or more replica vBuckets and/or index replicas each occupying the same group as their associated active vBuckets or indexes; meaning that vBucket data and/or index data may be lost if such a group becomes unavailable."

shivaniguptacb (changes requested):

modules/learn/pages/clusters-and-availability/groups.adoc
shivaniguptacb: Also the following bullet should be removed:
"Failover should be enabled for server groups only if three or more server groups have been established, and sufficient capacity exists to absorb the load of any failed-over group."

shivaniguptacb (changes requested):

modules/learn/pages/clusters-and-availability/groups.adoc
shivaniguptacb: This bullet also should be removed:
"If a cluster contains only one node, or only one group of nodes, enabling Group Failover has no effect; meaning that failover can only be performed on a per-node basis."

shivaniguptacb: Reword "For information on failing over individual nodes, see [Fail a Node over and Rebalance]" to 
"For information on failover see [Failover and Rebalance]" with same link

shivaniguptacb (changes requested):

shivaniguptacb (approved):
Looks good now.

PR #2575, opened 2022-03-30
DOC-6711: Modified cbbackupmgr compatibility table

tonyjhillman:


jamesl33 (approved):
Presuming I've read the table correctly, LGTM!

Might not be a concern for the moment, but I believe 7.1.0 will only support backing up, back to 6.5.0 (we're dropping support for 6.0.x and 5.x.x).

PR #2573, opened 2022-03-30
Doc 7653

tonyjhillman:

PR #2586, opened 2022-04-04
DOC-7653a: Added review feedback on bloom filters and in memory compr…

tonyjhillman:
…ession


tonyjhillman:
Opened to complete https://github.com/couchbase/docs-server/pull/2573/

tonyjhillman:
This version omits the REST page and the reference to it in the nav.
It uses Plasma page, rather than memory page.

For 'disk I/O' , 'disk read/reads' is substituted.

Efficiency improvement is specified as related to 'mutation processing when the mutations are insert heavy'.

I have removed reference to the duration of the periodicity of compression being shorter than that for ejection.

I have removed the REST page, as internal - I will open a separate ticket to update the UI and do further settings REST modifications. (See https://issues.couchbase.com/browse/DOC-9911.)

PR #2597, opened 2022-04-07
DOC-9843 CB does not enforce total memory usage on node

RayOffiah:
Added calculation for memory allocation.

### DIFF
--- a/modules/learn/pages/buckets-memory-and-storage/memory.adoc
+++ b/modules/learn/pages/buckets-memory-and-storage/memory.adoc
@@ -2,6 +2,7 @@
 :description: Couchbase Server memory-management ensures high performance and scalability.
 :page-aliases: understanding-couchbase:buckets-memory-and-storage/memory,architecture:managed-caching-layer-architecture
 
+:stem: latexmath
 [abstract]
 {description}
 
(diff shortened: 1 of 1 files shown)


tonyjhillman (approved):
I believe this is correct. Let's see what Dave says.

modules/learn/pages/buckets-memory-and-storage/memory.adoc
```diff
@@ -16,8 +17,8 @@ The memory-quota allocation specified for a given service applies to every insta
 For example, if 2048 MB is specified for the xref:services-and-indexes/services/analytics-service.adoc[Analytics Service], and the Analytics Service is run on three of a cluster's five nodes, each of the three instances of the service is duly allocated 2048 MB.
 Note that it is not possible to have different memory allocations across multiple instances of the same service within a single cluster.
 
-By default, Couchbase Server allows 90% of a node's total available memory to be allocated to the server and its services.
-Consequently, if a node's total available memory is 100 GB, any attempt to allocate memory beyond 90 GB produces an error.
+As an _approximate_ guide, Couchbase Server allows stem:[max(80\% \times total\_memory, total\_memory-1)] of a node's total memory to be allocated to the Server and its services. +
+So if a node has 100 GB of memory available, then the server will allocate up 99 GB (100 GB - 1 GB) as this is greater than 80 GB (80% of 100 GB).
```
dave-finlay: I think we should state it a little bit differently. 

I'm thinking maybe should say something like: "As a guideline, Couchbase recommends that no more than 90% of a node's memory (80% on nodes with a small amount of total memory) be allocated to the server and its services. The firm limits on server memory allocation are: max(total_memory - 1, 80% * total_memory) where total_memory is the maximum memory on the node in GiB."

RayOffiah: Done

PR #2609, opened 2022-04-13
Doc 9924a

tonyjhillman:


sitaramv (commented):
line 406-411, 517-520,  533-537, 603-607

These seems odd to see as it is we need to convey better so that propositional documentation.

Example 406-411 Query performs covering UNNEST IndexScan by applying all the predicates using ixf_sched;   Even ORDER BY key uses array index key, it can use index order and pass LIMIT, OFFSET to indexer.

tonyjhillman:
Happy to change these. However, I have no direct experience of this technology, so would be guessing. In the meantime, I can do a version of the rewritten 406-411 that you have put above. Do you want to provide drafts for the other three passages also?

sitaramv (commented):
406-411, Query performs covering UNNEST IndexScan by applying all the predicates using ixf_sched; Even ORDER BY key uses array index key, it can use index order and pass LIMIT, OFFSET to indexer.
517-520, Query performs covering UNNEST IndexScan by applying the predicates on both leveles of ARRAY using ixf_sched_nested
533-537, Query performs covering UNNEST IndexScan by applying the predicates on both leveles of ARRAY using ixf_sched_nested; and Uses Index Agg
regation.
603-607, Query performs covering IndexScan by applying all the predicates using ixf_sched. The query-syntax is intuitive, since the multiple fields within the array have not required complex indexing.

sitaramv (approved):

tonyjhillman:
Thanks.

PR #2644, opened 2022-04-25
DOC-9876 non-root install info needs to include download site for aarch64, new in 7.1

RayOffiah:
Changes made for

https://issues.couchbase.com/browse/DOC-9876

### DIFF
--- a/modules/install/pages/non-root.adoc
+++ b/modules/install/pages/non-root.adoc
@@ -1,6 +1,9 @@
 = Non-Root Install and Upgrade
 :description: pass:q[Couchbase Server can be installed on any supported Linux distribution by users who do not have _root_ or _sudo_ privileges.]
 
+:x86-install-location: https://packages.couchbase.com/cb-non-package-installer/cb-non-package-installer-x86_64
+:aarch64-install-location: https://packages.couchbase.com/cb-non-package-installer/cb-non-package-installer-aarch64
+
 [#abstract]
 {description}
 A non-root server, once installed, can be upgraded by the same users.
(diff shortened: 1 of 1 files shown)


tonyjhillman (approved):

modules/install/pages/non-root.adoc
```diff
@@ -128,32 +131,41 @@ To perform a non-root installation of Couchbase Server on any supported Linux di
 . Using `wget` or `curl`, download the https://packages.couchbase.com/cb-non-package-installer/cb-non-package-installer binary.
 Note that the URI is the same for all supported Linux distributions.
```
hyunjuV: I think that these two lines (131-132) should be removed:
```
. Using `wget` or `curl`, download the https://packages.couchbase.com/cb-non-package-installer/cb-non-package-installer binary.
Note that the URI is the same for all supported Linux distributions.
```
Reason is because the `https://packages.couchbase.com/cb-non-package-installer/cb-non-package-installer` will continue to be the download for "cb-non-package-installer" (which is an x86_64 binary) for backwards compatibility -- since that's what it was before 7.1 -- so, for any customer continuing to use that URI to download the x86_64 cb-non-package-installer binary in 7.1 (like, if they are using it in a script), it will continue to work.   However, since this binary download doesn't have a platform suffix, it may be confusing to new users.  Since we don't need these two lines in the doc, I'm thinking it's best to just leave them out.

modules/install/pages/non-root.adoc
```diff
@@ -128,32 +131,41 @@ To perform a non-root installation of Couchbase Server on any supported Linux di
...
-. Run the `cb-non-package-installer` binary, to install Couchbase Server.
+. Run the `cb-non-package-installer-#<platform-suffix>#` binary, to install Couchbase Server.
 For example:
 +
+[source, console, subs=+quotes]
 ----
- ./cb-non-package-installer --install --install-location ./cb-install \
- --package ./couchbase-server-enterprise-6.6.0-7853-centos7.x86_64.rpm
+ ./cb-non-package-installer-#<platform-suffix># --install --install-location ./cb-install \
+ --package ./couchbase-server-enterprise-6.6.0-7853-centos7.#<platform-suffix>#.rpm
```
hyunjuV: We should update the package name in lines 160-161 to reflect 7.1, and also, since (as far as I know), currently, only amzn2 is supported for both aarch64 and x86_64, use amzn2 that as the example package.

Instead of 
`--package ./couchbase-server-enterprise-6.6.0-7853-centos7.#<platform-suffix>#.rpm`
use
`--package ./couchbase-server-enterprise-7.1.0-amzn2.#<platform-suffix>#.rpm`

modules/install/pages/non-root.adoc
```diff
@@ -250,9 +262,10 @@ To upgrade an existing non-root Couchbase Server,
 Specify the new package to be used for upgrade; and specify the install location of the currently resident Couchbase Server.
 For example:
 +
+[source, console, subs=+quotes]
 ----
-./cb-non-package-installer --upgrade --install-location ./cb-install \
---package ./couchbase-server-enterprise-6.6.0-7854-centos7.x86_64.rpm
+./cb-non-package-installer-#<platform-suffix># --upgrade --install-location ./cb-install \
+--package ./couchbase-server-enterprise-6.6.0-7854-centos7.#<platform-suffix>#.rpm
```
hyunjuV: Since this is part of an upgrade example (line 268), let's use 7.0.0 (instead of 7.1.0).
Instead of 
`--package ./couchbase-server-enterprise-6.6.0-7854-centos7.#<platform-suffix>#.rpm`
Use
`--package ./couchbase-server-enterprise-7.0.0-amzn2.#<platform-suffix>#.rpm`

hyunjuV (changes requested):

PR #2650, opened 2022-04-27
DOC-9074 Docker: clarify that a single-node upgrade isn't possible

RayOffiah:
Added instructions and note about not being able to upgrade a single docker node.

https://issues.couchbase.com/browse/DOC-9074

### DIFF
--- a/modules/install/partials/cannot-upgrade-docker-single-node.adoc
+++ b/modules/install/partials/cannot-upgrade-docker-single-node.adoc
@@ -1,0 +1,21 @@
+////
+This file contains a couple of warning about not being able to upgrade a single node Docker
+container cluster without a second node to balance onto.
+////
+
+// tag::cannot-upgrade-single-docker-node-note[]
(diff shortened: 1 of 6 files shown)


tonyjhillman (approved):

modules/install/partials/cannot-upgrade-docker-single-node.adoc
```diff
@@ -0,0 +1,21 @@
+////
+This file contains a couple of warning about not being able to upgrade a single node Docker
```
markspolakovs, line 2: nit:
```suggestion
This file contains a couple of warnings about not being able to upgrade a single node Docker
```

markspolakovs (approved):
Not 100% sure on the volume reuse aspect but LGTM

markspolakovs:
Discussed this with @malarky - technically it _is_ possible to reuse a volume when launching a new container, but this doesn't help when upgrading because CB won't upgrade the existing data on that volume. So for upgrading a Docker CB node a rebalance is effectively the only way.

PR #2659, opened 2022-05-03
DOC-9719: Removed incorrect list of categories

tonyjhillman:


RayOffiah (commented):
@tonyjhillman 

Hi there. Quick question.

Looking at the original JIRA, I had the impression that the list needed to be corrected based on the list here:

https://docs.couchbase.com/server/current/rest-api/rest-logs-get.html

But the PR looks like it's removing the list entirely?

tonyjhillman:
I think it's fine to omit. The correct list is on the main page.

RayOffiah (approved):

PR #2664, opened 2022-05-03
DOC-9718 Need a docs page that describes all supported browsers across all currently supported Couchbase server versions.

RayOffiah:
https://issues.couchbase.com/browse/DOC-9718

Added column to the supported web browsers section on install-platforms.adoc.


RichardSmedley:
Note, we have a Capella browser support page you may want to link to after the table - https://docs.couchbase.com/cloud/reference/browser-compatibility.html

RayOffiah:
Yep, I'll add it.

PR #2680, opened 2022-05-05
Forum link is dead

jon-strabala:
The formum post https://forums.couchbase.com/t/how-to-add-fts-index-for-geo-search/29742/6 is now a dead link
need to add a page alias


modules/fts/pages/fts-supported-queries-geo-spatial.adoc
```diff
@@ -1,4 +1,5 @@
 = Geospatial Queries
+:page-aliases: fts-geospatial-queries
```
RayOffiah:
```suggestion
:page-aliases: fts-geospatial-queries.adoc
```

OpenDevise now recommend that we start using the full file name when using page aliases.

RayOffiah (changes requested):

RayOffiah (approved):

PR #2687, opened 2022-05-09
DOC-8756 Clarification needed about Data reader/writers and Ephemeral

RayOffiah:


tonyjhillman (approved):
Please don't check in until after Simon has done the build.....thanks.

PR #2697, opened 2022-05-12
DOC-5546 document how to adjust KV tracing threshold

RayOffiah:


modules/manage/pages/manage-logging/manage-logging.adoc
```diff
@@ -444,3 +446,54 @@ The Logging REST API provides the endpoints for retrieving log and diagnostic in
 
 To retrieve log information use the `/diag` and `/sasl_logs`
 xref:rest-api:logs-rest-api.adoc[REST endpoints].
+
+[#adjust-threshold-slow-op-logging]
+== Adjusting Threshold for Logging Slow Operations
+
+It is possible to examine and/or alter the logging threshold for slow-running operations.
+This is done using the `mcctl` command that comes packaged with the Couchbase server installation.
```
ingenthr, line 454: I believe this would need to be done for each node of the cluster, but it would be good to check with primary sources.  If so, it would be good to mention that here.

Also, are the settings persisted?  I honestly don't know, but a user would want to know.

modules/manage/pages/manage-logging/manage-logging.adoc
```diff
@@ -444,3 +446,54 @@ The Logging REST API provides the endpoints for retrieving log and diagnostic in
...
+
+.Set logging threshold example
+[source, bash]
+----
+/opt/couchbase/bin/mcctl --host localhost -u Administrator -P password \
+set sla '{"version":1, "DELETE_BUCKET":{"slow":"100ms"}}'
+----
+
+In this example, the threshold for the `DELETE_BUCKET` operation is being set to 100ms. If a bucket deletion operation takes longer than this, then an message will be logged. 
+It is also possible to set the threshold for all the op-codes in a single command by using the `default` code.
```
ingenthr: What are the valid units and ranges?  `s` and `ms` are demonstrated.  can I set "999 us"?  can I set "1 ms"?

modules/manage/pages/manage-logging/manage-logging.adoc
```diff
@@ -444,3 +446,54 @@ The Logging REST API provides the endpoints for retrieving log and diagnostic in
...
+
+=== Setting the Threshold
+
+The `mcctl` command line interface is also used to set the thresholds for the memcahe operations:
+
+.Set logging threshold example
+[source, bash]
+----
+/opt/couchbase/bin/mcctl --host localhost -u Administrator -P password \
+set sla '{"version":1, "DELETE_BUCKET":{"slow":"100ms"}}'
```
ingenthr: Above there are spaces between the number and the strings, but here there is not a space.  Typo?

ingenthr (changes requested):
I think there might be a typo, thus the request for changes, but have recommended some other things to consider.  

Thanks for getting this documented Ray!

modules/manage/pages/manage-logging/manage-logging.adoc
RayOffiah: Yes, that makes sense. I'll have a chat with someone and confirm

RayOffiah:
Not a typo. The get sla command formats the output with a string between the number and the 'ms'

modules/manage/pages/manage-logging/manage-logging.adoc
RayOffiah: Not a typo. The `get sla` command formats the strings to add in the space.

modules/manage/pages/manage-logging/manage-logging.adoc
RayOffiah: The settings are not persisted between restarts.

RayOffiah: Yes, this does need to be done for each node in the cluster. I'll add a note.

modules/manage/pages/manage-logging/manage-logging.adoc
ingenthr: Up to you if you want to mention this.  Maybe it's not worth it.  The expectation is that you would read-what-you-write and vice versa.  Since this interface doesn't follow that principle, it may or may not be worth mentioning.  Obviously not a fault of the documentation, but I bet someone will get it wrong.  Hopefully it just fails in a reasonable way.

modules/manage/pages/manage-logging/manage-logging.adoc
ingenthr: wowza, nanos are rather pointless, but okay.  :)

ingenthr (approved):
Thanks for taking the feedback @RayOffiah !  LGTM.  Just one comment for you to consider, but I don't think it's necessary that we address the differing formats.

modules/manage/pages/manage-logging/manage-logging.adoc
RayOffiah: I'm going to quickly try doing a set sla with the space. It might worth both ways.

RayOffiah: Hah! Of course it works both ways. The comments in the code didn't indicate it, but it does makes sense to parse with either a space or no space. Right,  I'm going to adjust the text to put spaces in the set examples, then add a tip to the users know that it will also work without spaces as a small convenience.

No need for another review. 

I'm going to put the change in version 7.0 too, since 7.1 has only just been released.

PR #2702, opened 2022-05-12
DOC-9923 Neo - System Resource Requirements, missing ARM

RayOffiah:


modules/install/pages/pre-install.adoc
```diff
@@ -4,6 +4,11 @@
 [abstract]
 {description}
 
+[tabs]
+====
+Intel Processors::
```
b33f: I wouldn't refer to this as Intel Processors..   As we support AMD x86 cpus as well as Intel.  So I would suggest to just stick with x86 Processors

b33f (changes requested):

modules/install/pages/pre-install.adoc
RayOffiah: Yes, good point.

tonyjhillman (approved):

PR #2756, opened 2022-06-15
DOC-9247: 7.1 Search Service file transfer

tonyjhillman:


tonyjhillman:
If this seems ok, I'll do 7.0.2 next.

modules/ROOT/nav.adoc
```diff
@@ -800,6 +800,7 @@ include::cli:partial$cbcli/nav.adoc[]
     *** xref:rest-api:rest-fts-advanced.adoc[Advanced]
     *** xref:rest-api:rest-fts-statistics.adoc[Search Statistics]
     *** xref:rest-api:rest-fts-query.adoc[Active Queries]
+    *** xref:rest-api:rest-fts-partition-file-transfer.adoc[Rebalance with File Transfer]
```
sreekanth-cb: Rebalance with File Transfer => doesn't sound the best wording in my limited knowledge. (not a native speaker)

File Transfer Based Rebalance  Or
Rebalance based on file transfer  Or

may sound better perhaps?

modules/rest-api/pages/rest-fts-partition-file-transfer.adoc
```diff
@@ -0,0 +1,79 @@
+= Rebalance with File Transfer
+:description: The Search-Service REST API allows new partitions to be created by file transfer.
```
sreekanth-cb: The Search-Service REST API helps to configure the file transfer based rebalance.  ?

modules/rest-api/pages/rest-fts-partition-file-transfer.adoc
```diff
@@ -0,0 +1,79 @@
...
+----
+
+[#description]
+== Description
+
+The Search Service automatically partitions its indexes across all Search nodes in the cluster, ensuring optimal distribution, following rebalance.
+
+To achieve this, in versions of Couchbase Server prior to 7.1, by default, partitions needing to be newly created were entirely _built_, on their newly assigned nodes.
+In 7.1+, by default, new partitions are created by the _transfer_ of partition files from old nodes to new nodes: this significantly enhances performance.
+This is an Enterprise-only feature, which requires all Search Service nodes _either_ to be running 7.1 or later; _or_ to be running 7.0.2, with the feature explicitly switched on.
```
sreekanth-cb: We can specify here that - this endpoint would be useful in the above situation where the customer is running 7.0.2.

modules/rest-api/pages/rest-fts-partition-file-transfer.adoc
```diff
@@ -0,0 +1,79 @@
...
+----
+PUT /api/managerOptions
+----
+
+[#description]
+== Description
+
+The Search Service automatically partitions its indexes across all Search nodes in the cluster, ensuring optimal distribution, following rebalance.
+
+To achieve this, in versions of Couchbase Server prior to 7.1, by default, partitions needing to be newly created were entirely _built_, on their newly assigned nodes.
```
sreekanth-cb, line 19: "To achieve this, in versions of Couchbase Server prior to 7.1, by default, partitions needing to be newly created were entirely _built_, on their newly assigned nodes."

do we need this statement here as well in the REST endpoint description?

modules/rest-api/pages/rest-fts-partition-file-transfer.adoc
```diff
@@ -0,0 +1,79 @@
...
+In 7.1+, by default, new partitions are created by the _transfer_ of partition files from old nodes to new nodes: this significantly enhances performance.
```
sreekanth-cb, line 20: Shall we add additional info like -> The users on 7.1+ can use this endpoint if they wish to disable this feature for some reason. (please use the proper English wordings for my proposals)

modules/rest-api/pages/rest-fts-partition-file-transfer.adoc
```diff
@@ -0,0 +1,79 @@
...
+
+----
+curl -X PUT http://<ip-address-or-domain-name>:8094/api/managerOptions
+  -u <username>:<password>
+  -d '{"disableFileTransferRebalance": [ "true" | "false" ]}'
+  -H "Content-type:application/json"
+----
+
+If the value specified for the key `disableFileTransferRebalance` is `false` (which is the default in 7.1+), new Search-Service partitions are created during rebalance by means of partition-file transfer.
+If the value is `true`, partitions are created by means of partition building.
```
sreekanth-cb: by means of partition building => by means of partition building from scratch over DCP connections from KV/Data service.

modules/rest-api/pages/rest-fts-partition-file-transfer.adoc
```diff
@@ -0,0 +1,79 @@
...
+
+== Examples
+
+The following example _disables_ the creation of new partitions by means of file-transfer:
+
+----
+curl -X PUT http://localhost:8094/api/managerOptions \
+-u Administrator:password \
+-d '{"disableFileTransferRebalance": "true" }' \
+-H "Content-type:application/json"
```
sreekanth-cb: -H option perhaps should go before the body-d option. (usual practice). Same everywhere.

sreekanth-cb (commented):
few minor comments ...

modules/ROOT/nav.adoc
tonyjhillman: Ok. I'll use Rebalance Based on File Transfer.

modules/rest-api/pages/rest-fts-partition-file-transfer.adoc
tonyjhillman: I've put: The Search-Service REST API configures rebalance based on file transfer.

modules/rest-api/pages/rest-fts-partition-file-transfer.adoc
tonyjhillman: I've put: _or_ to be running 7.0.2, with the feature explicitly switched on by means of this endpoint.

modules/rest-api/pages/rest-fts-partition-file-transfer.adoc
tonyjhillman: I think it needs to be in the overview for sure: I've tried leaving it out from the REST page, but that makes it harder to understand why it is an Enterprise only feature: we have to contrast with the build option, in order to make that clear. So on balance I would leave it.

modules/rest-api/pages/rest-fts-partition-file-transfer.adoc
tonyjhillman: I've added: Note that users of 7.1+ can explicitly switch the feature _off_ by means of this endpoint; in which case _partition build_ is used to establish new partitions, rather than file transfer.

modules/rest-api/pages/rest-fts-partition-file-transfer.adoc
tonyjhillman: I've put: If the value is `true`, partitions are created by means of partition build, from scratch; over DCP connections from the Data Service.

modules/rest-api/pages/rest-fts-partition-file-transfer.adoc
tonyjhillman: ok

modules/rest-api/pages/rest-fts-partition-file-transfer.adoc
sreekanth-cb: 👍

sreekanth-cb (approved):

PR #2770, opened 2022-06-17
DOC-10139: Discussion of scopes and collections

simon-dew:
Docs issue: [DOC-10139](https://issues.couchbase.com/browse/DOC-10139)

### DIFF
--- a/modules/learn/pages/buckets-memory-and-storage/buckets-memory-and-storage.adoc
+++ b/modules/learn/pages/buckets-memory-and-storage/buckets-memory-and-storage.adoc
@@ -1,5 +1,5 @@
 = Buckets, Memory, and Storage
-:description: pass:q[Couchbase Server saves items in _Buckets_.]
+:description: pass:q[Couchbase Server saves data within _Buckets_.]
 :page-aliases: understanding-couchbase:buckets-memory-and-storage/buckets-memory-and-storage
 
 [abstract]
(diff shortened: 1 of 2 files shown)


simon-dew:
Complete discussion of scopes and collections is not appropriate for the buckets page, as these are described fully in their own page. I mentioned scopes and collections in the abstract, and linked to the scopes and collections page at the end.

krugster (commented):
I wonder if it might be good to take a bit more holistic view of how we are presenting these topics.  There are multiple places for a reader to land on that all discuss buckets, scopes and collections in different ways and with varying levels of detail.  Even with the best cross-linking, a reader is likely to find it very hard to know what they don't know and if they only read one of the multiple pieces of documentation on the topic, they won't be getting a complete picture.

For example:
- https://docs.couchbase.com/server/current/learn/buckets-memory-and-storage/buckets-memory-and-storage.html
- https://docs.couchbase.com/server/current/learn/data/scopes-and-collections.html
- https://docs.couchbase.com/server/current/tutorials/buckets-scopes-and-collections.html#scopes-and-collections
- might be more that I haven't found!


I noticed that there's also no singe place where we explicitly connect the terms of bucket<->database, scope<->schema, collection<->table, document<->row (https://blog.couchbase.com/scopes-and-collections-for-modern-multi-tenant-applications-couchbase-7-0/)

simon-dew:
These are great points @krugster . It's worth noting that for Elixir, we are aiming to overhaul the user documentation to improve the new starter experience, which will necessarily cover the introduction of key concepts like this. As such I'm going to merge this PR as a quick fix, but we'll take your comments on board as we plan the reworked documentation.

The nearest thing we have to a comparison with traditional relational database structures is here: [Logical Hierarchy](https://docs.couchbase.com/server/current/n1ql/n1ql-intro/sysinfo.html#logical-hierarchy). It's hidden away in the Query section, and the terminology needs tightening up.

PR #2773, opened 2022-06-19
Update eventing-Terminologies.adoc

nawazish-couchbase:
Typo `can trigger` repeating twice in the `Eventing Keyspaces` section

PR #2781, opened 2022-06-23
Update eventing-Terminologies.adoc

jon-strabala:
Typo `can trigger` repeating twice in the `Eventing Keyspaces` section


jon-strabala:
Updates based upon 7.0 #2773 and are the same

PR #2783, opened 2022-06-24
DOC-10199: Removed old server nodes intro page

tonyjhillman:


tonyjhillman:
Thank you. This old page was replaced with the page that now groups servers and nodes.

PR #2788, opened 2022-06-27
DOC-9218: [How-to Guides for Server Developers] Cost-Based Optimizer

simon-dew:
The following draft documentation is ready for review:

[Developer Guides › Optimizing Performance](https://simon-dew.github.io/docs-site/DOC-9218/server/current/guides/optimize.html)
[Developer Guides › Optimizing Performance › Cost-Based Optimizer](https://simon-dew.github.io/docs-site/DOC-9218/server/current/guides/cbo.html)

Docs issue: [DOC-9218](https://issues.couchbase.com/browse/DOC-9218)

PR #2806, opened 2022-07-05
NO ISSUE: Add alt text to syntax diagrams

simon-dew:
Following on from [discussion](https://github.com/couchbase/docs-server/pull/2788#discussion_r911742406) in #2788

### DIFF
--- a/modules/n1ql/pages/n1ql-language-reference/adaptive-indexing.adoc
+++ b/modules/n1ql/pages/n1ql-language-reference/adaptive-indexing.adoc
@@ -46,7 +46,7 @@ Refer to the {createindex}[CREATE INDEX] statement for details of the syntax.
 include::partial$grammar/ddl.ebnf[tag=simple-array-expr]
 ----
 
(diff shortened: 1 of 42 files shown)


simon-dew:
Assigning this to you since you raised it @osfameron 😁 

My only question is whether the alt text is too long. Might get irritating if this is repeated several times on a page. Would it be adequate just to say "syntax diagram"? Or maybe full alt text for the first instance on the page, abbreviated afterwards?

osfameron (approved):
This looks good.
I'm not an accessibility expert, but https://abilitynet.org.uk/news-blogs/five-golden-rules-compliant-alt-text puts that as Type 2 image (informational) and a description is fine, given that the information is already presented in an alternative way (the EBNF listing).

Sounds to me like the shorter text suggested would be fine if you prefer it, but this is good too 👍

simon-dew:
Thanks. I think I'll use the longer alt text as it is.

PR #2817, opened 2022-07-11
NO ISSUE: Additional release notes for 6.6.x

simon-dew:
* MB-48179: SyncDeletes do not update maxDelRevSeqno
* MB-48713: Rev id going backwards for SyncWrite add

### DIFF
--- a/modules/release-notes/pages/relnotes.adoc
+++ b/modules/release-notes/pages/relnotes.adoc
@@ -50,6 +50,9 @@ This section highlights the notable issues fixed in this release.
 | https://issues.couchbase.com/browse/MB-34280[MB-34280^]
 | *Summary*: Memcached should disallow DCP connections with names greater than 200 characters
 
+| https://issues.couchbase.com/browse/MB-48713[MB-48713^]
+| *Summary*: [Ephemeral] Rev id going backwards for SyncWrite add
+
 | https://issues.couchbase.com/browse/MB-49321[MB-49321^]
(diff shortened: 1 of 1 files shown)


simon-dew:
From comment on Slack. Two things to note:

1. The titles of fixed issues are handled inconsistently between patch versions, and I have not tried to fix this. I have tried to make sure the new fixed issues are consistent within the patch version in which they are placed.

2. I have added the new fixed issue for 6.6.4 _after_ the Analytics Service fixed issues, out of the usual canonical order of Services, because the Analytics Service in 6.6.4 contains a fix for a severe vulnerability.

simon-dew:
Oh one more thing, MB-48713 is also mentioned on the release notes for 7.0.x; I've left it unchanged there, because it was added as a patch fix for 7.0. But MB-48179 was also added as a fix in 7.0.2 and is not mentioned on the release notes for 7.0.x. Looks like we need a separate pull request for this.

tonyjhillman (approved):

daverigby (approved):
Thanks Simon

daverigby:
> Oh one more thing, MB-48713 is also mentioned on the release notes for 7.0.x; I've left it unchanged there, because it was added as a patch fix for 7.0. But MB-48179 was also added as a fix in 7.0.2 and is not mentioned on the release notes for 7.0.x. Looks like we need a separate pull request for this.

Good spot!

Agreed, we should have MB-48179 included in the 7.0.2 release notes.

PR #2822, opened 2022-07-19
DOC-10244: Added info on uploading intermediate certs

tonyjhillman:


modules/learn/pages/security/using-multiple-cas.adoc
```diff
@@ -86,3 +86,19 @@ For example, assume the existence of the following:
...
+a node certificate references its CA by means of a _concatenation_ of all certificates -- node and intermediates -- whose signature-chain leads to the CA.
+The result of this concatenation, the file `chain.pem`, is presented by Couchbase Server to each client that wishes to trace and determine the node's authority.
+
+This form of concatenation continues to be supported in 7.1.
+Alternatively, however, the intermediate certificate (or certificates) can be _uploaded_ to the cluster's trust store, and the node certificate left in _unconcatenated_ form.
+When client-access is attempted, Couchbase Server duly concatenates the node certificate and intermediate certificates, and presents the resulting file to the client.
+
+The upload of intermediate certificates potentially increases the efficiency with which certificate deployment is performed; since it allows node certificates to be uploaded without manual concatenation being needed for each.
```
timofey-barmin: > The upload of intermediate certificates potentially increases the efficiency with which certificate deployment is performed; since it allows node certificates to be uploaded without manual concatenation being needed for each.

Avoiding concatenation by itself is not the goal here (I suggest we remove that as it's too small to be mentioned).
The main reason for adding intermediate certs to the trust store: is to make it possible to skip the intermediate cert in node's chain.pem and/or in client certificates (when client cert auth is enabled).
In the client authentication case (when server verifies the client): Basically clients will not have to provide the intermediate cert for authentication in this case (they can but they don't have to).
In the server identity verification case (when client verifies the server): The server doesn't have to send that intermediate cert to the client for verification then (doesn't have to have it in chain.pem).

modules/learn/pages/security/using-multiple-cas.adoc
```diff
@@ -86,3 +86,19 @@ For example, assume the existence of the following:
...
+
+In versions of Couchbase Server prior to 7.1, as described in xref:learn:security/certificates.adoc#intermediate-certificates[Intermediate Certificates] and xref:learn:security/certificates.adoc#node-certificate[Node Certificates] (and as exemplified in xref:manage:manage-security/configure-server-certificates.adoc[Configure Server Certificates]),
+a node certificate references its CA by means of a _concatenation_ of all certificates -- node and intermediates -- whose signature-chain leads to the CA.
+The result of this concatenation, the file `chain.pem`, is presented by Couchbase Server to each client that wishes to trace and determine the node's authority.
+
+This form of concatenation continues to be supported in 7.1.
+Alternatively, however, the intermediate certificate (or certificates) can be _uploaded_ to the cluster's trust store, and the node certificate left in _unconcatenated_ form.
+When client-access is attempted, Couchbase Server duly concatenates the node certificate and intermediate certificates, and presents the resulting file to the client.
```
timofey-barmin: > When client-access is attempted, Couchbase Server duly concatenates the node certificate and intermediate certificates, and presents the resulting file to the client.

We probably do that but we actually don't have to do it (and maybe it would be better if we don't do it).
I suggest we remove this sentence.
Server's work here is to send the chain (whatever is in chain.pem) to the client. Client should have the rest of the chain (client's trusted certs).

The idea is actually simple (maybe we are overcomplicating it in the doc):
Whatever is in "the trust store" can be used for verification (be it verification of client cert or verification of server's cert when that node is connecting somewhere). 
Whatever is in chain.pem is sent to the client for server identity verification.

modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -368,15 +370,28 @@ Note that access by means of a client certificate must be specifically enabled,
 == Cluster Protection with Root, Intermediate, and Node Certificates
 
 Optionally, a root certificate can be used to sign an _intermediate_ certificate, which is then itself used to sign node certificates.
-This increases security, since it minimizes use of the private key associated with the root certificate, when many node or client-certificates are to be signed.
+This increases security, since it minimizes use of the private key associated with the root certificate, when many node or client certificates are to be signed.
+
+When a client attempts to access the cluster securely, it must be able to inspect both the certificate of the contacted node, and any intermediate certificates used to convey the authority of the cluster's root CA (see xref:learn:security/certificates.adoc#intermediate-certificates[Intermediate Certificates]).
```
timofey-barmin: > it must be able to inspect both the certificate of the contacted node

it must be able to build the certificate chain (chain of trust).

modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -368,15 +370,28 @@ Note that access by means of a client certificate must be specifically enabled,
...
+In Couchbase Server Version 7.1+, these certificates can be prepared for client inspection in either of two ways, which are:
+
+* Concatenation of all intermediate and node certificates into a single _chain.pem_ file, which is uploaded from the node's _inbox_.
+
+* Upload, from the node's inbox, of a _chain.pem_ file containing only the unconcatenated node certificate; and addition of each intermediate certificate to the cluster's _trust store_.
```
timofey-barmin: > Upload, from the node's inbox, of a _chain.pem_ file containing only the unconcatenated node certificate; and addition of each intermediate certificate to the cluster's _trust store_.

It's not correct.
In case when a client (like an app) connects to the couchbase-server (I think that's the case being described here), server's trusted store is not used for verification of server identity.
Client has its own trust store for that.
In other words either server provides the intermediate certs (chain.pem), or client takes them from its own trust store, but server's trust store has nothing to do with it.

Server's trust store is used in two cases:
1) when client cert is enabled: to verify client's certificates
2) when one node connects to another node: the first node verifies identity of another node then (basically the first node is a "client" in this case).

modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -368,15 +370,28 @@ Note that access by means of a client certificate must be specifically enabled,
...
+Both procedures are shown below.
+For more information, see xref:learn:security/using-multiple-cas.adoc#adding-intermediate-certificates-to-the-trust-store[Adding Intermediate Certificates to the Trust Store].
+
+The steps and descriptions in the procedures below assume that the previous procedure, xref:manage:manage-security/configure-server-certificates.adoc#root-and-node-certificates[Cluster Protection with Root and Node Certificates], has already been successfully completed; and that familiarity with the basic certificate-related concepts explained there has been attained.
 
 Note that corresponding, subsequent procedures that create certificates for _client_-authentication are provided in xref:manage:manage-security/configure-client-certificates.adoc#client-certificate-authorized-by-an-intermediate-certificate[Client Access: Intermediate Certificate Authorization] and xref:manage:manage-security/configure-client-certificates.adoc#java-client-access-intermediate-certificate-authorization[Java Client Access: Intermediate Certificate Authorization]
 
+[#intermediate-concatenation]
+=== Deploying an Intermediate Certificate via Concatenation
```
timofey-barmin: > Deploying an Intermediate Certificate via Concatenation

I don't like the word Concatenation here. Rather something like "as part of node's chain"

modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -604,23 +618,82 @@ sudo cp ca.pem /opt/couchbase/var/lib/couchbase/inbox/CA/.
 . Upload the root certificate, thereby activating it for the entire cluster.
 +
 ----
-curl -X POST http://10.144.220.103:8091/node/controller/loadTrustedCAs -u Administrator:password
+curl -X POST http://10.143.192.102:8091/node/controller/loadTrustedCAs -u Administrator:password
+----
+
+. Upload the node certificate, specifying the established password for the private key.
```
timofey-barmin, line 624: Are we providing instructions for generating **encrypted** node private key here? My understanding is that we are not.
Probably we should not use password here is well but can have a link to that page for reference though.

modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -604,23 +618,82 @@ sudo cp ca.pem /opt/couchbase/var/lib/couchbase/inbox/CA/.
...
++
+Note that when, as is typical, the cluster contains more than one node, the `/node/controller/reloadCertificate` command must be executed on each node, specifying the IP address of the node on which execution is occurring.
++
+This concludes the certificate-deployment process.
+The root certificate can be examined by means of Couchbase Web Console, as shown in xref:manage:manage-security/configure-server-certificates.adoc#see-root-certificate-with-couchbase-web-console[Step 13] of the previous example on this page.
+
+For more information using the REST API to manage certificates, see xref:rest-api:rest-certificate-management.adoc[Certificate Management API].
+
+[#intermediate-upload]
+=== Deploying an Intermediate Certificate via Upload
```
timofey-barmin: > Deploying an Intermediate Certificate via Upload

Rather something like "via Trusted store"

modules/rest-api/pages/load-trusted-cas.adoc
```diff
@@ -15,9 +15,11 @@ POST /node/controller/loadTrustedCAs
 [#description]
 == Description
 
-Loads trusted CA (or _root_) certificates that reside on the accessed node into the Couchbase-Server cluster.
-All loaded CA certificates can be accessed by all nodes.
-Loaded certificates can be used to provide authority to the cluster's nodes, and can be used to authenticate clients' access-attempts.
+Loads trusted certificates that reside on the accessed node into the Couchbase-Server cluster.
```
timofey-barmin: Maybe "in the trust store of the couchbase-server" ?

modules/rest-api/pages/load-trusted-cas.adoc
```diff
@@ -15,9 +15,11 @@ POST /node/controller/loadTrustedCAs
...
+All loaded certificates can be accessed by all nodes.
+Loaded CA (or _root_) certificates can be used to provide authority to the cluster's nodes, and can be used to authenticate clients' access-attempts.
+Loaded _intermediate_ certificates are provided to clients when secure client-access is attempted, so that the node's chain of authority can be inspected.
```
timofey-barmin: > Loaded _intermediate_ certificates are provided to clients when secure client-access is attempted, so that the node's chain of authority can be inspected.

That's not the goal, but rather a side effect here. We should not mention it.

timofey-barmin (commented):
Please take a look a my comments.

modules/manage/pages/manage-security/configure-server-certificates.adoc
tonyjhillman: Deploying an Intermediate Certificate via Trust Store

modules/rest-api/pages/load-trusted-cas.adoc
tonyjhillman: I've removed.

modules/rest-api/pages/load-trusted-cas.adoc
tonyjhillman: I've written: Loads trusted certificates into the Couchbase-Server trust store.

modules/manage/pages/manage-security/configure-server-certificates.adoc
tonyjhillman: I've written: Deploying an Intermediate Certificate as Part of the Node's Chain

modules/manage/pages/manage-security/configure-server-certificates.adoc
tonyjhillman: I've rewritten to : Upload, from the node's inbox, of a _chain.pem_ file containing only the unconcatenated node certificate; with the expectation that each intermediate certificate resides in the client's _trust store_.

modules/manage/pages/manage-security/configure-server-certificates.adoc
tonyjhillman: Ok, I've rewritten correspondingly.

modules/learn/pages/security/using-multiple-cas.adoc
tonyjhillman: I have rewritten. Is the following correct?

* When the node's chain contains one or more intermediate certificates, and the node presents only its node certificate (as the file _chain.pem_) to the client; if the client can locate those intermediate certificates and their corresponding CA in the client's trust store, authentication of the node can proceed.

* When the client's chain contains one or more intermediate certificates, and the client presents only its client certificate to the node; if Couchbase Server can locate those intermediate certificates and their corresponding CA in the Couchbase-Server trust store, authentication of the client can proceed.

Also supported are cases where _some_ but not _all_ of a chain's intermediates are directly presented, and the rest are available in the trust store of the recipient.

modules/manage/pages/manage-security/configure-server-certificates.adoc
tonyjhillman: Thanks. I have removed the '-d' and put in the following link.

=== Using an Encrypted, Private Node-Key

The above examples use an _unencrypted_ private key, for the node.
If an _encrypted_ private node-key is used, a passphrase must be registered for it, so that the key can be securely retrieved and used when required.
See the reference page xref:rest-api/upload-retrieve-node-cert.adoc[Upload and Retrieve a Node Certificate].

modules/learn/pages/security/using-multiple-cas.adoc
tonyjhillman: I've removed and rewritten, as below.

timofey-barmin (approved):
LGTM

PR #2819, opened 2022-07-11
DOC-10221: N1QL tutorial query context syntax incorrect

simon-dew:
The following draft documentation is ready for review:

* [Try a Query](https://docs.couchbase.com/server/current/getting-started/try-a-query.html#run-cbq)
* [CREATE COLLECTION](https://docs.couchbase.com/server/current/n1ql/n1ql-language-reference/createcollection.html#query-context)  
* [DROP COLLECTION](https://docs.couchbase.com/server/current/n1ql/n1ql-language-reference/dropcollection.html#query-context)
* [CREATE INDEX](https://docs.couchbase.com/server/current/n1ql/n1ql-language-reference/createindex.html#query-context)
* [CREATE PRIMARY INDEX](https://docs.couchbase.com/server/current/n1ql/n1ql-language-reference/createprimaryindex.html#query-context)
* [DROP INDEX](https://docs.couchbase.com/server/current/n1ql/n1ql-language-reference/dropindex.html#query-context)

Removed quotes from query context throughout.

Docs issue: [DOC-10221](https://issues.couchbase.com/browse/DOC-10221)

PR #2827, opened 2022-07-19
DOC-10221: N1QL tutorial query context syntax incorrect

simon-dew:
Backport of #2819 to release/7.0

### DIFF
--- a/modules/getting-started/pages/try-a-query.adoc
+++ b/modules/getting-started/pages/try-a-query.adoc
@@ -202,7 +202,7 @@ For example, the following cbq command sets the query context to `travel-sample.
 
 [source,n1ql]
 ----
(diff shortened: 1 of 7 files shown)


RayOffiah (approved):
Maybe they were supposed to be backticks.

simon-dew:
I'm sure that these code samples were tested — but it would have been by hand, unfortunately, so there's no record of a passing test. I _think_ that maybe the tester 👀 was using an older, preinstalled version of cbq with a newer version of Couchbase Server? But I haven't had time to check this hypothesis.

What I do know is, (a) the code works without single quotes in cbq version 2.0 , and (b) cbq version 2.0 is the version that's installed with Couchbase Server 7.0 and later, so that's good enough for me.

PR #2830, opened 2022-07-20
DOC-9193for6.6: Added note on CE users limit

tonyjhillman:


RayOffiah (approved):
LGTM 👍🏾

If you have a block of text that is repeated in a number of places, then you could put it in an include file, or create an attribute for it.

PR #2870, opened 2022-08-16
DOC-10317 ARM v8 now supported on Ubuntu 20.04

RayOffiah:
Added notes for https://issues.couchbase.com/browse/CBD-4734


tonyjhillman:
I've transferred these onto DOC-10333 (which is now merged) and DOC-10314 (which is still in review, with the other release note info). Thanks!

PR #2899, opened 2022-09-07
DOC-10286 - Changed incorrect param name

sarahlwelton:
"num_replica" parameter was incorrectly written as "num_replicas" in the createindex.html page. 

The single occurrence has been corrected.


simon-dew (approved):
LGTM 👍

PR #2907, opened 2022-09-13
DOC-10051 - Remove semicolons from query/service requests

sarahlwelton:
Requests against `query/service` can't contain semi-colons without causing errors. 

I went through all of the files listed in the txt file on [DOC-10051](https://issues.couchbase.com/browse/DOC-10051) and removed semi-colons from the appropriate requests. 

Fix version in the ticket was listed as 7.1.1, so raising this against the 7.1 branch. 

Requesting a review from Sitaram as the reporter of the original ticket :)

### DIFF
--- a/modules/n1ql/pages/n1ql-language-reference/curl.adoc
+++ b/modules/n1ql/pages/n1ql-language-reference/curl.adoc
@@ -519,7 +519,7 @@ FROM CURL("http://localhost:8093/query/service",
     "user": "Administrator:password"}).results[*].t route
 JOIN `travel-sample`.inventory.airline
 ON KEYS route.airlineid
-LIMIT 4
+LIMIT 4;
 ----
 
 .Results
--- a/modules/n1ql/pages/n1ql-language-reference/curl.adoc
+++ b/modules/n1ql/pages/n1ql-language-reference/curl.adoc
@@ -519,7 +519,7 @@ FROM CURL("http://localhost:8093/query/service",
     "user": "Administrator:password"}).results[*].t route
 JOIN `travel-sample`.inventory.airline
 ON KEYS route.airlineid
-LIMIT 4;
+LIMIT 4
 ----
 
 .Results
--- a/modules/n1ql/pages/n1ql-language-reference/n1ql-error-codes.adoc
+++ b/modules/n1ql/pages/n1ql-language-reference/n1ql-error-codes.adoc
@@ -220,6 +220,8 @@ These errors are related to the service.
 | `Error processing [feature]`
 | Error processing that feature.
 
+For example, if you specify request parameters as form data and include an unescaped semicolon (;) in a statement, the Query Service REST API returns a `1040` error.
+
 | `1050`
 | `No [feature] value`
 | Missing value for that feature.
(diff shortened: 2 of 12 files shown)


sitaramv:
Examples are fixed. Can we say that they can't use semicolon non application json header

sarahlwelton:
> Examples are fixed. Can we say that they can't use semicolon non application json header

Sorry, I'll need more information about where the best spot in the docs is to put this!

sitaramv:
https://docs.couchbase.com/server/current/n1ql/n1ql-rest-api/examplesrest.html

Request Content Type

For POST requests, the parameters can be specified in the request body, using one of two content types:
       URL-encoded format (Content-type: application/x-www-form-urlencoded)
       JSON format (Content-type: application/json).
URL-encoded is the default.

For GET requests, the parameters are specified in the request URL in URL-encoded format. Clients should ensure that request parameters are URL-encoded when using the URL-encoded content type. URL-encoded format unescaped semicolon (;) is not valid.

simon-dew:
@sitaramv &mdash; please have a look at the related pull request [couchbaselabs/cb-swagger#82](https://github.com/couchbaselabs/cb-swagger/pull/82). Does this cover your comment?

sitaramv:
https://github.com/couchbaselabs/cb-swagger/pull/82  covers it

sitaramv (approved):

modules/n1ql/pages/n1ql-language-reference/curl.adoc
```diff
@@ -519,7 +519,7 @@ FROM CURL("http://localhost:8093/query/service",
     "user": "Administrator:password"}).results[*].t route
 JOIN `travel-sample`.inventory.airline
 ON KEYS route.airlineid
-LIMIT 4;
+LIMIT 4
```
simon-dew: This semicolon doesn't need to go, because it's part of the outer query, not part of the query within the CURL() function

It doesn't actually matter if you do remove it though; I'd leave it in, for consistency with the other examples in this page

modules/n1ql/pages/n1ql-language-reference/n1ql-error-codes.adoc
```diff
@@ -220,6 +220,8 @@ These errors are related to the service.
 | `Error processing [feature]`
 | Error processing that feature.
 
+For example, if you specify request parameters as form data and include an unescaped semicolon (;) in a statement, the Query Service REST API returns a `1040` error.
```
simon-dew: Is it worth rephrasing to emphasize that this example applies to the REST API?

```suggestion
For example, the Query Service REST API returns this error if you specify request parameters as form data and include an unescaped semicolon (;) in a statement.
```

simon-dew (changes requested):
Thanks @sarahlwelton. Couple of minor suggestions.

modules/n1ql/pages/n1ql-language-reference/curl.adoc
sarahlwelton:
```suggestion
LIMIT 4;
```

sarahlwelton: Did this as a suggestion commit.

simon-dew (approved):
LGTM 👍

PR #2905, opened 2022-09-12
DOC-10339 release/7.0: fixes numerical value '9001' to '8091'

franrob-projects:
Doc 10339 release/7.0: fixes numerical value '9001' to '8091'

PR #2909, opened 2022-09-16
DOC-10439 release/7.0: Fixes missing graphic

franrob-projects:
DOC-10439 release/7.0: Fixes missing graphic. 

Reverted back to:

image::data/transaction-mechanics-steps.png["Transaction mechanics explaining the high-level steps that a transaction follows"]

note: Enrico Pagayucan is not searchable as a reviewer


simon-dew (approved):
LGTM :+1:

Just to note that the change to `configure-saslauthd.adoc` should not be part of this pull request! This was fixed by #2905. Not a big deal in this case, but it's important to make sure that each pull request (a) starts from the current revision of the upstream branch, and (b) only includes the changes which are necessary for that particular pull request.

PR #2916, opened 2022-09-21
DOC-10240 - Added information on 5500 error code to N1QL Error Codes reference

sarahlwelton:
Added a new table row for 55xx error codes and added the 5500 error code, with links to existing information about setting memory quotas for understanding what would have caused the error.


simon-dew (approved):
👍  LGTM

PR #2918, opened 2022-09-23
Update convertAdvXMLtoJSON to handle more cases

jon-strabala:
This will now process ERSI Geospatial openstreetmap data (done for a POC with Ron Graham) for ERSI  This will now process Huawei OM data (done for a POC with Daniel Cadenas) for Ericsson takes 45 seconds compared to 45 minutes in Oracle.


RayOffiah:
Is this correct?

It doesn't appear to be a complete function?

RayOffiah (changes requested):
Could you have another look at this and confirm the function is complete?

PR #2922, opened 2022-09-27
DOC-10380: Updated cert info regarding node to node

tonyjhillman:


mohammed-madi (approved):
Thanks for fixing, Approved

PR #2926, opened 2022-09-29
DOC-9536: Modified description of version compatibiity

tonyjhillman:


tonyjhillman:
Question: Regarding the asterisked 6.6.0 entries, are these indeed (as it says) for local backup only, or are they for local 'backup and restore' only? Thanks.

jamesl33 (approved):
LGTM :+1:

jamesl33:
> Question: Regarding the asterisked 6.6.0 entries, are these indeed (as it says) for local backup only, or are they for local 'backup and restore' only? Thanks.

Good question, in theory, the limitation only applies to the backup side (a requirement on the DCP sync-backfill control flag). So I'm quite happy for this to stay as is, a cloud restore should work as expected to 6.5.x, 6.0.x and 5.5.x.

@hyunjuV please let me know if you disagree.

EDIT: Now that I think about it, the same restriction (asterisk) should apply for 6.6.1, 7.0 and 7.1.

tonyjhillman:
Can you verify that I've gotten this right? I've added asterisks to the other versions you've mentioned. Thanks.

jamesl33:
Looks good to me, I think the "6.6.0" and "6.6.1 and above" are now the same though :+1:

tonyjhillman:
I've removed the redundant line from the table. Thanks.

PR #2946, opened 2022-10-06
DOC-10482 n1ql feature controller

sarahlwelton:
...description in general-settings.adoc to specify what the setting....

...does in a general sense and emphasize that the user shouldn't...

...change the setting without technical support guidance.

I decided on an admonition here because it's just an <ul> element - and if we're elaborating, at least generally, on what the feature does, we should be a little more strong on not modifying it without assistance.


simon-dew (approved):
LGTM 👍

simon-dew:
Let's backport this one to release/7.1 also.

PR #2948, opened 2022-10-07
DOC-10455 N1QL to SQL++ Rebrand for Tutorial.

RayOffiah:
SQL++ rebrand for tutorial
Fixed tutorial code location, so it defaults to the latest version.

### DIFF
--- a/modules/tutorials/pages/java-tutorial/adding-course-enrollments.adoc
+++ b/modules/tutorials/pages/java-tutorial/adding-course-enrollments.adoc
@@ -70,7 +70,7 @@ include::{java-sample-location}AddEnrollments.java[]
 . Retrieve the `art history` course record.
 --
 +
-Each method uses a N1QL call to retrieve a single record from its respective collection.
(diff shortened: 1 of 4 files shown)


modules/tutorials/pages/java-tutorial/adding-course-enrollments.adoc
```diff
@@ -70,7 +70,7 @@ include::{java-sample-location}AddEnrollments.java[]
 . Retrieve the `art history` course record.
 --
 +
-Each method uses a N1QL call to retrieve a single record from its respective collection.
+Each method uses an {sqlpp} call to retrieve a single record from its respective collection.
```
simon-dew: Do we pronounce it "sequel plus plus" or "es queue ell plus plus"? If we make a style rule that it's the former, then we can leave the indefinite article unchanged.

```suggestion
Each method uses a {sqlpp} call to retrieve a single record from its respective collection.
```

simon-dew (approved):
LGTM 👍 , one comment but I've approved anyway

PR #2964, opened 2022-10-17
DOC-10486: N1QL to SQL++ Rebrand for Introductory Modules

simon-dew:
* Replaces `N1QL` with `{sqlpp}` attribute in running text
* Replaces `n1ql` with `sqlpp` in source code blocks
* Removes link text containing `n1ql` from xrefs so link text generated automatically

Docs issue: [DOC-10486](https://issues.couchbase.com/browse/DOC-10486)


simon-dew:
Granted, it's a bit pointless updating the release notes / what's new for 7.2 as they will ultimately be rewritten anyway

simon-dew:
As part of this update, I removed explicit link text from almost the whole `n1ql` section of the nav &mdash; we've been trending that way for a while, and I thought I would see it through to its logical conclusion. I didn't want to remove explicit link text from entire modules that are not my responsibility, but I have removed explicit link text for topics whose title previously contained the term `N1QL`.

RayOffiah (approved):
LGTM 👍🏾

PR #2970, opened 2022-10-18
DOC-10449 release/7.2: Changes N1QL to SQL++ for the FTS module in the docs server repo

franrob-projects:
Changes N1QL to SQL++ for the FTS module in the docs-server repo.

**Does** Change N1QL to {sqlpp} in running text
**Does** Change [source,n1ql] to [source,sqlpp]

**Do not** change n1ql in file names, directory names, extensions, links, or anchors
**Do not** change n1ql in commands, options, UI text, or output text

### DIFF
--- a/modules/fts/pages/fts-analyzers-search-functions.adoc
+++ b/modules/fts/pages/fts-analyzers-search-functions.adoc
@@ -1,15 +1,15 @@
 = Analyzers - Search Functions
 
-xref:n1ql:n1ql-language-reference/searchfun.adoc[Search functions] allow users to execute full text search requests within a N1QL query.
+xref:n1ql:n1ql-language-reference/searchfun.adoc[Search functions] allow users to execute full text search requests within a {sqlpp} query.
 
-In the context of N1QL queries, a full text search index can be described as one of the following :
+In the context of {sqlpp} queries, a full text search index can be described as one of the following :
 
 * xref:n1ql:n1ql-language-reference/covering-indexes.adoc[Covering index]
 
 * Non-covering index
 
 This characterization depends on the extent to which it could answer all aspects of the SELECT predicate and the WHERE clauses of a N1Q1 query.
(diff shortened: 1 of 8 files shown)


modules/fts/pages/fts-analyzers-search-functions.adoc
```diff
@@ -1,15 +1,15 @@
...
+xref:n1ql:n1ql-language-reference/searchfun.adoc[Search functions] allow users to execute full text search requests within a {sqlpp} query.
 
-In the context of N1QL queries, a full text search index can be described as one of the following :
+In the context of {sqlpp} queries, a full text search index can be described as one of the following :
 
 * xref:n1ql:n1ql-language-reference/covering-indexes.adoc[Covering index]
 
 * Non-covering index
 
 This characterization depends on the extent to which it could answer all aspects of the SELECT predicate and the WHERE clauses of a N1Q1 query.
```
simon-dew: Missed one!

```suggestion
This characterization depends on the extent to which it could answer all aspects of the SELECT predicate and the WHERE clauses of a {sqlpp} query.
```

modules/fts/pages/fts-analyzers-search-functions.adoc
```diff
@@ -1,15 +1,15 @@
...
-An N1QL query against a non-covering index will go through a "Verification phase.” In this phase  documents are fetched from the query service based on the results of the search index, and the documents are validated as per the clauses defined in the query.
+An {sqlpp} query against a non-covering index will go through a "Verification phase.” In this phase  documents are fetched from the query service based on the results of the search index, and the documents are validated as per the clauses defined in the query.
```
simon-dew: We're going with "a SQL++ query" rather than "an SQL++ query". (I'm not sure why this had "an" in the first place.) Also fixing mismatched straight and curly quotes.

```suggestion
A {sqlpp} query against a non-covering index will go through a "Verification phase". In this phase documents are fetched from the query service based on the results of the search index, and the documents are validated as per the clauses defined in the query.
```

simon-dew (changes requested):
Just a couple of changes, thanks @franrob-projects

PR #2990, opened 2022-10-25
DOC-10553: Added new tasks to rest cluster-tasks page

tonyjhillman:


tonyjhillman:
Warmup, collect, and orphan are not yet covered here. I have tried to get each of the others.

PR #3004, opened 2022-11-14
 DOC-10609 release/7.1.2: Changes to {sqlpp} instead of N1QL in running text

franrob-projects:
Changes to {sqlpp} instead of N1QL in running text. But leaves N1QL in the output of the VERSION() function.


simon-dew (commented):
First of all, my mistake entirely — the N1QL to `{sqlpp}` replacement only applies to release/7.2 and later, not to release/7.1. Sorry about that!

Secondly, I’m slightly confused, because the issue is to document the `VERSION()` and `DS_VERSION()` functions?

PR #3007, opened 2022-11-16
DOC-10615: Release notes for 7.1.3

tonyjhillman:


westmann (commented):
We have a "Security" section that covers 2 dependencies that contained vulnerabilities and we also have a JDK update in the "Analytics Service" section that is also addressing vulnerabilities. I'm wondering if we should just move the JDK to the "Security" section and remove the "Analytics Service" section.

tonyjhillman:
At Ian McCloy's request, the security section has been removed.

tonyjhillman:
At Ian's further request, I've removed the analytics issue also. Hope that's okay...

PR #3024, opened 2022-12-14
Correct bucket create eviction parameter

BenHuddleston:
It should be "fullEviction" rather than "full". The following error message is given when creating a bucket with "full" evictionPolicy:

"evictionPolicy":"Eviction policy must be either 'valueOnly' or 'fullEviction' for couchbase buckets"

The rest of the page appears to correctly list the parameter as "fullEviction".

Unsure if this is the correct branch.


osfameron (approved):
Looks good to me 👍

PR #3069, opened 2023-02-06
DOC-10800: Migrate how-to guides to docs-devex

simon-dew:
Removes the how-to guides from the `docs-server` repo for `release/7.2` branch. See also:

* [couchbase/docs-site#655](https://github.com/couchbase/docs-site/pull/655)
* [couchbaselabs/docs-devex#5](https://github.com/couchbaselabs/docs-devex/pull/5)

Docs issue: DOC-10800


sarahlwelton (approved):
Makes sense to me, just removing the duplication of the content. I'm sure we can streamline this process to cut down on the PRs in the future, but since we're still figuring out how it works... :)

simon-dew:
I think the migrated content has to be removed from its original location, or it will break the build. Antora doesn't like two files with the same name occupying the same component and module. Makes for complicated interlinked PRs (like this one) but each module only has to be migrated once.

PR #3098, opened 2023-03-03
Fix apostrophes in example commands

CB-DavidPancic:


simon-dew (approved):
Apostrophe to single straight quote looks good to me. I don't know enough about this endpoint to comment on the period at the end of the payload string.

PR #3106, opened 2023-03-20
Doc 10899

tonyjhillman:


modules/learn/pages/data/change-history.adoc
```diff
@@ -0,0 +1,184 @@
...
+When the change history is full, old records are automatically removed (by means of compaction), to allow space for new records.
+
+Change history can be either _on_ or _off_.
+This is determined at bucket-level: if the change history is _on_, it is active for _all collections_ in the bucket.
+
+For a bucket's change history to be switched _on_, the capacity of the change history must be established, as a positive integer corresponding either to _bytes_ or _seconds_.
+If no capacity is specified, no change history is generated.
+Optionally, positive integers can be specified for _both_ bytes and seconds: in such cases, record-removal occurs as soon as either capacity is reached.
+
+The minimum capacity for the change history is _2 Gigabytes_ (which would be specified as `2147483648`).
```
jimwwalker: gibibytes if we're following IEC standards - but the example certainly clarifies the true value. Similar pebibytes

modules/learn/pages/data/change-history.adoc
```diff
@@ -0,0 +1,184 @@
...
+The change history resides on disk.
+Its capacity is administrator-specified.
+When the change history is full, old records are automatically removed (by means of compaction), to allow space for new records.
+
+Change history can be either _on_ or _off_.
+This is determined at bucket-level: if the change history is _on_, it is active for _all collections_ in the bucket.
+
+For a bucket's change history to be switched _on_, the capacity of the change history must be established, as a positive integer corresponding either to _bytes_ or _seconds_.
+If no capacity is specified, no change history is generated.
+Optionally, positive integers can be specified for _both_ bytes and seconds: in such cases, record-removal occurs as soon as either capacity is reached.
```
jimwwalker: For the time configuration, referring to it as capacity and "as soon as either capacity is reached" doesn't explain well to me how the time configuration affects the available history.

Also with respect to the time parameter there's also a lag meaning "as soon as" is inaccurate description - records are only evaluated for removal during compaction - which may never occur if fragmentation is low and never increases (i.e. database remains unchanged).

The same for size - if the history size exceeds the configuration it's only if compaction is triggered  that we would remove records - so a user could exceed this limit but not exceed fragmentation thresholds and the system won't take action.

Maybe we can clarify around the following points?

_size_ compaction will keep the size of the history to no more than this value.
_time_ compaction will remove records that have an age which exceeds this configuration.

For the user we intended that these two values be used together and that they should make a decision about how many seconds of history they desire to keep yet set a size limit to overall ensure that unexpected increases in operations don't consumer more disk than is desirable (i.e. all of it).

modules/learn/pages/data/change-history.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+The minimum capacity for the change history is _2 Gigabytes_ (which would be specified as `2147483648`).
+The maximum is _1.8 Petabytes_ (which would be specified as `18446744073709551615`).
+If an integer outside this range is specified, an error is flagged, no file-size is established, and change history remains disabled for the bucket.
+
+Note that the change history is created _per bucket_, and is replicated in accordance with the number of intra-cluster replicas defined for the bucket.
+Therefore, if two replicas are configured, and the specified maximum size is 2 GB, the total size used for the change history across the cluster becomes 6 GB.
+
+If the change history for a bucket is switched _on_, the recording of changes can be performed in either of two ways, which are:
```
jimwwalker: For the following statements - i'd suggest we only discuss that the user should be considering which collections are "de duplication off" - note from the actually API that the user would invoke to configure collections we only talk about history=true or history=false, rather than talk solely about the process of de-duplication.

I'd suggest this section just mentions that once the bucket has history switched on that collections will need to be considered, e.g. "an exact change history for a collection can only be recorded when a collection has history=true"

iirc the default configuration is that all collections will have history=true so the user needs to consider which ones to disable if they so wish.

modules/learn/pages/data/change-history.adoc
```diff
@@ -0,0 +1,184 @@
...
+Change history can be enabled with either:
+
+* The REST API, using `POST /pools/default/buckets` (to create a bucket) or `POST /pools/default/buckets/<bucketName>` (to edit); specifying one or both of the parameters `historyRetentionBytes` and `historyRetentionSeconds`.
+See xref:rest-api:rest-bucket-create.adoc[Creating and Editing Buckets].
+
+* The CLI, using `couchbase-cli bucket-create` (to create a bucket) or `couchbase-cli bucket-edit` (to edit a bucket); specifying one or both of the parameters `history-retention-bytes` and `history-retention-seconds`.
+See xref:cli:cbcli/couchbase-cli-bucket-create.adoc[bucket-create] and xref:cli:cbcli/couchbase-cli-bucket-edit.adoc[bucket-edit].
+
+[#deduplication-bucket-wide]
+=== Deduplication Bucket-Wide Default Settings
```
jimwwalker: I'd avoid solely talking about deduplication, if we abstractly call this history, but mention part of history enabled means deduplication is disabled were good - reason behind this is in case history=true in future is expanded to include more than just deduplication off.

modules/learn/pages/data/change-history.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+A change history can be created with _deduplication_ either _off_ or _on_.
+If deduplication is _off_, the change history is more complete.
+If deduplication is _on_, the change history is more compact.
```
jimwwalker: Again i'd remove this case talking about history=false - it's not just that the change history more compact - it's just not a change history. I wouldn't want anyone thinking that there is any value in configuring history bytes+seconds yet having history=false

modules/learn/pages/data/change-history.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+* The REST API, using `POST /pools/default/buckets` (to create a bucket) or `POST /pools/default/buckets/<bucketName>` (to edit); specifying the parameter `historyRetentionCollectionDefault`.
+See xref:rest-api:rest-bucket-create.adoc[Creating and Editing Buckets].
+
+* The CLI, using `couchbase-cli bucket-create` (to create a bucket) or `couchbase-cli bucket-edit` (to edit a bucket); specifying the parameter `enable-history-retention-by-default`.
+See xref:cli:cbcli/couchbase-cli-bucket-create.adoc[bucket-create] and xref:cli:cbcli/couchbase-cli-bucket-edit.adoc[bucket-edit].
+
+=== Deduplication Collection-Specific Settings
+
+A change history can be made with deduplication either _off_ or _on_.
```
jimwwalker: As per earlier comments, with dedup on (history=false) the change history is questionable - we shouldn't guide the user to think about their history=true use-cases

modules/learn/pages/data/change-history.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+* The REST API, using `POST /pools/default/buckets/<bucket_name>/scopes/<scope_name>/collections` (to create a collection) or `PATCH /pools/default/buckets/<bucket_name>/scopes/<scope_name>/collections/<collection_name>` (to edit); specifying the parameter `history`.
+See xref:rest-api:creating-a-collection.adoc[Creating and Editing a Collection].
+
+* The CLI, using `couchbase-cli collection-manage` (to create or edit a collection); specifying the parameter `enable-history-retention`.
+See xref:cli:cbcli/couchbase-cli-collection-manage.adoc[collection-manage].
+
+=== Change-History Status and Content-Access
+
+The deduplication status of each collection can be returned by means of the `cbstats` tool, with the arguments `collections` and `collection-details`.
```
jimwwalker: cbstats will return history:true, similar to previous comments we deliberatley avoid mentioning the deduplication configuration

modules/learn/pages/data/change-history.adoc
```diff
@@ -0,0 +1,184 @@
...
+Note that generally speaking, deduplication results in a greater use of CPU, memory, and disk; due to the key-handling requirement for deduplicated items.
+
+* `ep_total_deduplicated_flusher`.
+When items are to be _flushed_ (persisted), the contents of multiple checkpoints are batched, so that all can be flushed.
+If deduplication is switched _on_, the checkpoints may already have had items removed through deduplication.
+However, a further check is made across checkpoints, prior to writes being persisted, to find and remove duplicate items.
+The total number of items recently removed from the flusher in this way is represented by `ep_total_deduplicated_flusher`.
+
+* `history_start_seqno`.
+The smallest document _seqno_ that is part of the history retention.
```
jimwwalker: "The smallest" doesn't quite work for me - i'd suggest removing this sentence and just starting from "This seqno is the.."

modules/learn/pages/data/change-history.adoc
```diff
@@ -0,0 +1,184 @@
...
+This seqno is the starting point of the history window.
+Hence, all documents having a higher seqno will be retained.
+As more and more data is removed from the window, this seqno will be incremented, to correspond to a later document.
+
+* `history_disk_size`.
+On-disk compressed size (after Magma’s block compression and per document compression) of the fragmented and unfragmented data that lies within the history window, expressed _per vBucket_.
```
jimwwalker: The value reported through cbstats all isn't per vBucket, it's the  total size for all vbuckets on the node.

modules/learn/pages/data/change-history.adoc
```diff
@@ -0,0 +1,184 @@
...
+* `ep_magma_history_time_evicted`.
+The total amount of data (in bytes) so far removed from the change history, due to the limit established with `historyRetentionSeconds` (REST API) or `history-retention-seconds` (CLI).
+
+* `ep_magma_history_size_evicted`.
+The total amount of data (in bytes) so far removed from the change history, due to the limit established with `historyRetentionBytes` (REST API) or `history-retention-bytes` (CLI).
+
+* `ep_db_history_file_size`.
+On-disk compressed size (after Magma’s block compression and per document compression) of the fragmented and unfragmented data that lies within the history window.
+
+* `vb_max_history_disk_size`.
```
jimwwalker: In prometheus this is actually seen as kv_vb_max_history_disk_size_bytes and also accepts a "state" label

e.g. user can do 

4 stats (active, replica, pending and dead)
curl -X GET "http://localhost:9000/pools/default/stats/range/kv_vb_max_history_disk_size_bytes?bucket=default" -u Administrator:asdasd | jq '.'

1 stat using state label
curl -X GET "http://localhost:9000/pools/default/stats/range/kv_vb_max_history_disk_size_bytes?bucket=default&state=replica" -u Administrator:asdasd | jq '.'

modules/learn/pages/data/change-history.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+Metrics are provided for retrieval by either the REST API or Prometheus.
+See xref:rest-api:rest-statistics.adoc[Statistics] for information on statistics-retrieval, and the xref:metrics-reference:metrics-reference.adoc[Metrics Reference] for a complete list of available statistics.
+
+In the following descriptions, the _history window_ refers to the time or space specified by the administrator for the change history (by means of, say, the parameters `historyRetentionBytes` and `historyRetentionSeconds`, used with the REST API).
+
+* `kv_total_enqueued`, `kv_total_deduplicated`, and `kv_total_deduplicated_flusher`.
+These have the same significance as their equivalents, provided by `cbstats`; described immediately above.
+
+* `ep_magma_history_time_evicted`.
```
jimwwalker: In prometheus this is kv_ep_magma_history_time_evicted_bytes

modules/learn/pages/data/change-history.adoc
```diff
@@ -0,0 +1,184 @@
...
+The total amount of data (in bytes) so far removed from the change history, due to the limit established with `historyRetentionSeconds` (REST API) or `history-retention-seconds` (CLI).
+
+* `ep_magma_history_size_evicted`.
```
jimwwalker: In prometheus this is kv_ep_magma_history_size_evicted_bytes

modules/learn/pages/data/change-history.adoc
```diff
@@ -0,0 +1,184 @@
...
+The total amount of data (in bytes) so far removed from the change history, due to the limit established with `historyRetentionBytes` (REST API) or `history-retention-bytes` (CLI).
+
+* `ep_db_history_file_size`.
```
jimwwalker: In prometheus this is kv_ep_db_history_file_size_bytes

modules/learn/pages/data/change-history.adoc
```diff
@@ -0,0 +1,184 @@
...
+On-disk compressed size (after Magma’s block compression and per document compression) of the fragmented and unfragmented data that lies within the history window.
+
+* `vb_max_history_disk_size`.
+Maximum amount of history retained across all vBuckets.
+The size here is _on-disk compressed_.
+
+* `ep_magma_history_logical_disk_size`.
```
jimwwalker: In prometheus this is kv_ep_magma_history_logical_disk_size_bytes

modules/learn/pages/data/change-history.adoc
```diff
@@ -0,0 +1,184 @@
...
+Size of fragmented and unfragmented data (after per document compression) that lies within the history window.
+This, along with `ep_magma_history_logical_data_size`, can be used to compute fragmentation in history.
+
+* `ep_magma_history_logical_data_size`.
```
jimwwalker: In prometheus this is kv_ep_magma_history_logical_data_size_bytes

modules/learn/pages/data/change-history.adoc
```diff
@@ -0,0 +1,184 @@
...
+As more and more data is removed from the window, this seqno will be incremented, to correspond to a later document.
+
+* `history_disk_size`.
+On-disk compressed size (after Magma’s block compression and per document compression) of the fragmented and unfragmented data that lies within the history window, expressed _per vBucket_.
+
+For more information, see the reference page for `cbstats` xref:cli:cbstats/cbstats-all.adoc[all].
+
+Note that the use of a tool such as _Grafana_ with some of these statistics -- so as to describe a continuum of readings over time -- may produce more insight than the inspection of an individual figure.
+
+==== Statistics Provided for REST and Prometheus
```
jimwwalker, line 128: The following comments correct the statistics keys into their prometheus form - which I believe is the same key to be used on REST API which you link to in "rest-statistics.adoc"

jimwwalker (changes requested):

modules/learn/pages/data/change-history.adoc
tonyjhillman: Thanks. I've changed both.

modules/learn/pages/data/change-history.adoc
tonyjhillman: I've now talked about 'size', and have restricted the description to the point you have raised.

modules/learn/pages/data/change-history.adoc
tonyjhillman: I have now removed deduplication, except for the API list, and discuss 'change history' only, in term of its being on/off.

modules/learn/pages/data/change-history.adoc
tonyjhillman: I've changed.

modules/learn/pages/data/change-history.adoc
tonyjhillman: I've modified.

modules/learn/pages/data/change-history.adoc
tonyjhillman: Ok.

modules/learn/pages/data/change-history.adoc
tonyjhillman: I've changed accordingly.

modules/learn/pages/data/change-history.adoc
tonyjhillman: I've modified accordingly.

modules/learn/pages/data/change-history.adoc
tonyjhillman: Thanks. I've fixed.

modules/learn/pages/data/change-history.adoc
tonyjhillman: I've added the 'kv_' to each of the affected stats. I've also included examples of using this metric.

modules/learn/pages/data/change-history.adoc
tonyjhillman: ok

modules/learn/pages/data/change-history.adoc
tonyjhillman: ok

modules/learn/pages/data/change-history.adoc
tonyjhillman: ok

modules/learn/pages/data/change-history.adoc
tonyjhillman: ok

modules/learn/pages/data/change-history.adoc
tonyjhillman: ok

modules/learn/pages/data/change-history.adoc
tonyjhillman: Thanks.

modules/learn/pages/data/change-history.adoc
```diff
@@ -0,0 +1,245 @@
...
+
+When xref:learn:buckets-memory-and-storage/storage-engines.adoc#storage-engine-magma[Magma] storage is used for a bucket, the changes made to documents within the bucket's collections can be recorded, in a _change history_.
+The change history resides on disk.
+Its capacity is administrator-specified.
+When the change history is full, old records are automatically removed (by means of compaction), to allow space for new records.
+
+Change history can be either _on_ or _off_.
+This is determined at bucket-level: if the change history is _on_, it is potentially active for _all collections_ in the bucket.
+
+If change history is _on_, the size of the change history should be specified in both _seconds_ and _bytes_:
```
jimwwalker: Suggest redraft this sentence - the "If change history is on" contradicts the "should"  later in the sentence?

Maybe something like "The size of the change history can be specified.... When either or both are configured the change history is _on_" ?

jimwwalker: Although later in the section you make the distinction about "either" config enables the feaure

modules/learn/pages/data/change-history.adoc
```diff
@@ -0,0 +1,245 @@
...
+This bucket-wide default setting can be established with either:
+
+* The REST API, using `POST /pools/default/buckets` (to create a bucket) or `POST /pools/default/buckets/<bucketName>` (to edit); specifying the parameter `historyRetentionCollectionDefault`.
+See xref:rest-api:rest-bucket-create.adoc[Creating and Editing Buckets].
+
+* The CLI, using `couchbase-cli bucket-create` (to create a bucket) or `couchbase-cli bucket-edit` (to edit a bucket); specifying the parameter `enable-history-retention-by-default`.
+See xref:cli:cbcli/couchbase-cli-bucket-create.adoc[bucket-create] and xref:cli:cbcli/couchbase-cli-bucket-edit.adoc[bucket-edit].
+
+[#individual-collection-setting]
+=== Individual Collection-Setting
```
jimwwalker, line 68: I wonder if it's worth a brief hint as to why someone would do this? Overall they can manage the size of their history better if they consider what collections need history and which do not.

modules/learn/pages/data/change-history.adoc
```diff
@@ -0,0 +1,245 @@
...
+The `cbstats` tool, using the `all` argument, returns statistics that include the following:
+
+* `ep_total_enqueued` and `ep_total_deduplicated`.
+These respectively provide the total number of checkpointed items, and the number of items discarded due to deduplication.
+The numbers correspond to items for the entire bucket, and include both active and replica vBuckets.
++
+If deduplication is switched off, `ep_total_enqueued` represents all checkpointed items.
+In such a case, if the bucket has one replica, and the value of `ep_total_enqueued` is _n_, the number of writes enqueued to the change history is  _n/2_.
++
+If deduplication is switched _on_, the statistic `ep_total_deduplicated` represents the number of items recently removed from checkpoints for the whole bucket, through deduplication.
```
jimwwalker: remove "recently" , this stat represents the current total deduplicated for the time the bucket has been "up"

modules/learn/pages/data/change-history.adoc
```diff
@@ -0,0 +1,245 @@
...
++
+By inspecting system performance (in terms of CPU, memory, disk, and measured latency) when deduplication is respectively on and off, insight can be gained into how optimisations can be achieved.
+Note that generally speaking, deduplication results in a greater use of CPU, memory, and disk; due to the key-handling requirement for deduplicated items.
+
+* `ep_total_deduplicated_flusher`.
+When items are to be _flushed_ (persisted), the contents of multiple checkpoints are batched, so that all can be flushed.
+If deduplication is switched _on_, the checkpoints may already have had items removed through deduplication.
+However, a further check is made across checkpoints, prior to writes being persisted, to find and remove duplicate items.
+The total number of items recently removed from the flusher in this way is represented by `ep_total_deduplicated_flusher`.
```
jimwwalker: again remove "recently"

jimwwalker (changes requested):

modules/learn/pages/data/change-history.adoc
tonyjhillman: I've written: This allows a determination to be made as to which collections need their change history to be recorded, and which do not.

modules/learn/pages/data/change-history.adoc
tonyjhillman: I've removed the later sentence, and here, have written: The size of the change history can be specified both in _seconds_ and in _bytes_.
If either or both are configured, change history is _on_.

modules/learn/pages/data/change-history.adoc
tonyjhillman: I've removed 'recently', and have written: the number of items removed from checkpoints for the whole bucket, through deduplication, while the bucket has been _up_.

modules/learn/pages/data/change-history.adoc
tonyjhillman: I've removed 'recently'.

PR #3125, opened 2023-03-29
DOC-10844: Developer Preview information for 7.2

tonyjhillman:


hyunjuV (approved):
Looks good to me.  I think that you already modified some wording around the tools package in a different change -- both versions look fine to me.

PR #3141, opened 2023-04-21
DOC-10850: Added note on XDCR and time-elapse

tonyjhillman:


tonyjhillman:
When we have this right, I'll backport also.

modules/learn/pages/clusters-and-availability/xdcr-filtering.adoc
```diff
@@ -9,6 +9,16 @@
...
 _XDCR filtering_ allows a document to be included in or excluded from a filtered replication, based on the document's fields and values.
 
+XDCR applies the filter during replication processing, as the data is passed through the replication pipeline.
+Filter-evaluation is performed only in the following circumstances:
+
+* A document is either created or modified.
+
+* A replication is restarted for all documents.
+
+Note that filter-evaluation is _not_ performed based on the measurement of elapsed time: therefore, the replication of a document cannot be triggered by establishing a filter on a time-value within the document.
```
hyunjuV: Since the time value doesn't have to be numeric -- can be any date, date time value, it might be better to use the term "date time value" instead of "time-value":
instead of "filter on a time-value within the document",
"filter on a date time value within the document"

modules/learn/pages/clusters-and-availability/xdcr-filtering.adoc
```diff
@@ -9,6 +9,16 @@
...
+For example, a filter that associates a future time, specified as a numeric value, with a key named `expiration_time` does _not_ trigger replication when the specified time is reached.
```
hyunjuV: For example, a filter that associates a future date time, with a key named `expiration_time` does _not_ trigger replication when the specified date time is reached.

hyunjuV (approved):
I made a couple of suggested changes; otherwise, approved.

modules/learn/pages/clusters-and-availability/xdcr-filtering.adoc
tonyjhillman: Ok, I've changed.

tonyjhillman: Thanks!

PR #3170, opened 2023-05-11
DOC-10890: Added query system events to reference page

tonyjhillman:


Tabitha-Oldfield (approved):
This looks good to me - thank you for adding it!

PR #3192, opened 2023-06-14
DOC-11160: Added preserve durability setting to UI and REST API

tonyjhillman:


modules/manage/pages/manage-settings/general-settings.adoc
```diff
@@ -97,14 +97,18 @@ image::manage-settings/node-availability.png["The Node Availability panel",548,a
...
+* *Auto-failover after _x_ seconds for up to _y_ node*: After the timeout period set here as _x_ seconds has elapsed, an unresponsive or malfunctioning node is failed over, provided that the limit on actionable events set here as _y_ (with the default value of 1) has not yet been reached.
 Data replicas are promoted to active on other nodes, as appropriate.
 This feature can only be used when three or more nodes are present in the cluster.
 The number of seconds to elapse is configurable: the default is 120; the minimum permitted is 5; the maximum 3600.
 This option is selected by default.
+
 * *Auto-failover for sustained data disk read/write failures after _z_ seconds*: After the timeout period set here as _z_ seconds has elapsed, a node is failed over if it has experienced sustained data disk read/write failures.
-The timeout period is configurable: the default length is 120 seconds; the minimum permitted is 5; the maximum 3600.
+The timeout period is configurable: the default length is 240 seconds; the minimum permitted is 5; the maximum 3600.
```
anuthan: I'm not aware we changed it from 120 to 240 seconds, could you point me to the improvement?

modules/manage/pages/manage-settings/general-settings.adoc
```diff
@@ -97,14 +97,18 @@ image::manage-settings/node-availability.png["The Node Availability panel",548,a
...
 This checkbox can only be checked if *Auto-failover after _x_ seconds for up to _y_ node* has also been checked.
+This option is selected by default.
+
+* *Preserve durable writes*: If this checkbox is checked (which is the default), and the node to be automatically failed over is a Data Service node that hosts data for a bucket whose Minimum Durability Level is greater than *none*; auto-failover occurs only if xref:learn:data/durability.adoc#majority[Majority] can be maintained across the nodes on which the bucket's data resides.
```
anuthan: The default is unchecked I believe. 
Also, the check is performed on all buckets regardless of Minimum Durability Level. 
CC @BenHuddleston  to correct me if I'm mistaken.

modules/rest-api/pages/rest-cluster-autofailover-enable.adoc
```diff
@@ -77,6 +79,10 @@ This parameter is optional, and is only available in Couchbase Enterprise Editio
 The value can be either `true` (the default) or `false`.
 The parameter and its value are ignored, if `enabled` is set to `false`.
 
+* `failoverPreserveDurabilityMajority`.
+Can be `true` (the default) or `false`.
+If this setting is `true`, and the node to be automatically failed over is a Data-Service node that hosts data for a bucket whose Minimum Durability Level is greater than `none`; auto-failover occurs only if xref:learn:data/durability.adoc#majority[Majority] can be maintained across the nodes on which the bucket's data resides.
```
anuthan: Same comment as above.

anuthan (commented):
@BenHuddleston Can you give this a look when you are back from vacation? Thanks.

modules/manage/pages/manage-settings/general-settings.adoc
tonyjhillman: My error - I have corrected.

modules/manage/pages/manage-settings/general-settings.adoc
tonyjhillman: Neither the REST or the CLI docs specify exactly the default - it is rather implied as 'none'. (So I should clarify whether the default is indeed 'none'.) At any rate, both say 'none' means asynchronous writes, and I guess I assumed that that meant 'majority' was not an issue. 

(NB The UI for bucket creation provides a pull-down menu whose default is 'none'.)

BenHuddleston: Default should be that we don't perform this check to preserve durable writes to maintain the current behaviour (i.e. unchecked). I didn't do the UI work but testing this locally had this default to unchecked.

As Abhi said, Minimum Durability Level is irrelevant here.

modules/manage/pages/manage-settings/general-settings.adoc
```diff
@@ -97,14 +97,18 @@ image::manage-settings/node-availability.png["The Node Availability panel",548,a
...
 This feature can only be used when three or more nodes are present in the cluster.
 The number of seconds to elapse is configurable: the default is 120; the minimum permitted is 5; the maximum 3600.
 This option is selected by default.
+
 * *Auto-failover for sustained data disk read/write failures after _z_ seconds*: After the timeout period set here as _z_ seconds has elapsed, a node is failed over if it has experienced sustained data disk read/write failures.
 The timeout period is configurable: the default length is 120 seconds; the minimum permitted is 5; the maximum 3600.
 This checkbox can only be checked if *Auto-failover after _x_ seconds for up to _y_ node* has also been checked.
+This option is selected by default.
+
+* *Preserve durable writes*: If this checkbox is checked (which is the default), and the node to be automatically failed over is a Data Service node that hosts data for a bucket whose Minimum Durability Level is greater than *none*; auto-failover occurs only if xref:learn:data/durability.adoc#majority[Majority] can be maintained across the nodes on which the bucket's data resides.
```
BenHuddleston: > auto-failover occurs only if xref:learn:data/durability.adoc#majority[Majority] can be maintained across the nodes on which the bucket's data resides.
 
This is incorrect. A majority is the required number of nodes to accept and process a durable write. This check will still allow the failover of nodes such that a majority may not be possible post-failover.  The intent is, quite simply, to avoid failing over a node if it could result in the loss of an acknowledged and committed durable write. This comment explains the mechanics of the change, which hopefully helps to explain the intent too - https://github.com/couchbase/ns_server/blob/cd0d04312f5f2dec6f411c7f5002b16f21f09bb9/src/failover.erl#L547-L599. Please let me know if that's not clear.

modules/rest-api/pages/rest-cluster-autofailover-enable.adoc
BenHuddleston: Same comment about majority from above

BenHuddleston (changes requested):

modules/rest-api/pages/rest-cluster-autofailover-enable.adoc
fleray: > * `failoverPreserveDurabilityMajority`.
> Can be `true` (the default) or `false`.

=> No. In 7.2.0, it is **false** by default:

[Preserve durable writes is unchecked by default](https://issues.couchbase.com/secure/attachment/211446/Screenshot%202023-06-27%20at%2014.55.43.png)

Please make sure the upcoming doc page does not contradict what is inside the current UI.

Thanks.

modules/learn/pages/data/durability.adoc
```diff
@@ -216,16 +216,34 @@ If a durable write is attempted on a bucket that has been configured with three
...
+
+[#preserving-durable-writes]
+==== Preserving Durable Writes
+
+In Couchbase Enterprise Server 7.2+, auto-failover can be configured not to occur if a node's failover would result in the loss of durably written data.
+This means ensuring that at least one node containing the durably written data is preserved.
+
+If auto-failover is configured in this way:
+
+* If data has been durably written to _one_ node (which is to say, no replica exists), the node cannot be automatically failed over.
```
BenHuddleston: In the case in which we have 0 replicas a failover is possible. A different warning is presented, related to failing over a node without replicas:

Screenshot 2023-07-04 at 11 05 19

BenHuddleston: I'd suggest re-wording the following (and the below two/three node case), it feels potentially misleading/ambiguous:

> If data has been durably written to _one_ node

This is probably clearer if we word this in terms of the durability majority/replica count counfigured used for the write, i.e. "If data has been durably written with 0 replicas configured".

PR #3207, opened 2023-07-05
DOC-11085: Unisgned 64 bit integer truncation

tonyjhillman:


modules/learn/pages/clusters-and-availability/size-limitations.adoc
```diff
@@ -71,4 +72,9 @@ If the execution time for a document exceeds the specified value, the execution
 | Administrator-defined
 | The maximum permitted numbers of connections permitted for memcached connections, and for system-user connections, for the cluster.
 These numbers are established by means of the REST API: see xref:rest-api:rest-manage-cluster-connections.adoc[Managing Cluster Connections], for information.
+
+| Unsigned 64 bit integer
```
dhaggart: We don't have the concept of signed & unsigned. So just "64 bit integer" here.

modules/learn/pages/clusters-and-availability/size-limitations.adoc
```diff
@@ -71,4 +72,9 @@ If the execution time for a document exceeds the specified value, the execution
...
+| latexmath:[2^{63} - 1]
+| The integer is truncated if larger, due to floating-point arithmetic approximation.
```
dhaggart: I'd phrase it as something like:
Integers outside this range are stored as 64-bit floating-point values and are subject to approximation inherent in the IEEE floating-point format.

tonyjhillman: I've made changes - let me know if it's better. Thanks.

dhaggart (approved):

dhaggart (changes requested):
Sorry - I missed a mistake I made in the DOC issue;  the range is +-1 * (2^63-1)

tonyjhillman:
I've modified - does that look right?

dhaggart:
Yep, looks good now.  (Sorry for that.)

tonyjhillman:
Great, thanks. Last question - so, how far back does this inclusion need to be made? All the way to 5.0? (Although actually, we only update back to 6.6 at this point...older docs are archived.)

dhaggart:
This has always been the case AFAIK.  So as far back as you can with active docs...

tonyjhillman:
Thanks, I'll update the others.

PR #3226, opened 2023-07-13
DOC-10605: Added maximum query formula

tonyjhillman:


modules/install/pages/sizing-general.adoc
```diff
@@ -362,6 +362,9 @@ There are a few questions that will help size the cluster:
 Different queries have different resource requirements.
 A simple query might return results within milliseconds while a complex query may require several seconds.
 
+The number of queries that may be processed simultaneously may be approximated with the formula _CPU_cores * 4_.
```
kartika-nair: This should probably be changed to the ASCII underscore:
```suggestion
The number of queries that may be processed simultaneously may be approximated with the formula _CPU&#95;cores * 4_.
```

kartika-nair (changes requested):

modules/install/pages/sizing-general.adoc
tonyjhillman: Sorry, missed that. Thanks.

tonyjhillman: Is this required for prior releases also?..I assume so.

kartika-nair: Yes, I believe this calculation applies to all versions that aren't End of Life at the moment (Couchbase Server 6.6 and above).

PR #3291, opened 2023-07-25
DOC-6343 Add documentation for running multiple single node containers

RayOffiah:
https://issues.couchbase.com/browse/DOC-6343

### DIFF
--- a/modules/install/pages/getting-started-docker.adoc
+++ b/modules/install/pages/getting-started-docker.adoc
@@ -10,6 +10,7 @@ For more traditional Docker deployments, use the following sections below:
 
 * <<section_jvt_zvj_42b>>
 * <<section_msh_fbl_42b>>
+* <<section_deploy_multiple_clusters>>
 
 If you're simply looking for the official Couchbase Server Docker image, you can find it on https://hub.docker.com/_/couchbase/[Docker Hub^].
 
@@ -38,7 +39,16 @@ To set up Docker on the host computer, refer to Docker's https://www.docker.com/
 $ docker run -d --name db -p 8091-8096:8091-8096 -p 11210-11211:11210-11211 couchbase
 ----
 
-After running the above command, a single instance (`db`) of the latest https://hub.docker.com/_/couchbase/[official Couchbase Server container image^] is downloaded and installed onto the host computer.
+.Multiple instances with Docker
+****
+When running multiple instances on the same machine, make use of Docker's `-p` option to map `external:internal` ports used by the container.
+
+For example:
+
+`-p 9091:8091` instructs the container to map the external machine port `9091` to the container application's port `8091`.
+****
+
+After running the above command, a single instance (`db`) of the latest https://hub.docker.com/_/couchbase/[official Couchbase Server container image^] is downloaded and run on the host computer.
 If a traditional installation of Couchbase Server is running locally on the host computer, the port mappings specified using the `-p` option may fail.
 Ensure that you stop any local instance of Couchbase Server before running this command.
 ////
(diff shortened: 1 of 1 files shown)


RayOffiah:
Preview at:

https://rayoffiah.github.io/DOC-6343/server/current/install/getting-started-docker.html#section_deploy_multiple_clusters

tonyjhillman (approved):

modules/install/pages/getting-started-docker.adoc
```diff
@@ -323,6 +325,130 @@ image::docker-single-machine-db123.png["The 'Servers' tab showing three nodes in
...
+$ docker run -d --name db1 -p 8091-8096:8091-8096 -p 11210-11211:11210-11211 couchbase
+----
++
+[source,console]
+----
+$ docker run -d --name db2 -p 9091-9096:8091-8096 -p 12210-12211:11210-11211 couchbase
+----
++
+After running the above commands,
+three instances (`db1` and `db2`) of the latest https://hub.docker.com/_/couchbase/[official Couchbase Server container image^] are downloaded
```
malarky: Should be "two instances".

modules/install/pages/getting-started-docker.adoc
```diff
@@ -323,6 +325,130 @@ image::docker-single-machine-db123.png["The 'Servers' tab showing three nodes in
...
+and installed onto the host computer.
```
malarky: I wouldn't say "installed" when talking about a container.

modules/install/pages/getting-started-docker.adoc
```diff
@@ -323,6 +325,130 @@ image::docker-single-machine-db123.png["The 'Servers' tab showing three nodes in
...
++
+image::welcome-join-existing-cluster.png["The Couchbase Server setup screen.",400]
+. Click btn:[Join Existing Cluster].
+
++
+image::join-cluster-details.png[Joining an existing cluster, 450]
++
+. Fill in the IP address associated with container `db1`, along with the administrator password.
++
+. Press btn:[Join Cluster]
```
malarky: I though this section was about running multiple single-node clusters, so shouldn't be any discussion around adding nodes and rebalance.

modules/install/pages/getting-started-docker.adoc
```diff
@@ -323,6 +325,130 @@ image::docker-single-machine-db123.png["The 'Servers' tab showing three nodes in
...
+Make sure to run each of the following commands:
++
+[source,console]
+----
+$ docker run -d --name db1 -p 8091-8096:8091-8096 -p 11210-11211:11210-11211 couchbase
+----
++
+[source,console]
+----
+$ docker run -d --name db2 -p 9091-9096:8091-8096 -p 12210-12211:11210-11211 couchbase
```
malarky: You can't re-use `12210-12211:11210-11211`, need to find a new set as you did with `8091`.

malarky: Also, I don't think a client will be able to connect unless you set up Alternate Addresses.

malarky (changes requested):
Some specific comments, but also wondering if we just need to provide some general guidance on running multiple instances, regardless of how they are clustered.

I *think* (but you should test and verify), if you only want to get to the UI then you can get away with just using the `-p` for 8091 and incrementing the host port.  But, if you actually want an application to connect you'll have to set up alternate addresses for every container after the first one.

If you go over the basics (use of `-p` and alternate addresses), then users can set up as many containers as they want and then cluster them how they want.

modules/install/pages/getting-started-docker.adoc
RayOffiah: Fixed throughout document

PR #3292, opened 2023-07-25
Update fts-supported-queries-date-range.adoc

jon-strabala:
The documentation was completely missing for this complex feature/syntax - this is valid for all 7.X version and most likely 6.X versions of the documentation.

This is a recurring problem no one (our own staff) knows how to make Custom DateTime Parsers becuase they don't know that the templates have to be in Golang syntax.

### DIFF
--- a/modules/fts/pages/fts-supported-queries-date-range.adoc
+++ b/modules/fts/pages/fts-supported-queries-date-range.adoc
@@ -21,3 +21,43 @@ The [.param]`inclusive_start` and [.param]`inclusive_end` properties in the quer
  "field": "review_date"
 }
 ----
+
+The default datetime parser is *dateTimeOptional* which contaings the following layouts s per the syntax in https://pkg.go.dev/time :
+
+[cols="<,<", options="header"]
+|===
+|Format |Name
+
+|2006-01-02T15:04:05.999999999Z07:00 |time.RFC3339Nano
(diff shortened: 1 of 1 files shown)


sarahlwelton:
Jon, isn't this satisfied by: https://docs.couchbase.com/cloud/search/create-custom-date-time-parser.html

There is equivalent in Server, I just need to fix the ToC.

And if we're going to be phasing out this documentation with the ToC updates, I don't see a need to update this old file.

jon-strabala:
> Jon, isn't this satisfied by: https://docs.couchbase.com/cloud/search/create-custom-date-time-parser.html
> 
> There is equivalent in Server, I just need to fix the ToC.

I just saw this ... as such I want to adjust the PR as "create-custom-date-time-parser" just doesn't give enough information.  No one want's to spend time digging in other offsite documents - this burned both Ankit and myself last weekend.

So I want to update both files a bit (oh there is a third file) it all should be consistent at a minimum.

sarahlwelton:
No, we don't document 3rd party because it's subject to change and we can't be held responsible for maintaining documentation on the Go package. I would recommend that you don't make too many changes, here.

jon-strabala:
Understood, however one issue the 3rd party go site does not' document our default dateTimeOptions (date parser) IMHO what this is and what it covers should be documented.

I will think about things a bit more what I propose will be minimal

PR #3295, opened 2023-07-27
DOC-10720: Add Cluster Manager and XDCR Metrics Reference

malarky:
@osfameron Here's the Cluster Manager and XDCR metrics - please don't publish without the filter to remove `stability == internal`, thanks!

Also, deliberately left the "Under construction" label on the KV metrics as they are incomplete.


osfameron:
@malarky preview build at https://preview.docs-test.couchbase.com/DOC-10720/server/current/metrics-reference/data-service-metrics.html 👍

malarky:
Thanks @osfameron, looks good!

PR #3296, opened 2023-07-31
DOC-6486: Added mcstat page

tonyjhillman:


jameseh96 (commented):
Looks fine. In principle, we might want to write up short descriptions of each of the stat groups, but KV doesn't guarantee any stability to what stats are available where over mcstat, so that might be difficult to keep in sync.

Overall I think this will be good for customers to know mcstat exists, and they can look at the cmd help themselves for more detail

tonyjhillman:
Thanks, James!

PR #3301, opened 2023-07-31
DOC-9510: Updated cbstats memory page

tonyjhillman:


modules/cli/pages/cbstats/cbstats-memory.adoc
```diff
@@ -15,120 +15,96 @@ cbstats [host]:11210 memory
...
+ ep_blob_overhead:                   4247418
+ ep_item_num:                        3584
+ ep_kv_size:                         48627966
+ ep_max_size:                        209715200
+ ep_mem_high_wat:                    178257920
+ ep_mem_high_wat_percent:            0.85
+ ep_mem_low_wat:                     157286400
+ ep_mem_low_wat_percent:             0.75
+ ep_mem_used_primary:                68505152
+ ep_mem_used_secondary:              0
```
jameseh96, line 105: Might be worth adding these - they're reasonably straightforward for a user to grasp.

primary:

If this is a magma bucket, this is the memory usage of the bucket excluding magma. Else, this is equivalent to mem_used / bytes.

secondary:

If this is a magma bucket, this is the memory usage of the storage layer. Else, this is unused and will be zero.

jameseh96 (commented):
Looks good - I think the remaining undocumented stats are reasonable to leave as "internal", because their descriptions would include KV implementation specific details and could reasonably be misconstrued by customers 👍

tonyjhillman:
Thanks, James!

PR #3302, opened 2023-08-01
DOC-10643 Change code samples to use recommended swapiness of 1

RayOffiah:
https://issues.couchbase.com/browse/DOC-10643

### DIFF
--- a/modules/install/pages/install-swap-space.adoc
+++ b/modules/install/pages/install-swap-space.adoc
@@ -55,7 +55,7 @@ If the output doesn't match the <<recommended-swappiness-settings,recommended se
 +
 [source,console]
 ----
-sudo sh -c 'echo 0 > /proc/sys/vm/swappiness'
+sudo sh -c 'echo 1 > /proc/sys/vm/swappiness'
 ----
 +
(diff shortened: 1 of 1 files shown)


modules/install/pages/install-swap-space.adoc
```diff
@@ -74,12 +74,12 @@ sudo sh -c 'echo "" >> /etc/sysctl.conf'
 +
 [source,console]
 ----
-sudo sh -c 'echo "#Set swappiness to 0 to avoid swapping" >> /etc/sysctl.conf'
+sudo sh -c 'echo "#Set swappiness to 1 to avoid swapping" >> /etc/sysctl.conf'
```
floriancb: "to avoid swapping" is not quite correct now, with this setting. Maybe instead use "to minimize swapping"

floriancb (changes requested):

modules/install/pages/install-swap-space.adoc
RayOffiah: Yep, that makes sense.

floriancb (approved):

PR #3318, opened 2023-08-24
Doc 11430

tonyjhillman:


modules/rest-api/pages/rest-xdcr-connection-precheck.adoc
```diff
@@ -0,0 +1,112 @@
...
+The `taskId` thus returned can now be used to return the status of the connection-check that has been started.
+The following example duly returns the current status of this connection-check:
+
+----
+curl -X GET -u Administrator:password \
+http://localhost:8091/xdcr/connectionPreCheck?taskId=task/127.0.0.1/10 | jq '.'
+----
+
+If successful, the call returns an object such as the following:
+
```
hyunjuV, line 92: In the example below, the response shows "done": true -- this mean that the check is complete.  If the response shows "done": false, the check is not complete, and you should make the above GET call again.  It may take as long as 60 seconds for the response to show "done": true.

modules/rest-api/pages/rest-xdcr-connection-precheck.adoc
```diff
@@ -0,0 +1,112 @@
...
+  "username": "Administrator"
+}
+----
+
+The `taskId` thus returned can now be used to return the status of the connection-check that has been started.
+The following example duly returns the current status of this connection-check:
+
+----
+curl -X GET -u Administrator:password \
+http://localhost:8091/xdcr/connectionPreCheck?taskId=task/127.0.0.1/10 | jq '.'
```
hyunjuV: I remember that the http had to be in double quotes for the curl command to work -- so, I think that the example should show that -- save a user some frustration.
"http://localhost:8091/xdcr/connectionPreCheck?taskId=task/127.0.0.1/10" |  jq '.'

The design [document](https://docs.google.com/document/d/1dPQkwnJd68bsjGn1TA9Wdd6qtz15zYdyZ8vkawDtwUo/edit#) also showed the double quotes around the http url in all the examples -- need only for the GET call -- the POST Is fine the way it is.

hyunjuV (approved):
Looks good except that the pictures may change -- so may need to update later.
I also made couple of comments about the GET /xdcr/connectionPreCheck?taskId=<taskId> call.  I think that they are important information that should be added, but I'll go ahead and approve now.

modules/rest-api/pages/rest-xdcr-connection-precheck.adoc
tonyjhillman: I've added.

modules/rest-api/pages/rest-xdcr-connection-precheck.adoc
tonyjhillman: Ok.

tonyjhillman:
Thanks, I've modified. I'll update the graphics when the change has been made.

PR #3324, opened 2023-09-04
Doc 11268

tonyjhillman:


tonyjhillman:
Pre-built version available for review, as follows:

[Manage System Secrets](https://tonyjhillman.github.io/DOC-11268/server/current/manage/manage-security/manage-system-secrets.html)

[System Secrets API](https://tonyjhillman.github.io/DOC-11268/server/current/rest-api/rest-secret-mgmt.html)

modules/manage/pages/manage-security/manage-system-secrets.adoc
```diff
@@ -54,14 +54,27 @@ Note that the `master-password` CLI command must be run on the same host on whic
 [#password_rotation]
 == Performing Rotation
 
-The Couchbase Server Secret-Management system allows you to _rotate_ (periodically change, to reduce the risk of illicit discovery or deciphering) the different elements of the system:
+The Couchbase Server Secrets-Management system allows you to _rotate_ (periodically change, to reduce the risk of illicit discovery or deciphering) the different elements of the system:
 
-* *Master-password rotation*: This _first level of rotation_ is achieved by setting a new password, using the xref:cli:cbcli/couchbase-cli-master-password.adoc[CLI command], the xref:rest-api:rest-secret-mgmt.adoc[REST API method], or (other than on Mac OS X) the environment variable, as indicated above.
+* *Master-password rotation*:e _first level of rotation_ is achieved by setting a new password, using the xref:cli:cbcli/couchbase-cli-master-password.adoc[CLI command], the xref:rest-api:change-master-password.adoc[POST /node/controller/changeMasterPassword] method, or (other than on Mac OS X) the environment variable, as indicated above.
```
timofey-barmin: > e _first level of rotation_

Typo

modules/rest-api/pages/change-master-password.adoc
```diff
@@ -0,0 +1,48 @@
...
+[#http-methods-and-uris]
+== HTTP Method and URI
+
+----
+POST /node/controller/changeMasterPassword
+----
+
+== Description
+
+This command sets the master password.
```
timofey-barmin: We should probably mention that it sets the master password for this node (master password and all secrets management settings are per-node settings).

modules/rest-api/pages/system-secrets-configuration.adoc
```diff
@@ -0,0 +1,316 @@
...
+
+----
+GET /nodes/self/secretsManagement
+
+POST /node/controller/secretsManagement
+----
+
+[#description]
+== Description
+Configures the _encryption service_; which comprises the master password, encryption keys, key storage, and the location of key-control scripts.
```
timofey-barmin: I think I used slightly different terminology in the internal document which probably confused you. Sorry about that. I think we should stay consistent with the rest of the documentation, so I suggest we rename some terms here (on this page only).
"encryption key" -> "data key"
"Encryption key" is what we call "data key" in documentation.

"encryption service" -> "secrets management"
I think we don't use "encryption service" in documentation.

modules/rest-api/pages/system-secrets-configuration.adoc
```diff
@@ -0,0 +1,316 @@
...
+== Curl Syntax
+
+----
+curl -X GET http://<ip-address-or-hostname>:8091/nodes/self/secretsManagement
+  -u <username>:<password>
+
+curl -X POST http://<ip-address-or-hostname>:8091/node/controller/secretsManagement
+  -u <username>:<password>
+  -H "Content-Type:application/json"
+  -d "$(cat)" | jq <payload>
```
timofey-barmin: >  `-d "$(cat)" | jq <payload>`

just `-d <payload>`

modules/rest-api/pages/system-secrets-configuration.adoc
```diff
@@ -0,0 +1,316 @@
...
+
+== Curl Syntax
+
+----
+curl -X GET http://<ip-address-or-hostname>:8091/nodes/self/secretsManagement
+  -u <username>:<password>
+
+curl -X POST http://<ip-address-or-hostname>:8091/node/controller/secretsManagement
+  -u <username>:<password>
+  -H "Content-Type:application/json"
```
timofey-barmin: Please mention that `application/x-www-form-urlencoded` is also supported (it can be omitted in curl command because it is used by default)

If /node/controller/secretsManagement/encryptionService is used the curl format should be:

```
curl -X POST http://<ip-address-or-hostname>:8091/node/controller/secretsManagement
  -u <username>:<password> [-d <option>=<value>]*
```

If /node/controller/secretsManagement is used the format should be:

```
curl -X POST http://<ip-address-or-hostname>:8091/node/controller/secretsManagement
  -u <username>:<password> [-d encryptionService.<option>=<value>]*
```

modules/rest-api/pages/system-secrets-configuration.adoc
```diff
@@ -0,0 +1,316 @@
...
+[#description]
+== Description
+Configures the _encryption service_; which comprises the master password, encryption keys, key storage, and the location of key-control scripts.
+
+The *Full Admin*, *Local User Security Admin*, or *External User Security Admin* role is required.
+
+== Curl Syntax
+
+----
+curl -X GET http://<ip-address-or-hostname>:8091/nodes/self/secretsManagement
```
timofey-barmin: I just realized that we probably should document 

GET  /nodes/self/secretsManagement/encryptionService
POST /nodes/self/secretsManagement/encryptionService

instead of 

GET  /nodes/self/secretsManagement
POST /nodes/self/secretsManagement

they currently configure the same thing but the former looks simpler. Up to you.

I added examples for /nodes/self/secretsManagement/encryptionService to the internal doc (example 2).
I also added some examples for content-type "application/x-www-form-urlencoded" (example 3).

Please let me know if you have questions.

modules/rest-api/pages/system-secrets-configuration.adoc
```diff
@@ -0,0 +1,316 @@
...
+* `passwordSource`.
+The value can be `env` (which is the default) or `script`.
+If the value is `env`, the master password is read from the environment variable specified by `passwordEnv`.
+If the value is `script`, the master password is provided by the script specified by `passwordCmd`.
+
+* `passwordEnv`.
+The name of the environment variable from which the master password is read (provided that the value of `passwordSource` is `env`).
+By default, the value of `passwordEnv` is `CB_MASTER_PASSWORD`.
+
+* `passwordCmd`.
```
timofey-barmin, line 70: We need to describe how this command is used:
1. When it is executed: normally when node is starting, but also can be called if secrets manager restarts which can be caused by reconfiguration or by a bug.
2. It is expected that this command exits with status code 0. If exit status is different it is considered as an error.
3. It is expected that this command writes the master password to stdout.

modules/manage/pages/manage-security/manage-system-secrets.adoc
tonyjhillman: Thanks!

modules/rest-api/pages/change-master-password.adoc
tonyjhillman: I've modified accordingly.

modules/rest-api/pages/system-secrets-configuration.adoc
tonyjhillman: I've changed each instance of 'encryption key' to 'data key'; and have changed each instance of 'encryption service' to 'secrets management'.

modules/rest-api/pages/system-secrets-configuration.adoc
tonyjhillman: I've cut back to -d then the payload, so removing ... "$(cat)" | jq

modules/rest-api/pages/system-secrets-configuration.adoc
tonyjhillman: I have switched to the /nodes/self/secretsManagement/encryptionService URI for all.

modules/rest-api/pages/system-secrets-configuration.adoc
tonyjhillman: I've added.

modules/rest-api/pages/system-secrets-configuration.adoc
```diff
@@ -15,52 +15,46 @@ POST /node/controller/secretsManagement
...
-----
-{
-  "encryptionService": {
-    [<option>: <value>]*
-  }
-}
 ----
 
-Each `option` is described below.
+The `header` can be either `"Content-Type:application/json"` or `"application/x-www-form-urlencoded"`.
```
timofey-barmin: The `Content-Type` can be either `application/json` or `application/x-www-form-urlencoded`

tonyjhillman: Thanks.

timofey-barmin (approved):
LGTM

tonyjhillman:
Thanks again.

PR #3357, opened 2023-09-18
DOC-11487: Updated transactions/XDCR text

tonyjhillman:


modules/learn/pages/data/transactions.adoc
```diff
@@ -172,11 +172,21 @@ Note that you can choose to update the scan consistency level to `not_bounded` i
 
 === Transactions and Replication (XDCR)
 
-xref:learn:clusters-and-availability/xdcr-overview.adoc[Cross Data Center Replication] (XDCR) supports eventual consistency of transactional changes. No uncommitted changes will be ever sent to the target clusters.  Once committed, the transactional changes will arrive one by one at the target. If the connection is lost midway while receiving a transaction it is possible for the target to receive a partial transaction.
+xref:learn:clusters-and-availability/xdcr-overview.adoc[Cross Data Center Replication (XDCR)] supports eventual consistency of transactional changes.
+No uncommitted changes will be ever sent to target clusters.
+Once committed, the transactional changes arrive one by one at the target.
+If the connection is lost midway, it is possible for the target to receive a partial transaction.
```
nelio2k, line 178: I don't think this is possible. Considering transactions in a single document at this point: When a transaction is started, and before it is committed, the uncommitted transaction will live in a temporary xattr metadata wihtin the document. XDCR will not replicate this over. Thus, it's not possible for the target to receive anything *but* a document after the transaction is completed.

hyunjuV: If the transaction involves multiple documents, then, it is possible for the target to receive a partial transaction (in the scenario that the connection is lost midway), isn't it?   I think that's all it is saying.

hyunjuV (approved):

modules/learn/pages/data/transactions.adoc
hyunjuV: ^^ @nelio2k -- can you confirm?  If a transaction involves multiple documents, once the transaction is committed, xdcr does not guarantee that all of the documents in a transaction will be replicated at the same time, and therefore, if the connection is lost, the customer could see a partial transaction on the target -- where document A was replicated before the connection was lost, but document B was not.

nelio2k: Correct

PR #3359, opened 2023-09-19
Couchstore - not Magma

RichardSmedley:
DOC-11504


simon-dew (approved):
Would prefer it if the new sentence was on a new line, but not a blocker. LGTM 👍

RichardSmedley:
> Would prefer it if the new sentence was on a new line, but not a blocker. LGTM 👍

Thanks - yes, chose to follow existing admonition (already at two sentences on one line), rather than refactor as time was pressing that evening.

PR #3365, opened 2023-09-25
DOC-10720: Add missing XDCR Metrics

malarky:
@osfameron As discussed, a couple of XDCR metrics missing (they were added in 7.2.1).


osfameron (approved):
Looks good. I was just wondering "oh no, what do we do about 7.2.0?" and then saw the 'added' field which we added support for recently, good to see it's earning its keep already 👍

malarky:
> Looks good. I was just wondering "oh no, what do we do about 7.2.0?" and then saw the 'added' field which we added support for recently, good to see it's earning its keep already 👍

The system works! 😁

PR #3370, opened 2023-10-06
DOC-11544 Update upgrade instructions to reflect implications of Erlang and Open SSL upgrades

RayOffiah:
https://issues.couchbase.com/browse/DOC-11544

### DIFF
--- a/modules/install/pages/upgrade.adoc
+++ b/modules/install/pages/upgrade.adoc
@@ -76,18 +76,22 @@ For example, if you are upgrading three enterprise nodes (`*Node{nbsp}1*`, `*Nod
 | Starting Version |  Path to Current Version
 
 | 5.0.x
-| 5.0.x and 5.1.x -> 6.0.4+ -> 6.6 -> 7.2x
+| 5.0.x and 5.1.x -> 6.0.4+ -> 6.6 -> 7.2.x (except 7.2.3) -> 7.2.3
 
-5.0.x and 5.1.x -> 6.6 -> 7.2x
+5.0.x and 5.1.x -> 6.6 -> 7.2.x (except 7.2.3) -> 7.2.3
 
-5.5.0+ -> 6.6 -> 7.2x
+5.5.0+ -> 6.6 -> 7.2.x (except 7.2.3) -> 7.2.3
 
 | 6.x
-| 6.0 -> 6.5 -> 7.2x
+| 6.0 -> 6.6 -> 7.2.x (except 7.2.3) -> 7.2.3
 
 | 7.x
-| 7.0 -> 7.1 -> 7.2x
+| 7.0 -> 7.1 -> 7.2.3
+
+7.0 -> 7.2.x (except 7.2.3) -> 7.2.3
+
 |===
+
 
 [#table-upgrade-community]
 === Community Edition Upgrade Paths
@@ -97,16 +101,19 @@ For example, if you are upgrading three enterprise nodes (`*Node{nbsp}1*`, `*Nod
 | Starting Version | Path to Current Version
 
 | 5.x
-| 5.x -> 6.0 -> 6.5 -> 7.2x
+| 5.x -> 6.0 -> 6.6 -> 7.2.x (except 7.2.3) -> 7.2.3
 
 | 6.x
-| 6.0 -> 6.5 -> 7.2x
+| 6.0 -> 6.6 -> 7.2.x (except 7.2.3) -> 7.2.3
 
 | 7.x
-| 7.0 -> 7.1 -> 7.2x
+| 7.0 -> 7.1 -> 7.2.3
+
+7.0 -> 7.2.x (except 7.2.3) -> 7.2.3
+
 |===
 
-.Important note when upgrading from 7.0.4 to 7.2x on Windows 2019
+.Important note when upgrading from 7.0.4 to 7.2.x on Windows 2019
 [sidebar]
 ****
 Upgrading from version 7.0.4 -> 7.2x on Windows Server 2019 may result in a missing `java` executable files.


timofey-barmin:
Not sure if we need to specific instructions for 7.1. Probably we  should rather modify 7.0.x paths (they need to be modified anyway):
For 7.x we support the following upgrades:
7.0 -> 7.1 -> 7.2.3
7.0 -> 7.2.x (except 7.2.3) -> 7.2.3

For 6.x and 5.x we need to do that change as well. Basically we need to add "(except 7.2.3) -> 7.2.3" to all Enterprise and Community paths.

For example for 6.x:

6.0 -> 6.5 -> 7.2.x (except 7.2.3) -> 7.2.3

I think we also should add "." (dot) before x in all versions like 7.2x (make it 7.2.x) if it is not done intentionally (looks like it is a bug).

tonyjhillman (approved):

modules/install/pages/upgrade.adoc
```diff
@@ -76,19 +76,23 @@ For example, if you are upgrading three enterprise nodes (`*Node{nbsp}1*`, `*Nod
...
 
-5.0.x and 5.1.x -> 6.6 -> 7.2x
+5.0.x and 5.1.x -> 6.6 -> 7.2.x -> 7.2.x (except 7.2.3) -> 7.2.3
 
-5.5.0+ -> 6.6 -> 7.2x
+5.5.0+ -> 6.6 -> 7.2.x -> 7.2.x (except 7.2.3) -> 7.2.3
 
 | 6.x
-| 6.0 -> 6.5 -> 7.2x
+| 6.0 -> 6.6 -> 7.2.x -> 7.2.x (except 7.2.3) -> 7.2.3
```
timofey-barmin: Two "7.2.x" are unnecessary. Please change it to:
```
 6.0 -> 6.6 -> 7.2.x (except 7.2.3) -> 7.2.3
```

modules/install/pages/upgrade.adoc
```diff
@@ -76,19 +76,23 @@ For example, if you are upgrading three enterprise nodes (`*Node{nbsp}1*`, `*Nod
 | Starting Version |  Path to Current Version
 
 | 5.0.x
-| 5.0.x and 5.1.x -> 6.0.4+ -> 6.6 -> 7.2x
+| 5.0.x and 5.1.x -> 6.0.4+ -> 6.6 -> 7.2.x (except 7.2.3) -> 7.2.3
 
-5.0.x and 5.1.x -> 6.6 -> 7.2x
+5.0.x and 5.1.x -> 6.6 -> 7.2.x -> 7.2.x (except 7.2.3) -> 7.2.3
```
timofey-barmin: Two "7.2.x" are unnecessary. Please change it to
```
5.0.x and 5.1.x -> 6.6 -> 7.2.x (except 7.2.3) -> 7.2.3
```

modules/install/pages/upgrade.adoc
```diff
@@ -76,19 +76,23 @@ For example, if you are upgrading three enterprise nodes (`*Node{nbsp}1*`, `*Nod
...
 
-5.5.0+ -> 6.6 -> 7.2x
+5.5.0+ -> 6.6 -> 7.2.x -> 7.2.x (except 7.2.3) -> 7.2.3
```
timofey-barmin: Two "7.2.x" are unnecessary. Please change it to:
```
5.5.0+ -> 6.6 -> 7.2.x (except 7.2.3) -> 7.2.3
```

modules/install/pages/upgrade.adoc
```diff
@@ -76,19 +76,23 @@ For example, if you are upgrading three enterprise nodes (`*Node{nbsp}1*`, `*Nod
...
 
 | 6.x
-| 6.0 -> 6.5 -> 7.2x
+| 6.0 -> 6.6 -> 7.2.x -> 7.2.x (except 7.2.3) -> 7.2.3
 
 | 7.x
-| 7.0 -> 7.1 -> 7.2x
+| 7.0 -> 7.1 -> 7.2.3
```
timofey-barmin, line 89: There should be two options in case of 7.x:
```
7.0 -> 7.1 -> 7.2.3
7.0 -> 7.2.x (except 7.2.3) -> 7.2.3
```

modules/install/pages/upgrade.adoc
```diff
@@ -76,19 +76,23 @@ For example, if you are upgrading three enterprise nodes (`*Node{nbsp}1*`, `*Nod
...
+
+| 7.1.x
+| 7.2 -> 7.2.x -> 7.2.x (except 7.2.3) -> 7.2.3
```
timofey-barmin: 7.1.x is already covered by previous option (7.x) so I think this table row can be removed

modules/install/pages/upgrade.adoc
```diff
@@ -97,16 +101,16 @@ For example, if you are upgrading three enterprise nodes (`*Node{nbsp}1*`, `*Nod
...
-| 5.x -> 6.0 -> 6.5 -> 7.2x
+| 5.x -> 6.0 -> 6.6 -> 7.2.x (except 7.2.3) -> 7.2.3
 
 | 6.x
-| 6.0 -> 6.5 -> 7.2x
+| 6.0 -> 6.6 -> 7.2.x (except 7.2.3) -> 7.2.3
 
 | 7.x
-| 7.0 -> 7.1 -> 7.2x
+| 7.0 -> 7.1 -> 7.2.x (except 7.2.3) -> 7.2.3
```
timofey-barmin: When upgrading from 7.0 it is not necessary to go via 7.1. As well as it is not necessary to go via 7.2.x if you are already 7.1. So I think it should be changed to two options (same as for enterpise):

```
7.0 -> 7.1 -> 7.2.3
7.0 -> 7.2.x (except 7.2.3) -> 7.2.3
```

RayOffiah:
@timofey-barmin 

Ready for another review. 👍🏾

timofey-barmin (approved):
LGTM

PR #3374, opened 2023-10-13
Fix for DOC-14412 update Linux install package names 

ggray-cb:
Changed the pages mentioned in the bug, plus the Oracle Linux page which also had the old x86 package name.


bfavini (approved):
Looks good to me

tonyjhillman (approved):

PR #3377, opened 2023-10-23
Doc 11199 backup restore users and groups

ggray-cb:
Adds documentation for user and groups backup in 7.6 server.

This covers the following doc tickets:

- [DOC-11199](https://issues.couchbase.com/browse/DOC-11199) Document Backup and Restore Users and Groups
- [DOC-11279](https://issues.couchbase.com/browse/DOC-11279) Document backup and restore users REST API

Changes in this PR include:

- Added section to the cbbackupmgr page that covers what is actually backed up, including documenting the --disable-users and --overwrite-users flags. [See the preview output here](https://preview.docs-test.couchbase.com/doc-11199-backup-restore-users-and-groups/server/7.6/backup-restore/enterprise-backup-restore.html#whats-backed-up).
- Added REST API documentation for the backup endpoint. [Preview here.](https://preview.docs-test.couchbase.com/doc-11199-backup-restore-users-and-groups/server/7.6/rest-api/rbac.html#backup-and-restore-users-and-groups)
- Added link from the REST API overview doc to mention the ability to backup and restore users and groups using the RBAC API (see botoom of [this page](https://preview.docs-test.couchbase.com/doc-11199-backup-restore-users-and-groups/server/7.6/rest-api/backup-rest-api.html) fior a preview).
- Added two entries to the What's New page to cover the new cbbackupmgr and REST API features. [See preview here](https://preview.docs-test.couchbase.com/doc-11199-backup-restore-users-and-groups/server/7.6/introduction/whats-new.html),


modules/backup-restore/pages/enterprise-backup-restore.adoc
```diff
@@ -30,6 +30,40 @@ Each backup job in the [.term]_Backup Repository_ stores its backup in two ways:
 * All bucket data is stored in a small, secondary database.
 * All bucket creation scripts and configuration files are stored on the file system, as files.
 
+=== What's Backed Up
+
+By default, backups include your database's data and metadata. 
+Since Couchbase Server version 7.6 and later, `cbbackupmgr` also backs up users and groups.
```
hyunjuV: ... also backs up users and user groups.

(Just being specific about user groups so that the person does not think that we also backup LDAP groups -- in case the person reading just reads that sentence and stops there.)

modules/backup-restore/pages/enterprise-backup-restore.adoc
```diff
@@ -30,6 +30,40 @@ Each backup job in the [.term]_Backup Repository_ stores its backup in two ways:
...
+* user groups (Couchbase Server version 7.6 and later)
+* views
+
+Other flags let you exclude specific metadata, or select a subset of data to back up.
+See xref:backup-restore:cbbackupmgr-config.adoc[cbbackupmgr config] for a list of the arguments you can use to control what `cbbackupmgr` backs up.
+
+You can also use command line flags to control what data the `cbbackupmgr restore` command restores.  
+For example, use `--disable-users` to prevent `cbbackupmgr` from restoring users and groups. 
+You can also use the `--overwrite-users` to have `cbbackupmgr` overwrite existing users and groups in the database if the backup contains a matching user or group. 
+By default, `cbbackupmgr` does not overwrite existing users accounts in the database.
```
hyunjuV: By default, `cbbackupmgr` does not overwrite existing users in the database.

Instead of "users accounts", just use "users".  These are not accounts -- they are more like database credentials -- in Capella, they are just called database access names that you create inside a cluster.    Calling them "users accounts" may confuse some people (especially if they are also Capella users since in Capella, there are user accounts, separate from database access names).

modules/backup-restore/pages/enterprise-backup-restore.adoc
```diff
@@ -30,6 +30,40 @@ Each backup job in the [.term]_Backup Repository_ stores its backup in two ways:
...
+Instead, it restores just the user accounts in the backup that do not exist in database.
```
hyunjuV: Instead, it restores just the users in the backup that do not exist in database.

modules/introduction/partials/new-features-76.adoc
```diff
@@ -21,3 +21,6 @@ See xref:rest-api:rest-setting-security.adoc[Configure On-the-Wire Security].
 * Credentials for Couchbase-Server internal users can now be rotated at any time, by means of the REST API.
 See xref:rest-api:rest-rotate-internal-credentials.adoc[Rotate Internal Credentials].
 
+* By default, `cbbackupmgr` now backs up and restores groups and users, including roles and permissions. User passwords are saved in the backup in a hashed format. When restoring a backup, `cbbackupmgr` defaults to not overwriting existing users in the database with identically named users in the backup. You can change these default `cbbackupmgr` behaviors using the new `--disable-users` and `--overwrite-users` command-line arguments. See  xref:backup-restore:cbbackupmgr-config.adoc[cbbackupmgr config] and xref:backup-restore:cbbackupmgr-restore.adoc[cbbackupmgr restore] for more about these new arguments.
```
hyunjuV: ... now backs up and restores users and user groups.
Instead of "groups and users", use "users and user groups" -- since we don't backup LDAP groups, etc. -- if we jus say groups, people may think that we backup all types of groups.

modules/introduction/partials/new-features-76.adoc
```diff
@@ -21,3 +21,6 @@ See xref:rest-api:rest-setting-security.adoc[Configure On-the-Wire Security].
...
+
+* The Role-Based Access Control (RBAC) REST API has a new `backup` endpoint that lets you backup and restore user and groups. See xref:rest-api:rbac.adoc#backup-and-restore-users-and-groups[Backup and Restore Users and Groups].
```
hyunjuV: ... that lets you backup and restore users and user groups.

modules/rest-api/pages/backup-rest-api.adoc
```diff
@@ -18,3 +18,5 @@ Note that for all cluster references, in Couchbase Server Enterprise Edition Ver
 For a list of the methods and URIs covered in these pages, see the tables below.
 
 include::partial$rest-backup-service-table.adoc[]
+
+NOTE: You can use the xref:rest-api:rbac.adoc#backup-and-restore-users-and-groups[Role-Based Access Control (RBAC)] REST API to backup and restore just the users and groups defined in the database cluster.
```
hyunjuV: The rbac API's are not part of the Backup Service APIs.  So, they should not be mentioned here.
Line 22 should be removed.

This page documents Backup Service APIs -- i.e. the APIs that the UI uses for the Backup Service that the users can also use (if they don't want to use the UI).

Note:  The rbac endpoints are the Server endpoints that cbbackupmgr and the Backup Service (since the Backup Service uses cbbackupmgr) use to provide the user/user group backup/restore functionality.

So, if you are using the Backup Service, then, you are opting for the Backup Service to provide the backup/restore of the user/user groups, not to do it yourself.

modules/rest-api/pages/rbac.adoc
```diff
@@ -618,6 +619,130 @@ An attempt to add a user to one or more non-existent groups fails with `400 Bad
 
 The creation of an already existing group or local or external user succeeds with `200 OK`: the user is recreated with the newly specified role-assignments and group-memberships; and the group is recreated with the newly specified role-assignments and, optionally, a newly specified mapping.
```
hyunjuV, line 621: @timofey-barmin should review the correctness of the rbac endpoints here.

Added REST API documentation for the backup endpoint. [Preview here.](https://preview.docs-test.couchbase.com/doc-11199-backup-restore-users-and-groups/server/7.6/rest-api/rbac.html#backup-and-restore-users-and-groups)

hyunjuV (changes requested):

modules/rest-api/pages/rbac.adoc
```diff
@@ -618,6 +619,130 @@ An attempt to add a user to one or more non-existent groups fails with `400 Bad
...
+* `user:<local|external|\*>:<userWildcard>`: includes users. The `local`, `external`, or `*` expression sets the authentication domain to which the argument applies. The `userWildcard` wildcard expression determines which usernames the backup includes. For example, `include=user:local:tom`  includes just the user named `tom` defined in the local authentication domain. 
+* `permission:<permissionExpression>`: includes groups and users who have the permission in the `permissionExpression`. For example, `include=permission:cluster.bucket[travel-sample]!data_writer` includes any user or group that has data writer permissions on the `travel-sample` bucket.
+
+The `exclude` and `include` parameters are mutually exclusive. You cannot use both in the same method call. However, you can have multiple instances of one of the parameters in the method. For example, you use two instances of the `include` parameter in a single method call. The resulting backup contains the users and groups that match either `include` parameter. 
+
+The following example backs up all users defined in the local authentication domain and all groups.
```
timofey-barmin: Probably worth mentioning explicitly that it doesn't include the full administrator.

modules/rest-api/pages/rbac.adoc
```diff
@@ -618,6 +619,130 @@ An attempt to add a user to one or more non-existent groups fails with `400 Bad
...
 
+[backup-restore]
+== Backup and Restore Users and Groups
+
+Use the `/settings/rbac/backup` endpoint to back up and restore users and groups. 
+It is useful if you want to transfer or synchronize users between different database clusters. 
+
+=== Backup Users and Groups
+
+Use the `GET` method with the  `/settings/rbac/backup` endpoint to backup all users and groups in the database, including the Full Administrator account.
```
timofey-barmin: I think we want to make clear that the backup format is opaque and can be changed at any moment. Please also note that it is not even guaranteed that the backup is a JSON. I think it is fair to say that it is a string though, so I suggest we say that GET /settings/rbac/backup returns an opaque string.
The fact that we provide backup examples in documentation kind of makes it look like it is not opaque. So probably weshould remove the backup json examples as well.

modules/rest-api/pages/rbac.adoc
```diff
@@ -618,6 +619,130 @@ An attempt to add a user to one or more non-existent groups fails with `400 Bad
...
+----
+ curl -X PUT -u <administrator>:<password> 
+    http://<host>:<port>/settings/rbac/backup
+    --data-urlencode 'backup=<backupJSONData>''
+    -d canOverwrite=<true|false>
+----
+
+.Parameters
+
+backup=<backupJSONData>::
```
timofey-barmin: Since backup is opaque, we should not say it is JSON (please fix in all documentation)

modules/rest-api/pages/rbac.adoc
timofey-barmin: I think it makes sense to provide more examples.
 - Backup everything but admin (exclude=admin)
 - Backup only groups (include=group:*)
 - Backup all local users and admin (include=user:local:*&include=admin)

modules/rest-api/pages/rbac.adoc
```diff
@@ -618,6 +619,130 @@ An attempt to add a user to one or more non-existent groups fails with `400 Bad
...
++
+* `*`: includes everything. 
+* `admin`: includes only the Full Administrator
+* `group:<groupWildcard>`: includes only group definitions whose name matches the `groupWildcard` wildcard expression. For example, `include=group:analyst*` includes only groups whose name starts with the string `analyst`.
+* `user:<local|external|\*>:<userWildcard>`: includes users. The `local`, `external`, or `*` expression sets the authentication domain to which the argument applies. The `userWildcard` wildcard expression determines which usernames the backup includes. For example, `include=user:local:tom`  includes just the user named `tom` defined in the local authentication domain. 
+* `permission:<permissionExpression>`: includes groups and users who have the permission in the `permissionExpression`. For example, `include=permission:cluster.bucket[travel-sample]!data_writer` includes any user or group that has data writer permissions on the `travel-sample` bucket.
```
timofey-barmin: I think `cluster.bucket[travel-sample]!data_writer` is not a correct permission.
data_writer is a role, not a permission.
If you mean to write a permission that has access to bucket docs, I think it makes sense to write something like the following:
`cluster.collection[travel-sample:*:*].data.docs!any`

modules/rest-api/pages/rbac.adoc
ggray-cb: Should we be explcite in saying there's no guarantee that backups will work between versions (i.e. backup of users of a 7.6 cluster might not work on an 8.0 cluster)? Or  just leave it that he fornat is not guaranteed?

timofey-barmin: No, 

The fact that the format is opaque and can change from version to version doesn't mean that newer versions will not support old formats. 

Actually we should maintain backward compatibility for backups. At least for users and user groups there will be backward compatibility. I can't speak for other types of backups.

@hyunjuV 
Do we have any understanding currently for how long we should maintain backward compatibility for backup format?

hyunjuV: @ggray-cb 
We shouldn't say anything about future compatibility.   

We'll maintain backwards compatibility for backup formats for as long as possible.  This isn't something we should be documenting ahead of time (since we don't know ourselves and there's no reason to put ourselves into a corner).  When something changes, we'll document then.

cc @timofey-barmin

modules/backup-restore/pages/enterprise-backup-restore.adoc
ggray-cb: I beleive I've changed all instances of "groups" to "user groups" except for a ferw places in the APi docs where it seems redundant to resue the term multiple times in the same paragraph.

modules/rest-api/pages/backup-rest-api.adoc
ggray-cb: Removed.

modules/rest-api/pages/rbac.adoc
ggray-cb: Removed all instances of JSON, except for the return value from the PUT call, which I assume we want to state is JSON.

modules/rest-api/pages/rbac.adoc
ggray-cb: Added examples.

modules/rest-api/pages/rbac.adoc
ggray-cb: OK, changed to say we return an opaque string, with a clarification that the format for this string can change in the future (just in case anyone looks at the output and sees it is currently JSON).

modules/rest-api/pages/rbac.adoc
ggray-cb: Updated the example.

ggray-cb:
I believe I've addressed all of the feedback comments. Can you take a look and approve the PR?

timofey-barmin (approved):
LGTM (I reviewed only /settings/rbac/backup API documentation)

hyunjuV (approved):

PR #3381, opened 2023-10-31
DOC-11607: Text update for historyRetentionCollectionDefault

tonyjhillman:


modules/rest-api/pages/rest-bucket-create.adoc
```diff
@@ -702,8 +702,8 @@ The value can be either `true` (the default) or `false`.
 If the value is `true`, the change history records changes made to all collections within the bucket, unless this setting is overridden for one or more individual collections.
 (For details on per collection overriding, see xref:rest-api:creating-a-collection.adoc[Creating and Editing a Collection]).
 
-This parameter is ignored unless the value of `storageBackend` is specified as `magma`.
-This parameter is also ignored unless a positive value is specified for either `historyRetentionSeconds` or `historyRetentionBytes`, or both.
+If the value of `storageBackend` is not specified as `magma`, the request for a change history is rejected.
+The request is also rejected unless a positive value is specified for either `historyRetentionSeconds` or `historyRetentionBytes`, or both.
```
dnault: Line 705 looks good.

Line 706: A request with zero retention seconds & bytes is not rejected. Suggestion:

> Enabling change history has no effect unless a positive value is specified for either `historyRetentionSeconds` or `historyRetentionBytes`, or both.

tonyjhillman: Thanks, I've changed to this.

PR #3389, opened 2023-11-22
DOC-11544 Update upgrade instructions to reflect implications of Erlang and Open SSL upgrades

RayOffiah:
Version changed from 7.2.3 to 7.2.4


RayOffiah:
No need to commit this change until 7.2.4 Is released.

PR #3396, opened 2023-12-06
Doc 11557 user to dn mapping docs

ggray-cb:
Updates the LDAP sections of the documentaiton to handle the new "advanced" userToDnMapping option of using a list of regular expressions to map users to DNs.

Previews of changes pages:

* Manage / Manage Security / Manage Authentication / [Configure LDAP](https://preview.docs-test.couchbase.com/DOC-11557_UserToDnMapping_docs/server/7.6/manage/manage-security/configure-ldap.html#enable-ldap-user-authentication)
    * Updated affected screenshots.
    * Added Advanced heading to cover the new regular expression mapping feature.
    * Some minor edits to comply with style guide.
* Reference / REST API Reference / Security API / [Configure LDAP](https://preview.docs-test.couchbase.com/DOC-11557_UserToDnMapping_docs/server/7.6/rest-api/rest-configure-ldap.html). 
    * Added better description of what the arguments for the userToDnMapping parameter.
    * Linked back to the Manage section.


timofey-barmin (approved):
LGTM

bfavini (approved):

modules/manage/assets/images/manage-security/ldapAdvancedQuery.png
timofey-barmin, line 1: After taking another look, I think example on this screenshot needs to be changed. I suggest we change it.
1. First element username looks like the name of the service (service-.+), while we are usually not recommending using ldap for application users, so I suggest we change it to something that looks like a human email.
2. Second "match" doesn't have any re group (brackets), which is strange (not sure it will work).
3. Not clear what is rbsgrp.net. If it is a real website (it can become real in future), it could be problem, right?

ggray-cb: The example you are referring to is the placeholder text that shows up when you first enable the Advanced Query. So, if we want to change that in the screenshot, we need ot change it in the product. I'm fine with whatever it changes to.

timofey-barmin: Good point. I will work on the fix for the server.

timofey-barmin: The placeholder has been changed: https://review.couchbase.org/c/ns_server/+/202392

timofey-barmin (approved):
Thanks Gary, LGTM

PR #3397, opened 2023-12-11
DOC-10628: Remove incorrect information

ElliotFrancisHunter:
https://issues.couchbase.com/browse/DOC-10628

This PR addresses some information that is no longer correct, there will be similar PRs for all supported versions and 7.6.

In this page - [docs.couchbase.com/server/7.0/learn/services-and-indexes/indexes/storage-modes.html](https://docs.couchbase.com/server/7.0/learn/services-and-indexes/indexes/storage-modes.html#standard-index-storage) - in the box for enterprise edition, it is mentioned that "note that hole punching is required to enable auto-compaction of Global Secondary Indexes". This is incorrect. Plasma disk compaction is automatic and is enabled by default, irrespective of hole punching. Hole punching just allows plasma disk compaction to reclaim disk space at a finer granularity.


sarahlwelton (approved):
If you're sure that just removing this line, not editing, is sufficient, then happy to approve.

PR #3400, opened 2023-12-12
DOC-10628: remove linux caveat

ElliotFrancisHunter:
This is a PR for the ticket https://issues.couchbase.com/browse/DOC-10628
This affects 7.0, 7.1, 7.2 and 7.6.
The fix simply removes the caveat for Linux as there is no such caveat, the behaviour is uniform.

In this page - [docs.couchbase.com/server/7.0/learn/services-and-indexes/indexes/storage-modes.html](https://docs.couchbase.com/server/7.0/learn/services-and-indexes/indexes/storage-modes.html#standard-index-storage) - in the box for enterprise edition, it is mentioned that "note that hole punching is required to enable auto-compaction of Global Secondary Indexes". This is incorrect. Plasma disk compaction is automatic and is enabled by default, irrespective of hole punching. Hole punching just allows plasma disk compaction to reclaim disk space at a finer granularity.


sarahlwelton (approved):
Once again, looks fine

PR #3401, opened 2023-12-13
DOC-10867: Add use replica parameter

simon-dew:
Docs issue: DOC-10867

Added read replica parameters to cluster-level, node-level, and request-level settings and parameters. Also started improving the Settings and Parameters page following feedback.

Docs preview:

* [Cluster-level setting](https://ia.docs-test.couchbase.com/server/current/settings/query-settings.html#queryUseReplica)
* [Node-level setting](https://ia.docs-test.couchbase.com/server/current/settings/query-settings.html#use-replica-srv)
* [Request-level parameter](https://ia.docs-test.couchbase.com/server/current/settings/query-settings.html#use_replica_req)
* [Login details](https://couchbasecloud.atlassian.net/l/cp/2GtKM5Pr) &mdash; see IA Build

Linked pull requests:

* REST APIs: https://github.com/couchbaselabs/cb-swagger/pull/105


simon-dew:
- [x] Missed the bit about sequential scan

> Note that,even if use replica is turned on for a query, if the query uses the #sequentialScan index, and a data node goes down, the query might return with an error. This is because, currently, KV range scans cannot be started on replica v buckets.

simon-dew:
Adding @lamagnifica as a reviewer to take a look at the revamped [Named Parameters and Positional Parameters](https://ia.docs-test.couchbase.com/server/current/settings/query-settings.html#section_srh_tlm_n1b) section

dhanyagowrish (changes requested):
In the Cluster, Node, Request level descriptions for read from replica, could we please make a change here
Rather than  - 
_SELECT queries within a transaction cannot read from replica._

It should be - _Only SELECT queries that are not within a transaction can read from replica._

( Because the earlier statement implies that non-SELECT queries can read from replica. Which is not true. )

simon-dew:
> Only SELECT queries that are not within a transaction can read from replica

Done as part of [couchbaselabs/cb-swagger#105](https://github.com/couchbaselabs/cb-swagger/pull/105)

PR #3408, opened 2024-01-09
DOC-11737: Added misssing curl options to syntax

tonyjhillman:


sumukhbhat2701 (approved):
looks good to me.
I assume this change will be reflected in 7.6.0 docs when it releases as well? Thanks.

tonyjhillman:
Yes, it will appear in 7.6 and onwards.

PR #3414, opened 2024-01-11
Doc 11538 release note for 7.2.4

ggray-cb:
Second try at Release Notes PR,


ggray-cb:
Can't commit content into Ray's fork. Trying something else.

PR #3417, opened 2024-01-11
Removing the What's New in 7.2.4 heading because causes problems.

ggray-cb:
Removing the what's new in 7.2.4 heading


ggray-cb:
Irrelevant because the PruneAge branch has been unmerged. Will handle issue on that branch.

PR #3419, opened 2024-01-11
Doc 11652 prune audit logs

ggray-cb:
PR for pruneAge feature, which was pulled because it was merged too early. This PR will merge just before we publish 7.2.4.


bfavini (approved):
Approve

PR #3423, opened 2024-01-15
Doc 11731: Value of timeout can be 1 second

tonyjhillman:


modules/rest-api/pages/rest-cluster-autofailover-enable.adoc
```diff
@@ -47,6 +47,11 @@ Specifies the number of seconds that must elapse, with a node unavailable, befor
 Default setting is 120.
 The `timeout` parameter can only be specified when `enabled` is set to `true`.
 This parameter and its values are ignored if the value for the `enabled` parameter is `false`.
++
+The value of `timeout` can also be specified as 1 second.
+Note that this this low setting significantly increases the sensitivity of failure-detection
+and this, in turn, makes responses to _false failures_ more likely
+Additionally, CPU usage is likely to be considerably increased.
```
BenHuddleston: > Additionally, CPU usage is likely to be considerably increased.

I'm not sure if we should say "considerably" as it's quite subjective. Personally I'd suggest something like
"Additionally, more CPU resource will be consumed with lower failover timeout values (below 5 seconds). Testing a representative workload and measuring CPU usage and checking for false failovers is recommended".

modules/rest-api/pages/rest-cluster-autofailover-enable.adoc
```diff
@@ -47,6 +47,11 @@ Specifies the number of seconds that must elapse, with a node unavailable, befor
 Default setting is 120.
 The `timeout` parameter can only be specified when `enabled` is set to `true`.
 This parameter and its values are ignored if the value for the `enabled` parameter is `false`.
++
+The value of `timeout` can also be specified as 1 second.
```
BenHuddleston: I don't know if we would want to call out the difference in behaviour in the health monitoring between 1-5s timeout values and 5s-xyzs timeout values. Whilst we would see higher CPU usage between 1s and 5s (the lower we set the value, the higher the CPU usage will be), we should see no difference in CPU usage between 5s and 120s for example.

modules/rest-api/pages/rest-cluster-autofailover-enable.adoc
tonyjhillman: I've change the explanation, hopefully to include the concepts here. Let me know what you think.

modules/rest-api/pages/rest-cluster-autofailover-enable.adoc
tonyjhillman: Okay, I may not be reading the text in the same way, but I don't have the sense that we are making that comparison. Anyway, I've change the explanation,so that it's quite different now. So take a look at my revision and let me know what you think.

modules/rest-api/pages/rest-cluster-autofailover-enable.adoc
```diff
@@ -48,10 +48,16 @@ Default setting is 120.
...
 +
-The value of `timeout` can also be specified as 1 second.
-Note that this this low setting significantly increases the sensitivity of failure-detection
-and this, in turn, makes responses to _false failures_ more likely
-Additionally, CPU usage is likely to be considerably increased.
+The value of `timeout` can also be specified as `1` second.
+Note that low setting of this kind (anything below 5 seconds) significantly increases the sensitivity of failure-detection; and this, in turn, makes responses to _false failures_ more likely.
+Additionally, more CPU resources are consumed.
++
+Therefore, it is recommended that testing of a _representative workload_ should occur before the value of `timeout` is established as `1` for a production environment.
```
BenHuddleston, line 55: I'd replace
> before the value of `timeout` is established as `1`

with

> before a value of less than `5` is established for `timeout`

tonyjhillman: Thanks.

tonyjhillman: I've modified.

PR #3426, opened 2024-01-17
DOC-10720: Sort and add full KV metrics

malarky:
@osfameron As discussed, the belated sort and KV catch-up.

Left as two commits for clarity, please avoid the squash if possible, thanks!


osfameron (approved):

malarky:
Closing out to split in two...

PR #3435, opened 2024-01-24
DOC-9599: Update release note for MB-43232

simon-dew:
Server 6.5 is unsupported, but this may be useful for users of older versions or anyone who wants to know when the ticket was fixed. As the text was already provided it seems like a quick win.


simon-dew:
NB If there are no comments I'll merge this by close of play on 26 Jan

PR #3454, opened 2024-02-08
DOC-11673 Provide documentation for rebalance failure reason(s)

RayOffiah:
https://issues.couchbase.com/browse/DOC-11673

### DIFF
--- a/modules/rest-api/pages/rest-retrieve-cluster-rebalance-reason-codes.adoc
+++ b/modules/rest-api/pages/rest-retrieve-cluster-rebalance-reason-codes.adoc
@@ -1,0 +1,97 @@
+= Retrieving Cluster Rebalance Reason Codes
+
+:description: pass:q[If the cluster requires rebalancing, then the reason(s) for the rebalance can be ascertained using the `GET /pools/default` method.]
+:page-topic-type: reference
+
+[abstract]
+{description}
+
+== HTTP method and URI
+
+----
+GET /pools/default
+----
+
+[#rest-cluster-rebalance-reason-description]
+== Description
+
+Adding or removing one or more nodes from a cluster requires a rebalance for the cluster. Use this method and URI to retrieve the reason(s) for the rebalance.
+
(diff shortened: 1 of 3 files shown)


modules/rest-api/pages/rest-retrieve-cluster-rebalance-reason-codes.adoc
```diff
@@ -0,0 +1,97 @@
+= Retrieving Cluster Rebalance Reason Codes
+
+:description: pass:q[If the cluster requires rebalancing, then the reason(s) for the rebalance can be ascertained using the `GET /pools/default` method.]
```
bfavini, line 3: When the cluster needs to be rebalanced, you can retrieve the reason for the rebalance using the 'GET /pools/default' method.

modules/rest-api/pages/rest-retrieve-cluster-rebalance-reason-codes.adoc
```diff
@@ -0,0 +1,97 @@
...
+----
+GET /pools/default
+----
+
+[#rest-cluster-rebalance-reason-description]
+== Description
+
+Adding or removing one or more nodes from a cluster requires a rebalance for the cluster. Use this method and URI to retrieve the reason(s) for the rebalance.
+
+NOTE: The cluster only returns the reason codes if a rebalance is required.
```
bfavini, line 20: The cluster returns the reason code only if a rebalance is required.

bfavini, line 20: @RayOffiah I'm wondering about this...Is the purpose of this feature to find out whether a cluster needs a rebalance? Are we asking users to check for a rebalance code as a general health check on the cluster? If so, we might want to position this a little differently.

bfavini (approved):
I think this is fine--but we could probably use some clarification about where/why you use this (see my comment below.) Please do not hold up publishing for this...we could maybe just create a ticket to follow up.

PR #3455, opened 2024-02-08
Doc 11829 sequential scan role

ggray-cb:
Added new role to the roles page. [Preview is here](https://preview.docs-test.couchbase.com/DOC-11829_sequential_scan_role/server/7.6/learn/security/roles.html#query-sequential-scan).

### DIFF
--- a/modules/learn/pages/security/roles.adoc
+++ b/modules/learn/pages/security/roles.adoc
@@ -1947,12 +1947,11 @@ It allows the reading of bucket settings.
 [#query-sequential-scan]
 == Query Use Sequential Scan
 
-The *Query Use Sequential Scan* role (which is a _Query & Index_ role) allows users to perform a sequential scan query.
+The *Query Use Sequential Scan* role (a Query & Index role) allows users to perform a sequential scan query.
 This is a query of a keyspace that does not have a suitable index. 
 Couchbase Server limits sequential scan queries to users with this role because querying a large unindexed keyspace can be expensive.  
+This role does not grant the user the ability to read or mutate data or access to the Web Console. 
 Administrators automatically have permission to perform sequential scan queries.
-
-
 
 [#table_query_delete_role,cols="15,8,8,8,8",hrows=3]
 |===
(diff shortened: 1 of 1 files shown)


modules/learn/pages/security/roles.adoc
```diff
@@ -1943,6 +1943,59 @@ It allows the reading of bucket settings.
 ^| image:introduction/no.png[]
 |===
 
+
+[#query-sequential-scan]
+== Query Use Sequential Scan
+
+The *Query Use Sequential Scan* role (a Query & Index role) allows users to perform a sequential scan query.
```
dhaggart: It isn't "& Index". 
This reads as a type of query rather than an access method used in a query, to me.  I'd have something more like:
```
... allows users to execute a query that uses a sequential scan to access the keyspace.
```

modules/learn/pages/security/roles.adoc
```diff
@@ -1943,6 +1943,59 @@ It allows the reading of bucket settings.
...
+This is a query of a keyspace that does not have a suitable index.
```
dhaggart: The planner will only select a sequential scan as the access method when a suitable index is not available.

modules/learn/pages/security/roles.adoc
```diff
@@ -1943,6 +1943,59 @@ It allows the reading of bucket settings.
 ^| image:introduction/no.png[]
 |===
 
+
+[#query-sequential-scan]
+== Query Use Sequential Scan
+
+The *Query Use Sequential Scan* role (a Query & Index role) allows users to perform a sequential scan query.
+This is a query of a keyspace that does not have a suitable index. 
+Couchbase Server limits sequential scan queries to users with this role because querying a large unindexed keyspace can be expensive.
```
dhaggart: ...limits the use of sequential scans to access data to users...

dhaggart (commented):
Please see individual comments.

modules/learn/pages/security/roles.adoc
ggray-cb: "Query & Index" here is referring to where this role appears in the Web Console's  UI. You have to drop down the Query & Index category to find Query User Sequential Scans.

dhaggart: Ack.

modules/learn/pages/security/roles.adoc
```diff
@@ -1943,6 +1943,59 @@ It allows the reading of bucket settings.
...
+
+.2+^h| Resources
+4+^h| Privileges
+
+^h| *Read*
+^h| *Write*
+^h| *Execute*
+^h| *Manage*
+
+^| query_use_seqential_scans
```
simon-dew: Got to confess, I don't fully understand the layout of the tables on this page, but I think it looks inconsistent having the underlying role name listed here. The other tables use natural language descriptions in the Resources column. Maybe replace this text with "Sequential Scans" or "Query Sequential Scans"? The role name is given at the top of the table.

simon-dew (changes requested):
Mostly LGTM, just one comment

modules/learn/pages/security/roles.adoc
ggray-cb: Makes sense. I've made the change and updated the perview site.

modules/learn/pages/security/roles.adoc
ggray-cb: I changed it a bit to say that it appears under "Query & Index" in the Web Console's list of roles. Hopefully, that's a bit clearer than just saying it's a such-and-such role.

simon-dew (approved):
LGTM 👍

PR #3456, opened 2024-02-12
DOC-11673 Provide documentation for rebalance failure reasons

RayOffiah:
https://issues.couchbase.com/browse/DOC-11673

### DIFF
--- a/modules/rest-api/pages/rest-retrieve-cluster-rebalance-reason-codes.adoc
+++ b/modules/rest-api/pages/rest-retrieve-cluster-rebalance-reason-codes.adoc
@@ -1,0 +1,106 @@
+= Getting Cluster Rebalance Reason Codes
+
+:description: pass:q[When the cluster/bucket/service needs to be rebalanced, you can retrieve the reason for the rebalance using the 'GET /pools/default' method.]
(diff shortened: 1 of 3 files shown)


modules/rest-api/pages/rest-retrieve-cluster-rebalance-reason-codes.adoc
```diff
@@ -0,0 +1,97 @@
...
+== HTTP method and URI
+
+----
+GET /pools/default
+----
+
+[#rest-cluster-rebalance-reason-description]
+== Description
+
+Adding or removing one or more nodes from a cluster requires a rebalance for the cluster. Use this method and URI to retrieve the reason(s) for the rebalance.
```
bryandmc: To be fair, it's not just adding or removing nodes that could require a rebalance.. Maybe don't include that line and instead add something else. You could start instead with "Use this method..". Might be worth adding a little more to this main description as well. Largely you want to explain that 1.) at any point if you need a rebalance, we will include a more detailed reason as to why the rebalance is needed. 2.) it can be either a bucket or service that requires the rebalance or both.. obviously you can massage that all a bit but those are the main points.

PR #3477, opened 2024-03-01
Certificate changes for 7.6

ggray-cb:
This PR covers several certificate/encryption-related features:

* [MB-58276 Document removal of support for TLS1 and TLS1.1](https://issues.couchbase.com/browse/MB-58276)
* [DOC-11682 Document email / UI alert for certificate expiry](https://issues.couchbase.com/browse/DOC-11682)
* [DOC-11297 Add X.509 Elliptic Curve Keys support to the docs](https://issues.couchbase.com/browse/DOC-11297)
* [DOC-11280 Change list of default ciphers / explain change](https://issues.couchbase.com/browse/DOC-11280)
* [DOC-11071 Document support for PKCS12 format of cert](https://issues.couchbase.com/browse/DOC-11071)

Overview of changes, with links to the preview site:

* [What’s New in Version 7.6](https://preview.docs-test.couchbase.com/preview-certificate-changes-7.6/server/7.6/introduction/whats-new.html): 
  * Moved all security and authentication features to their own section (which will later be done to organize other content). 
  * Added entries for all certificate-related new features. Note that this is the only work done for DOC-11280 because there is no main documentation that covers the area covered by the changes.
* [Certificates](https://preview.docs-test.couchbase.com/preview-certificate-changes-7.6/server/7.6/learn/security/certificates.html): 
  * heavily edited this topic to comply with our documentation standards. Some content was removed because it wasn't relevant to this topic or was redundant. 
  * Added PKCS 12 support for node certificates. 
  * Added section on certificate expiration that mentions alerts for expiration and mentions the expiration can be changed.
  * Expanded section on private keys to discuss support for EC keys. 
* [Configure Server Certificates](https://preview.docs-test.couchbase.com/preview-certificate-changes-7.6/server/7.6/manage/manage-security/configure-server-certificates.html): also heavily edited for standards and clarity. Added section on using PKCS 12 certificates.
* [Upgrade](https://preview.docs-test.couchbase.com/preview-certificate-changes-7.6/server/7.6/install/upgrade.html): added a "Before You Upgrade" section and moved existing warning about .NET SDKs and mixed mode.  Added section about the removal of TLS 1 and 1.1.
* [On-the-Wire Security](https://preview.docs-test.couchbase.com/preview-certificate-changes-7.6/server/7.6/learn/security/on-the-wire-security.html): changes mainly for removasl of TLS 1 and 1.1.
* [Manage On-the-Wire Security](https://preview.docs-test.couchbase.com/preview-certificate-changes-7.6/server/7.6/manage/manage-security/manage-tls.html): standards and clarity edits, plus changes for TLS 1 and 1.1 removal.
* [Alerts](https://preview.docs-test.couchbase.com/preview-certificate-changes-7.6/server/7.6/manage/manage-settings/configure-alerts.html): Added certificate expiration to list of possible alerts. Also updated out-of-date screenshot.
* [Setting Alerts](https://preview.docs-test.couchbase.com/preview-certificate-changes-7.6/server/7.6/rest-cluster-email-notifications.html): fixed error in several examples. Updated some example output for limits. Added limit setting for expiration alerts.
* [Configure On-the-Wire Security](https://preview.docs-test.couchbase.com/preview-certificate-changes-7.6/server/7.6/rest-api/rest-intro.html): changes for TLS 1 and 1.1 removal, mainly around tlsMinVersion setting.


modules/introduction/partials/new-features-76.adoc
```diff
@@ -83,3 +74,44 @@ This new setting is useful if you want most of the documents in a bucket to auto
...
+See xref:rest-api:rest-setting-security.adoc[Configure On-the-Wire Security].
+
+* Credentials for Couchbase-Server internal users can now be rotated at any time, by means of the REST API.
+See xref:rest-api:rest-rotate-internal-credentials.adoc[Rotate Internal Credentials].
+
+* The Role-Based Access Control (RBAC) REST API has a new `backup` endpoint that lets you backup and restore user and user groups. 
+See xref:rest-api:rbac.adoc#backup-and-restore-users-and-groups[Backup and Restore Users and Groups]. 
+
+* You can now enable alerts for certificate expiration. 
+When enabled, Couchbase Server alerts you when server, node, or client certificates are within 30 days of expiration.
```
timofey-barmin: > when server, node, or client certificates are within 30 days of expiration. 

trusted certificates, node certificates, xdcr certificates

modules/learn/pages/security/certificates.adoc
```diff
@@ -1,342 +1,360 @@
...
+You may choose to replace these certificates to use ones signed by a well-known CA, or to add extensions to the certificates.
+Administrators must create all client certificates and certificate chains because Couchbase Server does not generate them by default.
 
-By default, Couchbase Server provides minimally defined _root_ and _entity_ certificates to protect cluster and individual nodes.
-Administrators can replace these, with customized or externally acquired certificates, and with corresponding certificate-chains.
-All certificates and certificate-chains for client-side use need to be explicitly created and/or deployed by administrators, as appropriate.
+The private key of the default self-signed cluster certificate is not available.
+Therefore, you cannot create custom certificate-chains based on the default certificate.
+To create custom certificate-chains, generate a custom cluster certificate and private key. 
+Another option is to get a root CA from an external authority and create intermediate certificates based on it, as described in xref:#intermediate-certificates[Intermediate Certificates].
```
timofey-barmin: > To create custom certificate-chains, generate a custom cluster certificate and private key. 
> Another option is to get a root CA from an external authority and create intermediate certificates based on it, as described in xref:#intermediate-certificates[Intermediate Certificates].

These two statements seem confusing to me. 
I find previous version less confusing:
`Administrators can replace default node certificates with customized or externally acquired certificates`
We also can add something like the following if you think it makes sense:
`The default self-signed root certificate can be removed from the trust store then.`

modules/learn/pages/security/certificates.adoc
```diff
@@ -1,342 +1,360 @@
...
 [#certificate-hierarchies]
-== Certificate Hierarchies
+=== Certificate Hierarchies
 
-On a Couchbase-Server cluster, certificates are managed and deployed _hierarchically_.
-The certificate with overall authority for the cluster is at the top of the hierarchy; and its authority is inherited by all certificates on lower tiers of the hierarchy.
-One certificate _issues_ another certificate &#8212; thereby letting the new certificate inherit its own authority and occupy the tier immediately below itself &#8212; by the process of _certificate signing_.
+A Couchbase Server cluster organizes certificates in a hierarchy.
+The certificate with overall authority for the cluster is at the top of the hierarchy.
+This certificate is the root certificate, also called the cluster certificate, because it is the certificate for the cluster as a whole.
```
timofey-barmin: The terminology has changed a bit since we introduced multiple ca support. Since you are rewriting it maybe we should not use obsolete terminology.
For example, we don't actually have cluster certificate anymore. We used to have a single root cert for the cluster, and all node certs were supposed to be signed by that cert, but it has changed.
Now every node can be issued by a separate ca and that will be fine. In this case it is not really clear what to call the cluster certificate. I suggest we get rid of this term as it seems confusing.

Instead of "the cluster certificate" we now have the trust store which is basically a list of root and intermediate certificates that are considered trusted by the cluster (it is cluster wide). The meaning of it is simple: we trust any certificate that was signed by a certificate from the trust store.
Basically any node certificate chain should end with a certificate from the trust store.

modules/learn/pages/security/certificates.adoc
```diff
@@ -1,342 +1,360 @@
...
+
+By default, Couchbase Server generates self-signed root and entity certificates to protect cluster and individual nodes.
+You can replace these certificates with customized or externally acquired certificates and with corresponding certificate-chains.
+You may choose to replace these certificates to use ones signed by a well-known CA, or to add extensions to the certificates.
+Administrators must create all client certificates and certificate chains because Couchbase Server does not generate them by default.
 
-By default, Couchbase Server provides minimally defined _root_ and _entity_ certificates to protect cluster and individual nodes.
-Administrators can replace these, with customized or externally acquired certificates, and with corresponding certificate-chains.
-All certificates and certificate-chains for client-side use need to be explicitly created and/or deployed by administrators, as appropriate.
+The private key of the default self-signed cluster certificate is not available.
```
timofey-barmin: > The private key of the default self-signed cluster certificate is not available.

I think it would be better to say "default self-signed root certificate"

modules/learn/pages/security/certificates.adoc
```diff
@@ -1,342 +1,360 @@
...
-Thus, if client and server share the same root certificate, they are likely to trust one another.
-
-Couchbase Server provides a default, self-signed cluster certificate, which is created and deployed when the first node in a cluster is created.
-A new, replacement cluster certificate can be deployed on the cluster by means of the Couchbase REST API or CLI.
-The current cluster certificate is always visible on the *Root Certificate* panel of the *Security* screen of Couchbase Web Console.
+Couchbase Server provides a default self-signed cluster certificate which it generates and deploys while creating the first node in a cluster.
+You can deploy a new replacement cluster certificate on the cluster using the Couchbase REST API or command line interface.
+For example, you can replace the certificate with one supplied by a well-known CA to make authentication with third parties easier.
+The current cluster certificate is always visible on the *Root Certificate* panel of the *Security* screen of Couchbase Server Web Console.
```
timofey-barmin: > Couchbase Server provides a default self-signed cluster certificate which it generates and deploys while creating the first node in a cluster.
You can deploy a new replacement cluster certificate on the cluster using the Couchbase REST API or command line interface.
For example, you can replace the certificate with one supplied by a well-known CA to make authentication with third parties easier.
The current cluster certificate is always visible on the *Root Certificate* panel of the *Security* screen of Couchbase Server Web Console.

This paragraph seems to be obsolete. There is no root certificate panel in UI anymore.
I think we can write about the trust store here instead.
Main points:
- This is a list of trusted certificates that is used for peer verification when (1) TLS connections are established between couchbase-nodes, and (2) when clients authenticate using client certificates.
- It is cluster wide
- the default self-signed root certificate (which is generated while creating the first node in a cluster) is added to the trust store automatically. It can be removed from the trust store if all nodes use custom certificates.
- Trusted certificates are visible on Certificates panel of the Security screen

modules/learn/pages/security/certificates.adoc
```diff
@@ -1,342 +1,360 @@
...
-Consequently, custom certificate-chains based on the default certificate _cannot_ be created.
-To create custom certificate-chains, a custom cluster certificate and private key must be generated (or alternatively, a Root CA must be acquired from an external authority, and certificate-chains then created based on an authority-signed _intermediate certificate_, as described below).
+You supply a certificate for the cluster in X.509 format. 
+The certificate must be in a Privacy Enhanced Mail (PEM) file. 
+These files have the extension `.pem`. 
+The file can have any name, although `ca.pem` is commonly used. 
+They contain a single certificate. 
+With this format, you supply a separate private key file in  PKCS #1 or PKCS #8 format, with a `.key` extension. 
+The name of the private key often corresponds to the name of the certificate it generates. 
+For example, `ca.key` would be used to generate `ca.pem`.
```
timofey-barmin: > You supply a certificate for the cluster in X.509 format. 
The certificate must be in a Privacy Enhanced Mail (PEM) file. 
These files have the extension `.pem`. 
The file can have any name, although `ca.pem` is commonly used. 
They contain a single certificate. 
With this format, you supply a separate private key file in  PKCS #1 or PKCS #8 format, with a `.key` extension. 
The name of the private key often corresponds to the name of the certificate it generates. 
For example, `ca.key` would be used to generate `ca.pem`.

Needs some changes.
- Trusted certificates can be added and removed from the trust store.
- Trusted certificates must be in a Privacy Enhanced Mail (PEM) format.

please remove:
- They contain a single certificate. 
- With this format, you supply a separate private key file in  PKCS #1 or PKCS #8 format, with a `.key` extension. 
- The name of the private key often corresponds to the name of the certificate it generates. 
- For example, `ca.key` would be used to generate `ca.pem`.

modules/learn/pages/security/certificates.adoc
```diff
@@ -1,342 +1,360 @@
...
 
-Examples of creating cluster and other certificates based on private keys are provided in xref:manage:manage-security/configure-server-certificates.adoc[Configure Server Certificates]
+For examples of creating cluster and other certificates based on private keys see in xref:manage:manage-security/configure-server-certificates.adoc[Configure Server Certificates]
 
-==== Multiple Root CAs
+=== Multiple Root CAs
```
timofey-barmin: >Multiple Root CAs

Probably can be combined with previous paragraph as we don't have the cluster certificate anymore.

modules/learn/pages/security/certificates.adoc
```diff
@@ -1,342 +1,360 @@
...
+Couchbase Server supports using multiple root CA certificates.
+Together, these certificates constitute a trust store, which Couchbase Server uses to determine whether or not to trust a client. 
+If the client's root CA in the client's trust chain is in the cluster's trust store, Couchbase Server trusts the client.
 
-The Couchbase-Server trust store includes one or more Root CAs that are used to provide authority to individual nodes: each node has at most one node certificate, potentially with a concatenated trust chain; and this points to a single Root CA in the cluster's trust store.
-A single Root CA may be pointed to by the trust chains of multiple nodes.
-All Root CAs in the trust store are accessible to all nodes, and multiple Root CAs may therefore be used in facilitating encrypted node-to-node communication.
+The Couchbase Server trust store provides authority to individual nodes.
+Each node has at most one node certificate, potentially with a concatenated trust chain.
+The certificate or its chain of trust  points to a single oot CA in the cluster's trust store.
```
timofey-barmin: > oot

a typo

modules/learn/pages/security/certificates.adoc
```diff
@@ -1,342 +1,360 @@
...
+The trust chains of multiple nodes can point to a single root CA.
```
timofey-barmin: > The trust chains of multiple nodes can point to a single root CA.

Probably worth adding that they can point to a single root CA but don't have to.

modules/learn/pages/security/certificates.adoc
```diff
@@ -1,342 +1,360 @@
...
 
-An _intermediate certificate_ (sometimes referred to as a _subordinate certificate_, an _Intermediate CA_, or a _Signing CA_) can be either:
+An intermediate certificate--sometimes referred to as a subordinate certificate, an intermediate CA, or a signing CA--can be either:
 
-* A certificate signed by the cluster certificate (the _Root CA_).
+* A certificate signed by the cluster certificate, which is often  called the root CA.
 
-* One of a chain of intermediate certificates, each of which has been signed by that preceding it in the chain; except the first, which has been signed by the cluster certificate.
+* One of a chain of intermediate certificates, each of which was signed by one preceding it in the chain. 
+The first certificate in the chain must be signed by the cluster certificate.
```
timofey-barmin: > The first certificate in the chain must be signed by the cluster certificate.

... must be signed by a root certificate that is in the trust store

modules/learn/pages/security/certificates.adoc
```diff
@@ -1,342 +1,360 @@
...
 
-The main purpose of the intermediate certificate is to sign _node certificates_, _client certificates_, or other _intermediate certificates_; and thereby convey the authority of the cluster certificate to the node or client certificates _indirectly_.
-This allows the cluster certificate's own private key to be only minimally used (and thereby more securely maintained) when multiple nodes or clients need to be signed &#8212; possibly across multiple clusters, subnets, or data centers.
+The main purpose of the intermediate certificate is to sign node certificates, client certificates, or other intermediate certificates. 
+Signing node certificates convey the authority of the cluster certificate to the node or client certificates indirectly.
```
timofey-barmin: "the cluster certificate" -> "the root certificate"

modules/learn/pages/security/certificates.adoc
```diff
@@ -1,342 +1,360 @@
...
+Using the intermediate certificate instead of the cluster certificate's own private key limits the key's use, making it more secure.
```
timofey-barmin: > Using the intermediate certificate instead of the cluster certificate's own private key limits the key's use, making it more secure.

Using the intermediate certificate instead of the root certificate limits the root private key's use, making it more secure.

modules/learn/pages/security/certificates.adoc
```diff
@@ -1,342 +1,360 @@
...
 
-The default certificates provided by Couchbase Server do not include intermediates: entity certificates are all signed directly by the root (the _cluster_ certificate).
-However, if customized certificates and certificate-chains are substituted by the administrator, intermediate certificates can be defined and used.
+The default certificates provided by Couchbase Server do not include intermediates: entity certificates are all signed directly by the cluster certificate.
```
timofey-barmin: > The default certificates provided by Couchbase Server do not include intermediates: entity certificates are all signed directly by the cluster certificate.

The default certificates provided by Couchbase Server do not include intermediates: entity certificates are all signed directly by the self-signed root certificate.

modules/learn/pages/security/certificates.adoc
```diff
@@ -1,342 +1,360 @@
...
-Certificate-based security for a Couchbase Server-cluster is thereby provided, in a limited form, _out-of-the-box_ (supporting, for example, all the standard Couchbase-Server secure ports &#8212; which are described in xref:install:install-ports.adoc[Couchbase Server Ports] &#8212;  and xref:learn:clusters-and-availability/node-to-node-encryption.adoc[Node-to-Node Encryption]).
-However, broader security requirements may need to be supported by means of _customized_ certificates, configured to include special extensions and _Subject Alternative Names_, based on an administrator-selected root authority.
+Each node has its own certificate  signed  by the cluster certificate. 
+Couchbase Server automatically generates the node certificates by creating a new private key and signing the certificate using cluster's certificate. 
+This process occurs whenever you create a single-node cluster and when you add additional nodes.
```
timofey-barmin: It looks confusing to me. I would rephrase it. Main points are the following:

> Each node has its own certificate  signed  by the cluster certificate. 

There is no such thing as cluster certificate.
I think it can be just:
Each node has its own server certificate.

> Couchbase Server automatically generates the node certificates by creating a new private key and signing the certificate using cluster's certificate. 
This process occurs whenever you create a single-node cluster and when you add additional nodes.
 
It happens only in the "by default" case. If customer uploads custom certs it is not happening
It is mentioned in the end of the paragraph but it is confusing. I think it should be mentioned in the beginning. Something like the following:

By default node certificate is generated at node's first start, and when the node joins a cluster.
The default OOTB node certificate is signed by the cluster's OOTB self-signed root certificate. 
Node can be provisioned with custom server certificate at any moment (before or after joining the cluster).
If node is provisioned with custom node certificate, no certificate is regenerated when that node joins the cluster.

modules/learn/pages/security/certificates.adoc
```diff
@@ -1,342 +1,360 @@
...
-Information included in the certificate identifies the user by means of a _username_.
+A client can use a client certificate to identify itself to Couchbase Server.
+The certificate allows the server to authenticate the client, and to authorize the client's associated user.
+Information included in the certificate identifies the user by means of a username.
 
-Couchbase Server creates and uses client certificates by default, for inter-node communication; but these are not visible to the user.
-Client certificates required for XDCR or SDK-client access must be explicitly created by the administrator; based on a customized, replacement cluster certificate.
+Couchbase Server creates and uses client certificates by default for inter-node communication.
+These certificates are not visible to the user.
+Administrators must create the client certificates for XDCR or SDK-client connections explicitly based on a customized replacement cluster certificate.
```
timofey-barmin: > Administrators must create the client certificates for XDCR or SDK-client connections explicitly based on a customized replacement cluster certificate.

the ending looks confusing I suggest removing it:
Administrators must create the client certificates for XDCR or SDK-client connections explicitly.

modules/learn/pages/security/certificates.adoc
```diff
@@ -1,342 +1,360 @@
...
 
-When authenticating a client that uses certificate-based authentication, Couchbase Server asks the client to present the client certificate.
-Couchbase Server inspects the client certificate: if this is ascertained to be part of a chain that leads to a root authority that is recognized by Couchbase Server, the client may be trusted.
-The certificate's time-validity and other details are checked.
-If the certificate has not expired and is valid in all other necessary respects, the _username_ provided by the certificate is determined, and this is checked by Couchbase Server against registered users and their roles.
-If the user exists, and the associated roles are appropriate, access is granted; otherwise, access is denied.
+When authenticating a client that uses certificate-based authentication, Couchbase Server asks the client to present its certificate.
+If Couchbase Server finds that the certificate's chain of trust leads to a root authority that it recognizes, it trusts the client.
```
timofey-barmin, line 191: > If Couchbase Server finds that the certificate's chain of trust leads to a root authority that it recognizes, it trusts the client.

I think we should explicitly say that root certificate should be in the trust store.

modules/learn/pages/security/on-the-wire-security.adoc
```diff
@@ -63,18 +63,23 @@ A _service_ can be _any_ of the Couchbase Server xref:learn:services-and-indexes
...
 
-The minimum version can be established globally and for all services as `tlsv1`, `tlsv1.1`, `tlsv1.2` or `tlsv1.3`.
-Note, however, that `tlsv1` and `tlsv1.1` are _deprecated_ in Couchbase Server Version 7.2.
-Also, see xref:learn:security/on-the-wire-security.adoc#cipher-suite-configuration-limitations[Cipher-Suite Configuration Limitations], below, for further information.
+IMPORTANT: Couchbase Server 7.6 and later do not support TLS versions 1 and 1.1. 
+Support for these earlier versions was deprecated in Couchbase Server 7.2. 
+When upgrading from a pre-7.6 version of Couchbase Server to version 7.6 or later, the upgrade process sets any TLS minimum setting that's lower than 1.2 to 1.2.
 
-The cluster-wide default value for the minimum TLS version is `tlsv1.2`: this is used by a service if no value has been specified for that service, and no global value has been specified.
+Couchbase Server defaults to requiring TLS version 1.2 or later if you have set Couchbase Server to enforce TLS.
```
timofey-barmin, line 76: > Couchbase Server defaults to requiring TLS version 1.2 or later if you have set Couchbase Server to enforce TLS. 

Seems like "if you have set Couchbase Server to enforce TLS." is not necessary.

modules/manage/assets/images/manage-settings/emailAlertsScreenInitial.png
timofey-barmin, line 1: New screenshot is much much smaller. Is this intended?

modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -1,142 +1,86 @@
...
-Modification of the procedures will likely be required, for the preparation of certificates for different platforms and cluster-configurations.
+The procedures in this page are only limited examples.
+They cover the basic steps for creating certificates.
+When creating and deploying certificates for your own database, you often have to modify these steps to suit your environment.
 
-Two procedures are provided, each of which configures X.509 certificates on Ubuntu 18 for a one-node Couchbase Server-cluster.
-Before attempting to follow either procedure, see the conceptual and architectural information provided in xref:learn:security/certificates.adoc[Certificates].
+This page gives detailed steps to configure X.509 certificates on a Linux-based single node Couchbase Server. 
+It demonstrates two scenarios. 
+The first shows directly signing the node's certificate using the cluster's certificate.
```
timofey-barmin: "the cluster's certificate. " -> "the root certificate. "

modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -1,142 +1,86 @@
...
+The procedures in this page are only limited examples.
+They cover the basic steps for creating certificates.
+When creating and deploying certificates for your own database, you often have to modify these steps to suit your environment.
 
-Two procedures are provided, each of which configures X.509 certificates on Ubuntu 18 for a one-node Couchbase Server-cluster.
-Before attempting to follow either procedure, see the conceptual and architectural information provided in xref:learn:security/certificates.adoc[Certificates].
+This page gives detailed steps to configure X.509 certificates on a Linux-based single node Couchbase Server. 
+It demonstrates two scenarios. 
+The first shows directly signing the node's certificate using the cluster's certificate. 
+The second shows creating an intermediate certificate from the cluster certificate and using that to sign the node's certificate.
```
timofey-barmin: "the cluster certificate" -> "root certificate"

modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -241,175 +147,229 @@ keyUsage = digitalSignature,keyEncipherment
...
-The creation of the customized extensions file should occur once for each node, with each customized extensions file containing only those extensions that apply to the current node.
+NOTE: Couchbase Enterprise Server Version 7.2 and later requires that the node's certificate identifies the node in a Subject Alternative Name extension.
+Without this identification, Couchbase Server reports an error when you upload the certificate to the node or when you try to add the node to the cluster. 
+For more information, see xref:learn:security/certificates.adoc#node-certificate-validation[Node-Certificate Validation].
 
-. Create the node certificate, applying the certificate and digital signature of the appropriate authority, and the customized extensions file for the node, to the materials in the signing request.
-+
-Enter the following:
+. Create the node's certificate by signing it with the certificate and digital signature of the CA.
+In this example, the CA is the cluster, so the command to sign the node's certificate uses the cluster's key and certificate:
```
timofey-barmin: > In this example, the CA is the cluster, so the command to sign the node's certificate uses the cluster's key and certificate:

Not sure about "CA is the cluster". Rather the one who is generating the certs is CA. The cluster has nothing to do with it.

modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -241,175 +147,229 @@ keyUsage = digitalSignature,keyEncipherment
...
+For more information, see xref:learn:security/using-multiple-cas.adoc#adding-intermediate-certificates-to-the-trust-store[Adding Intermediate Certificates to the Trust Store].
 
-* Upload, from the node's inbox, of a _chain.pem_ file containing only the unconcatenated node certificate; with the expectation that each intermediate certificate resides in the client's _trust store_.
+When a client attempts to connect to the database securely, it uses the node's certificate to verify the node's identity.
+For it to verify the node identity, the client searches for a CA it trusts in the chain of certificates from the node, through any intermediates, to the cluster certificate.
+See xref:learn:security/certificates.adoc#intermediate-certificates[Intermediate Certificates] for more information.
 
-Both procedures are shown below.
-For more information, see xref:learn:security/using-multiple-cas.adoc#adding-intermediate-certificates-to-the-trust-store[Adding Intermediate Certificates to the Trust Store].
+In Couchbase Server Version 7.1 and later, you can supply the client with the chain of trust it requires in one of two ways:
```
timofey-barmin: > In Couchbase Server Version 7.1 and later, you can supply the client with the chain of trust it requires in one of two ways:

Seems like the word "client" is incorrect in this context. I think it should be "cluster node" instead.

modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -241,175 +147,229 @@ keyUsage = digitalSignature,keyEncipherment
...
 
-The steps and descriptions in the procedures below assume that the previous procedure, xref:manage:manage-security/configure-server-certificates.adoc#root-and-node-certificates[Cluster Protection with Root and Node Certificates], has already been successfully completed; and that familiarity with the basic certificate-related concepts explained there has been attained.
+* Concatenation of all intermediate and node certificates into a single `chain.pem` file, which you deploy to tne node.
```
timofey-barmin: > tne

typo

modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -241,175 +147,229 @@ keyUsage = digitalSignature,keyEncipherment
...
+The node provides this entire chain of trust to the client when it tries to connect securely.
 
-Note that corresponding, subsequent procedures that create certificates for _client_-authentication are provided in xref:manage:manage-security/configure-client-certificates.adoc#client-certificate-authorized-by-an-intermediate-certificate[Client Access: Intermediate Certificate Authorization] and xref:manage:manage-security/configure-client-certificates.adoc#java-client-access-intermediate-certificate-authorization[Java Client Access: Intermediate Certificate Authorization]
+* Deploy a `chain.pem` file containing just the node's certificate. 
+In this case, the client's trust store must already have all intermediate certificates that it needs to verify the node's identity.
```
timofey-barmin: > In this case, the client's trust store must already have all intermediate certificates that it needs to verify the node's identity.

"client's trust store" should be "cluster's trust store" in this context.

modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -527,124 +488,139 @@ keyUsage = digitalSignature,keyEncipherment
...
-curl -X POST http://10.143.192.102:8091/node/controller/reloadCertificate -u Administrator:password
+curl -X POST http://10.143.192.102:8091/node/controller/loadTrustedCAs \
+     -u Administrator:password
 ----
-+
-Note that when, as is typical, the cluster contains more than one node, the `/node/controller/reloadCertificate` command must be executed on each node, specifying the IP address of the node on which execution is occurring.
 
-This concludes the certificate-deployment process.
-The root certificate can be examined by means of Couchbase Web Console, as shown in xref:manage:manage-security/configure-server-certificates.adoc#see-root-certificate-with-couchbase-web-console[Step 13] of the previous example on this page.
+. Upload the node certificate, specifying the established password for the private key.
```
timofey-barmin: >  Upload the node certificate, specifying the established password for the private key.

Should be just "Upload the node certificate"

This specific example is not passing the private key passphrase. `-u Administrator:password` are credentials for the http request.

It seems that the node's private key is not encrypted in this case, so no password is needed.

modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -527,124 +488,139 @@ keyUsage = digitalSignature,keyEncipherment
...
+----
+curl -X POST http://10.143.192.102:8091/node/controller/reloadCertificate \
+    -u Administrator:password
+----
 
 For more information using the REST API to manage certificates, see xref:rest-api:rest-certificate-management.adoc[Certificate Management API].
 
 [#intermediate-upload]
-=== Deploying an Intermediate Certificate via Trust Store
+=== Deploy an Intermediate Certificate via Client Trust Store
```
timofey-barmin: Should be "Deploy an Intermediate Certificate via Cluster's Trust Store"

modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -527,124 +488,139 @@ keyUsage = digitalSignature,keyEncipherment
...
 
-Proceed as follows:
+The following example creates an intermediate certificate but does not concatenate it with the node's certificate. 
+After following these steps, you must make sure the client's certificate contains a chain of trust it can use to verify the node's identity.
```
timofey-barmin: > After following these steps, you must make sure the client's certificate contains a chain of trust it can use to verify the node's identity. 

Not sure I understand this phrase.
Probably it should be something like the following:
After following these steps, you must make sure the cluster's trust store contains the root and intermediate certificates it can use to verify the node's identity.

modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -675,96 +648,218 @@ cd ../root
...
 . Upload the root and intermediate certificates.
 +
+[source, console]
 ----
-curl -X POST http://10.143.192.102:8091/node/controller/loadTrustedCAs -u Administrator:password
+curl -X POST http://10.143.192.102:8091/node/controller/loadTrustedCAs \
+     -u Administrator:password
 ----
 
 . Upload the node certificate, specifying the established password for the private key.
```
timofey-barmin: As in the comment above, please remove "specifying the established password for the private key".

modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -675,96 +648,218 @@ cd ../root
...
+Also, copy the files to the node's inbox on its own filesystem.
+The files must be on the node for the REST API call to work.
 
-Note that when, as is typical, the cluster contains more than one node, the `/node/controller/reloadCertificate` command must be executed on each node, specifying the IP address of the node on which execution is occurring.
-
-This concludes the certificate-deployment process.
-The root certificate can be examined by means of Couchbase Web Console, as shown in xref:#see-root-certificate-with-couchbase-web-console[Step 13] of the previous example on this page.
+The node's certificate is now deployed.
+The node's certificate does not contain a chain of trust that includes the intermediate certificate.
+You must make sure each client certificate does have a copy.
```
timofey-barmin: > The node's certificate does not contain a chain of trust that includes the intermediate certificate.
You must make sure each client certificate does have a copy. 

Rather something like the following:
The node's certificate chain does not contain the intermediate certificate. When a cluster node verifies the chain of trust of another node, the intermediate certificate will be taken from the cluster's trust store now.

Please also remove "You must make sure each client certificate does have a copy. "

modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -675,96 +648,218 @@ cd ../root
...
-See the reference page xref:rest-api/upload-retrieve-node-cert.adoc[Upload and Retrieve a Node Certificate].
+PKCS #12 format certificates let you bundle certificates, private keys, and other objects into a single file.
+Couchbase Server supports using PKCS #12 files for deploying certificates, private keys, and chains of trust for nodes.
+It does not support using them for other purposes, such as client or cluster certificates. 
 
-[#configure-client-access-advanced]
-=== Configuring Client Access
+Couchbase Server imposes several requirements on using PKCS #12 files for node certificate deployment:
+
+* You must set a pass phrase on the PKCS #12 file you use to deploy certificates and keys.
```
timofey-barmin: Strictly speaking the pkcs12 file doesn't have to be password protected. So I am not sure if we can say it is a "must".

modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -675,96 +648,218 @@ cd ../root
...
+PKCS #12 format certificates let you bundle certificates, private keys, and other objects into a single file.
+Couchbase Server supports using PKCS #12 files for deploying certificates, private keys, and chains of trust for nodes.
+It does not support using them for other purposes, such as client or cluster certificates. 
 
-[#configure-client-access-advanced]
-=== Configuring Client Access
+Couchbase Server imposes several requirements on using PKCS #12 files for node certificate deployment:
+
+* You must set a pass phrase on the PKCS #12 file you use to deploy certificates and keys.   
+* You must pass the PKCS #12 file's passphrase as an argument to the REST API call you use to deploy the certificate.
```
timofey-barmin: Passing passphrase as plaintext is only one of the options (and not the most secure option).
Please check out https://docs.couchbase.com/server/current/rest-api/upload-retrieve-node-cert.html#json-passphrase-registration
Basically user needs to provide information on how to get the password or the password itself. And only if pkcs12 is password protected.

modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -675,96 +648,218 @@ cd ../root
...
+----
++
 
-Once the cluster has been protected by the deployment of root, intermediate, and node certificates described above, a _client_ certificate can be signed by a _client-intermediate_ certificate that itself inherits the authority of the root: this allows the client certificate to access the cluster.
-Client-certificate preparation varies, depending on the type of client to be supported.
-For steps to prepare a certificate supportive of Couchbase Server, see xref:manage:manage-security/configure-client-certificates.adoc#client-certificate-authorized-by-an-intermediate-certificate[Client Access: Intermediate-Certificate Authorization].
-For steps to prepare a certificate supportive of a Java client, see xref:manage:manage-security/configure-client-certificates.adoc#java-client-access-intermediate-certificate-authorization[Java Client Access: Intermediate-Certificate Authorization].
+. Load the node certificate and its private key by calling the xref:rest-api:upload-retrieve-node-cert.adoc[reloadCertificate] REST API. 
+You must pass the password for the PKCS #12 certificate as an argument to the REST API call:
```
timofey-barmin: Please mention that example uses plaintext passphrase for simplicity. We can also add a link to https://docs.couchbase.com/server/current/rest-api/upload-retrieve-node-cert.html#json-passphrase-registration
here (other options to specify the passphrase)

modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -675,96 +648,218 @@ cd ../root
...
++
+The JSON value you pass to the command supplies the password for the private key in the PKCS #12 certificate as plain text.
+Replace the `private-key-password` with the password you entered in step 2. 
 
-Note that access by means of a client certificate must be specifically enabled, on the cluster that is to be accessed: see xref:manage:manage-security/enable-client-certificate-handling.adoc[Enable Client-Certificate Handling].
+Couchbase Server extracts the private key and certificate from the `couchbase.p12` file and activates them on the node.
 
-[#protection-of-multi-node-clusters]
-== Protection of Multi-Node Clusters
+This example has the node's certificate directly signed by the cluster certificate.
```
timofey-barmin: "the cluster certificate" -> "the root certificate"

modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -675,96 +648,218 @@ cd ../root
...
+[#using-an-externally-provided-root-certificate]
+== Using an Externally Provided Root Certificate
 
-Note also that when a cluster is using uploaded certificates, and a new node is added or joined to a cluster, the operation is performed with reference to a particular node that is already a member of the cluster.
-Since this _cluster node_ and this _new node_ must be able to connect with one another, each must trust the CA of the other.
-Therefore:
+The examples in this page create a self-signed root certificate use that certificate's private key to sign other certificates.
+In production environments, you often want to use a root certificate signed by a well-known Certificate Authority.
+In this case, instead of creating the cluster certificate yourself, you will use the certificate signed by the external CA to sign certificates.
+You also often create intermediate certificates to sign the node and client certificates instead of directly signing them using the cluster certificate.
```
timofey-barmin: > In production environments, you often want to use a root certificate signed by a well-known Certificate Authority.

In production environments, you often want to use a node certificate issued by a well-known Certificate Authority.

> In this case, instead of creating the cluster certificate yourself, you will use the certificate signed by the external CA to sign certificates.

In this case, the CA will provide the root, intermediate, and node certificates for you. The intermediate certificate is optional though.

> You also often create intermediate certificates to sign the node and client certificates instead of directly signing them using the cluster certificate.

Please remove.

modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -675,96 +648,218 @@ cd ../root
...
 
-* If the two nodes already have the same CA, connection and node-addition (or node-joining) can occur.
+When a cluster uses the default auto-generated certificates, you do not need to generate a new certificate for new nodes.
+Once you configure the cluster to use custom certificates, you must generate a new certificate when adding or joining new nodes to the cluster.
+In Couchbase Server Version 7.1 always adds or joins new nodes  over an encrypted connection.
 
-* If the two nodes do _not_ have the same CA, the CA of each must be loaded onto the other; after which, connection and node-addition (or node-joining) can occur.
+When a cluster using custom certificates adds or joins a new node to itself, the new node must interact with an existing node.
+This interaction requires both the existing node and the new node to verify each other's identity using their chains of trust.
+Therefore, certificates of the existing and new node must have a common CA in their trust chain.
```
timofey-barmin: > Therefore, certificates of the existing and new node must have a common CA in their trust chain.

They don't have to have common CA.
But rather cluster and the-node-to-be-added should have each others root (CA) certificates added to their trust store. Basically they should trust each other.
Say A joins B. Then A should have B's root certificate added to A's trusted store. And B should have A's root certificate added to B's trusted store. 
In most cases it is satisfied automatically because usually all certs are issued by the same CA (so root cert is the same). But in some cases CA can be different (CA certs are rotated periodically, users also can switch from one CA to abother, etc...)

modules/manage/pages/manage-security/configure-server-certificates.adoc
```diff
@@ -675,96 +648,218 @@ cd ../root
...
+In most cases, you make sure the nodes can identify  each other by signing them with the cluster certificate or the same intermediate certificate.
+If, for some reason, you cannot use the same certificate to sign the node's certificates, you must concatenate the intermediate certificates to each node's certificate.
+This concatenation make sure the nodes can find a common CA in their chain of trust to identify each other.
```
timofey-barmin: > If, for some reason, you cannot use the same certificate to sign the node's certificates, you must concatenate the intermediate certificates to each node's certificate.
This concatenation make sure the nodes can find a common CA in their chain of trust to identify each other.

I am not sure I understand these phrases. Probably we need to discuss it.
Or we can remove it.

modules/manage/pages/manage-security/manage-tls.adoc
```diff
@@ -116,11 +118,11 @@ The Couchbase CLI allows cluster-wide on-the-wire security-settings to be retrie
 [#get-cluster-wide-settings-with-the-cli]
 === Get Cluster-Wide Settings, with the CLI
 
-The _cluster-wide_ security configuration includes both _global_ and _per service_ settings.
-The current configuration can be retrieved with the xref:cli:cbcli/couchbase-cli-setting-security.adoc[setting-security] command.
-This is shown by the following example.
-Note that the output is piped to the https://stedolan.github.io/jq[jq^] command, to optimise the output's readability:
+The cluster-wide security configuration includes both global and per service settings.
+You can view the current configuration using the xref:cli:cbcli/couchbase-cli-setting-security.adoc[setting-security] command.
+The following example pipes thsi command's output to the https://stedolan.github.io/jq[jq^] command to make the output more readable:
```
timofey-barmin: > thsi

typo

modules/manage/pages/manage-settings/configure-alerts.adoc
```diff
@@ -180,6 +180,14 @@ With no memory left for storing metadata, further requests to store data will al
 | The disk usage worker is stuck and unresponsive.
 | `disk_usage_analyzer_stuck`
 
+| Certificate has expired
+| A security certificate for the cluster, a node, or a client has expired.
```
timofey-barmin: >  cluster, a node, or a client

root, node, or xdcr certificate has expired

modules/manage/pages/manage-settings/configure-alerts.adoc
```diff
@@ -180,6 +180,14 @@ With no memory left for storing metadata, further requests to store data will al
 | The disk usage worker is stuck and unresponsive.
 | `disk_usage_analyzer_stuck`
 
+| Certificate has expired
+| A security certificate for the cluster, a node, or a client has expired. 
+| `cert_expired`
+
+| Certificate will expire soon
+| A security certificate for the cluster, a node, or a client expires within 30 days.
```
timofey-barmin: >  cluster, a node, or a client

root, node, or xdcr certificate expires within 30 days

bfavini (approved):
Great job! Looks good.

modules/introduction/partials/new-features-76.adoc
ggray-cb: Not clear on this comment... do expiring client certificates cause an alert, or is it just the three you mention?

timofey-barmin: Just the three I mentioned.

"client certificates" in this context is a bit too generic. For example, xdcr certificate is a client certificate.
Certificates that are used by SDKs for authentication are also client certificates (we don't send any alerts about this kind of certs), etc...

modules/manage/pages/manage-security/configure-server-certificates.adoc
ggray-cb: Do clients not use the node's certificate to determine the identity of the node? If so, I would think that this discussion applies to both other nodes and clients. What if I use the term "peer" to indicat ethis can be another node, a client, etc?

modules/manage/pages/manage-security/configure-server-certificates.adoc
ggray-cb: I think the idea behind this example is that you can use the intermediate certificate to sign the node's certificate and rely on any peer (another node or some client) that needs to indentify the node will have the intermediate certificate in its trust store. If we're just concerned about the nodes identifying each other, it makes sense to call it Deploy an Intermediate Certificate via Cluster's Trust Store. But then I would womnder why we need both of these examples. It's probably best to show them a single method of using an intermedaite certificate rather than have two slightly different examples.

modules/manage/pages/manage-security/configure-server-certificates.adoc
timofey-barmin: > Do clients not use the node's certificate to determine the identity of the node?

In general all components can use certificates for identity verification. Here we are talking about node configuration, right?
Also client is a very generic word and it can be very confusing. Not clear what is meant in this case. Is that SDK instance? or is that TLS client?

Since we are talking about node certificates here, and chain and trusted certificates are uploaded to nodes I think it is less confusing to just say " you can supply the node with the chain of trust with it requires in one of the following ways:"

modules/manage/pages/manage-security/configure-server-certificates.adoc
timofey-barmin: I think you are right here. We probably can make it less confusing by something like this:
```
You must make sure any TLS client connecting to that node has that intermediate
certificate added to its trusted store. Since cluster node can also play a role of TLS client,
intermediate certificate must be added to the server's trusted store as well
```

modules/manage/pages/manage-security/configure-server-certificates.adoc
timofey-barmin: Actually you changed my mind about that, we probably can use peer here. Because in some cases it is used on the client side, sometimes on the server side. In some cases it can be SDKs, in some cases it is server node. Yes, I agree peer should be ok.

bryandmc (approved):
So the sections I looked at specifically were the ones regarding change of ciphers, change of available tlsMinVersion numbers and gave the rest a quick read and it seems fine to me. Please have other's finish their reviews since I didn't go into detail into sections I didn't cover originally.

PR #3483, opened 2024-03-06
Update rest-cluster-email-notifications.adoc

tonklon:
I don't see this parameter is optional. If not included with a POST request, a 400 Bad Request is returned. Also all the examples include a value for the enabled parameter.


ggray-cb (approved):
Verified that the parameter is required (8.0 returns a specific error stating enabled must be supplied.

PR #3502, opened 2024-03-12
MB-61081 remove _eventing _transactions collections

osfameron:


anuthan (commented):
LGTM. Thanks for quick turnaround.

PR #3512, opened 2024-03-14
DOC-11705: Update XDCR Metrics Labels

malarky:
@osfameron Here's the PR for [DOC-11705](https://issues.couchbase.com/browse/DOC-11705), should apply cleanly to the current 7.6 metrics too.


osfameron (approved):
👍

PR #3531, opened 2024-03-18
[DOC-11973] Removed line about NRU

julia-cb:
Jira ticket: https://issues.couchbase.com/browse/DOC-11973

Change in this PR:
- Removed line about NRU from https://docs.couchbase.com/server/current/learn/buckets-memory-and-storage/buckets.html (Couchbase doesn't use NRU)

At some point, when we have more bandwidth, we should aim to document that buckets use the MFU algorithm to determine what to eject (more info here https://github.com/couchbase/kv_engine/blob/v7.2.0/docs/HiFi_MFU_eviction_policy.md)


julia-cb:
@jimwwalker Hi Jim, I went ahead and removed the NRU line from the page. I'll open a ticket for myself to document MRU in more detail later, but for now I think just the removal should be ok. Thanks! :)

ggray-cb (approved):

PR #3536, opened 2024-03-20
KV Range Scan + Query without index

RichardSmedley:


modules/introduction/partials/new-features-76.adoc
```diff
@@ -158,6 +158,12 @@ For more information on the vector search, see xref:vector-search:vector-search.
 
 === Data Service
 
+* Couchbase Server 7.6 introduces KV Range Scan.
+A method for accessing documents without necessarily knowing the IDs (keys).
```
bfavini: KV Range Scan is a method for accessing...

modules/introduction/partials/new-features-76.adoc
```diff
@@ -158,6 +158,12 @@ For more information on the vector search, see xref:vector-search:vector-search.
...
+The most common use case is to get all documents whose IDs start with the same prefix, a `prefixScan`,
```
bfavini: The most common use case is to retrieve all documents

modules/introduction/partials/new-features-76.adoc
```diff
@@ -158,6 +158,11 @@ For more information on the vector search, see xref:vector-search:vector-search.
 
 === Data Service
 
+* KV Range Scan is a method for accessing documents without necessarily knowing the IDs (keys).
```
bfavini: @RichardSmedley I chatted with Vishal about this--we iterated on the wording a bit and came up with: 

Introduces KV Range Scan, used  to retrieve all documents in a specified range directly from the Data service. Note that in this initial version, you will achieve better performance using a direct fetch or retrieval from a Query with an Index.

bfavini (changes requested):
See note re: conversation with Vishal

modules/introduction/partials/new-features-76.adoc
RichardSmedley: Updated

bfavini (approved):

PR #3543, opened 2024-04-02
Updated Release notes for 7.6.1

bfavini:
Updated release notes for 7.6.1


waynesiu:
The release notes look good.  Will let you know when we should push it live.  Thanks.

RayOffiah (approved):
All good from my point of view.

PR #3578, opened 2024-05-10
DOC-12163 thp-disable steps are not applicable to Rocky linux

RayOffiah:
https://issues.couchbase.com/browse/DOC-12163

### DIFF
--- a/modules/install/pages/thp-disable.adoc
+++ b/modules/install/pages/thp-disable.adoc
@@ -279,18 +279,18 @@ WantedBy=basic.target
 sudo systemctl daemon-reload
 ----
 
-. Ensure that the service will start whenever the system is rebooted.
-+
-[source, console]
-----
-sudo systemctl enable disable-thp
-----
-
 . Start the service.
 +
 [source, console]
 ----
 sudo systemctl start disable-thp
+----
+
+. Ensure that the service will start whenever the system is rebooted.
++
+[source, console]
+----
+sudo systemctl enable disable-thp
 ----
 
 [#verify-thp-service]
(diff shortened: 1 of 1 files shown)


jishnuthankamani (approved):
Looks good to me.

owend74 (commented):
Hi Ray - have we run a test to ensure this work as expected?
Also could get a review from Chris Hillery, who is an expert on out build/install environments?

thanks

RayOffiah:
@jishnuthankamani has run through the changes and tells me they work. (I don't have a Rocky Linux set up at the moment).

I can certainly get Chris to take a look.

modules/install/pages/thp-disable.adoc
```diff
@@ -237,3 +242,66 @@ If THP is properly disabled, the output of both commands should be the following
...
+----
+
+. Reload the `systemctl` files.
++
+[source, console]
+----
+sudo systemctl daemon-reload
+----
+
+. Ensure that the service will start whenever the system is rebooted.
```
ceejatec: Probably best to move this "enable" step to the end, as it was in the DOC ticket.

ceejatec:
Actually, I strongly suspect that we should change this page to _only_ document this method for disabling THP. All Linux distributions we support use systemd these days, and anything referencing /etc/init.d is likely out of date. Give me a little bit to see if I can verify that. In the meantime, if it's urgent that this change go in because of Rocky Linux support, then it's fine as-is (I made one small suggestion in a separate comment).

We _might_ even be able to include this configuration in the couchbase-server package itself, which would render this page obsolete. I'm less sure about that, though; I know we've been reluctant to automate things that muck with customers' system configurations in the past. I'll check on that too.

PR #3590, opened 2024-05-28
Update Transaction related topic link text

sudosalim:
Currently these links look like the same links copied three times. Change the link text to mentioned the SDK for each link

### DIFF
--- a/modules/learn/pages/data/transactions.adoc
+++ b/modules/learn/pages/data/transactions.adoc
@@ -322,10 +322,15 @@ xref:n1ql:n1ql-manage/query-settings.adoc#atrcollection_req[atrcollection] for d
 
 == Related Topics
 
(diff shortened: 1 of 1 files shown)


simon-dew:
Note to @RichardSmedley : I've noticed this issue in other documents. It might be better to fix this issue at the SDK end by giving each page a distinct title, including the name of the SDK language, or by setting the [reftext][1] attribute for each page. You can use the [navtitle][1] attribute to keep the short title of the page in the navigation.

[1]: https://docs.antora.org/antora/latest/page/reftext-and-navtitle/

RichardSmedley (approved):
:+1:

RichardSmedley:
Thanks @sudosalim - have added :reftext: to each title, as per @simon-dew 's suggestion,
and updated with all SDKs that have Distributed ACID txns.

PR #3601, opened 2024-06-07
DOC-12188 new failover recovery metrics

ggray-cb:
This PR covers the impacts of [MB-61751](https://issues.couchbase.com/browse/MB-61751) which adds new metrics for the Cluster Manager to track failovers and rebalances.

Updates, with links to preview:

* Added section to [What's New](https://preview.docs-test.couchbase.com/cm-metrics/server/current/introduction/whats-new.html#cluster-manager) to cover new feature.
* Updated metrics JSON file from recent build to update the metrics shown in [Cluster Manager Metrics](https://preview.docs-test.couchbase.com/cm-metrics/server/current/metrics-reference/ns-server-metrics.html). Hand-edited this file to change all instances of the 7.2.6 version to 7.6.2. This was done because while these metrics will be backported to 7.2.6, 7.6.2 will be released before 7.2.6. Therefore, we should refere to these as originating in 7.6.2 until w ship 7.2.6. A future update will need to do that.


malarky (approved):
Looks good to me :shipit:

PR #3609, opened 2024-06-22
DOC-12225-7.6-Slow-operations-threshold-logging-documentation-should-be-clearer-that-the-command-needs-to-be-run-on-every-node

RayOffiah:
https://issues.couchbase.com/browse/DOC-12225


CB-DavidPancic (approved):
The wording of the original note is slightly awkward

"To change the threshold across the cluster, then all the configurations must be applied to each node."

However regardless I'm happy with the change. Thanks!

RayOffiah:
Okay, I've made the wording a bit clearer.

PR #3612, opened 2024-06-25
Updated the cores per bucket recommendations

ggray-cb:
Changed from 02 to 0.4 and emphasized per-node. Also some language cleanup.


malarky (approved):
Looks good, thank you!

PR #3616, opened 2024-06-27
Backporting DOC-11218

ggray-cb:
(cherry picked from commit 37fc4bf2df2bafe8ca8ce88c8d2807ac182aeea3)

### DIFF
--- a/modules/install/pages/sizing-general.adoc
+++ b/modules/install/pages/sizing-general.adoc
@@ -4,7 +4,7 @@
 [abstract]
 {description}
 
-When you plan to deploy a Couchbase Server cluster, perhaps the most common (and important) question that comes up is: _How many nodes do I need and what size do they need to be?_
+When you plan to deploy a Couchbase Server cluster, perhaps the most common (and important) question that comes up is: how many nodes do I need and what size do they need to be?
 
 With the increasing number of Couchbase services and the flexibility of the Couchbase Data Platform, the answer to this question can be challenging.
 This guide aims to help you better size your deployment.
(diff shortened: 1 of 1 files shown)


modules/install/pages/sizing-general.adoc
```diff
@@ -220,7 +220,9 @@ Note that this amount is in addition to the RAM requirements for the operating s
 
 When sizing, you must account for raw CPU overhead when using a high number of buckets.
 
-* We recommend that you set aside 0.2 core/bucket for operational stability. This overhead does not account for any front-end workload which must be sized separately.
+* Your best practice is to allocate 0.4 cores per bucket on each node to maintain operational stability.
```
osfameron, line 223: Do we really not use "we" any more for best practices? "Your best practice" seems awkward, and a little passive aggressive to me. But I guess needs must, if we're stomping down on "we" 🤷 

Moot, as this is a backport, but just curious.

osfameron (approved):

modules/install/pages/sizing-general.adoc
ggray-cb: Vale's guidance is to avoid first person plural ("we"). Vale is good. Vale must be obeyed. 

I suppose "The best practice..." could also work. But that's also a bit passive and begs the question how this wisdom came about. I tend to avoid using the company's name ("Couchbase's test has determined the best practice to be...") when it collides with the product name because it suggests our product has gained sentience and is telling you what to do. 

(I don't think it's slated to do that until after version 10.0 or so.)

osfameron: I for one welcome our robotic/AI overlords

PR #3626, opened 2024-07-03
[DOC-12334] Added warning about upgrade

julia-cb:
Addressing @gsauere's urgent request to add a warning about upgrading to 7.6.x from 7.2/7.1.


julia-cb:
Merging PR right now and building prod as this is an urgent update.

PR #3648, opened 2024-07-19
[DOC-12362]: Feedback on Swap Space and Kernel Swappiness | RHEL8 Known Issue

RayOffiah:
Added a sidebar and a link explaining the problem and the resolution.

https://issues.couchbase.com/browse/DOC-12362

### DIFF
--- a/modules/install/pages/install-swap-space.adoc
+++ b/modules/install/pages/install-swap-space.adoc
@@ -51,6 +51,7 @@ If the output doesn't match the <<recommended-swappiness-settings,recommended se
 
 . Change the swappiness setting.
 +
+[upperroman]
 .. Set the value for the running system.
 +
 [source,console]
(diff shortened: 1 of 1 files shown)


modules/install/pages/install-swap-space.adoc
```diff
@@ -51,6 +51,7 @@ If the output doesn't match the <<recommended-swappiness-settings,recommended se
 
 . Change the swappiness setting.
 +
+[upperroman]
```
osfameron: This is the bulleted list style?
Is this part of our standard tone? I don't find `upperroman` elsewhere in docs-server

osfameron (approved):
Looks good (one comment about `upperroman` inline)

modules/install/pages/install-swap-space.adoc
RayOffiah: No, that's Roman numbers for an ordered sublist. 

I'd better check the style guide and see what the form is for sublists. 

…
…

https://docs.couchbase.com/styleguide/ordered-list.html
Not too much comment around that area.

I'll drop a note to Sarah.

RayOffiah: Sorted. Had a chat with Sarah, and I've left as the Antora default.

PR #3671, opened 2024-08-15
[DOC-10769-7.2]: Remove the obsolete 'Auto-Failover during Rebalance' paragraph. Removed paragraph related to the auto-failover aborting a rebalance.

RayOffiah:
Removed offending paragraph.

@JackBakes, if you could take a Quick Look at this change and make sure that it is correct, that would be greatly appreciated.

### DIFF
--- a/modules/learn/pages/clusters-and-availability/automatic-failover.adoc
+++ b/modules/learn/pages/clusters-and-availability/automatic-failover.adoc
@@ -9,14 +9,14 @@
 == Understanding Automatic Failover
 
 _Automatic Failover_ — or _auto-failover_ — can be configured to fail over one or more nodes automatically. No immediate administrator intervention is required.
-Specifically, the Cluster Manager autonomously detects and verifies that the nodes are unresponsive, and then initiates the _hard_ failover process.
+Specifically, the Cluster Manager autonomously detects and verifies that the nodes are unresponsive and then initiates the _hard_ failover process.
 Auto-failover does not fix or identify problems that may have occurred.
 Once appropriate fixes have been applied to the cluster by the administrator, a rebalance is required.
 Auto-failover is always _hard_ failover.
 For information on how services are affected by hard failover, see xref:learn:clusters-and-availability/hard-failover.adoc[Hard Failover].
 
 This page describes auto-failover concepts and policy.
-For information on managing auto-failover, see the information provided for Couchbase Web Console at xref:manage:manage-settings/general-settings.adoc[General] (which provides information on general cluster-settings), for the REST API at xref:rest-api:rest-cluster-autofailover-intro.adoc[Managing Auto-Failover], and for the CLI at xref:cli:cbcli/couchbase-cli-setting-autofailover.adoc[setting-autofailover].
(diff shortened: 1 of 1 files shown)


modules/learn/pages/clusters-and-availability/automatic-failover.adoc
```diff
@@ -224,15 +224,9 @@ xref:cli:cbcli/couchbase-cli-setting-autofailover.adoc[setting-autofailover] wit
 [#auto-failover-during-rebalance]
 == Auto-Failover During Rebalance
 
-Couchbase Server provides a setting to determine whether, once enabled, auto-failover should specifically be triggered during xref:learn:clusters-and-availability/rebalance.adoc[Rebalance], in the event of a node becoming unresponsive.
-
-If auto-failover _has_ been set to be triggered, following the configured timeout period, the rebalance is stopped; then, auto-failover is duly triggered.
 Following auto-failover, rebalance is _not_ automatically re-attempted.
```
hyunjuV: Just for additional clarity, I think we should have a sentence before one starting with "Following auto-failover,..."  -- something like below:

If an auto-failover event occurs during a rebalance, the rebalance is stopped; then, auto-failover is triggered.  Following auto-failover, rebalance is _not_ automatically re-attempted.  At this point, the cluster is likely to be in an unbalanced state: therefore, rebalance should be performed manually; and the unresponsive node fixed and restored to the cluster, as appropriate.

modules/learn/pages/clusters-and-availability/automatic-failover.adoc
```diff
@@ -224,15 +224,9 @@ xref:cli:cbcli/couchbase-cli-setting-autofailover.adoc[setting-autofailover] wit
...
 At this point, the cluster is likely to be in an unbalanced state: therefore, rebalance should be performed manually; and the unresponsive node fixed and restored to the cluster, as appropriate.
 
-If auto-failover has _not_ been set to be triggered, unless there is manual intervention, no failover occurs.
-This may cause the rebalance to hang for an indeterminate period; before failing, with error messages.
-
 For information on setting auto-failover in the context of rebalance, see the information on xref:manage:manage-settings/general-settings.adoc[General] settings.
```
hyunjuV: This line -- the current line 230 -- "For information on setting auto-failover in the context of rebalance, ... " should also be removed since users can no longer set/unset auto-failover in the context of rebalance.

hyunjuV (changes requested):

modules/learn/pages/clusters-and-availability/automatic-failover.adoc
RayOffiah: Done 👍🏾

modules/learn/pages/clusters-and-availability/automatic-failover.adoc
RayOffiah: Fixed 👍🏾

ElliotFrancisHunter (approved):
Looks good from the docs POV. Happy to approve 👍🏼

hyunjuV (approved):

PR #3697, opened 2024-09-16
DOC-10648 windows install permissions

ggray-cb:
Updated the requirements for installing and running on Windows to state what user the CB windows sevice runs on, and what permissions you need to start and stop it.

See [preview of this change](https://preview.docs-test.couchbase.com/windows-permissions/server/current/install/install-package-windows.html).


DeaconLinkhorn (approved):
All looks good to me.

PR #3705, opened 2024-09-25
[DOC-12579]: Need to update release note description for MB-63286

RayOffiah:
Changed the discovered bug version to 7.6.0

### DIFF
--- a/modules/release-notes/partials/docs-server-7.2.6-release-note.adoc
+++ b/modules/release-notes/partials/docs-server-7.2.6-release-note.adoc
@@ -34,7 +34,7 @@ For details on new features and enhancements, see the  xref:introduction:whats-n
 | Issue resolved.
 
 | https://jira.issues.couchbase.com/browse/MB-63286[MB-63286]
-| An issue occurred caused by a race condition in the index recovery code path, which may result in item count mismatch and wrong query results. This  issue may occur during an Indexer restart.  However, as part of the file-based rebalance process introduced in `7.2.6`, a recovery of the index is performed after the index is moved, which increases the likelihood that this race condition might be reached.
+| An issue occurred caused by a race condition in the index recovery code path, which may result in item count mismatch and wrong query results. This  issue may occur during an Indexer restart.  However, as part of the file-based rebalance process introduced in `7.6.0`, a recovery of the index is performed after the index is moved, which increases the likelihood that this race condition might be reached.
 | Issue resolved.
 
 |===


modules/release-notes/partials/docs-server-7.2.6-release-note.adoc
```diff
@@ -34,7 +34,7 @@ For details on new features and enhancements, see the  xref:introduction:whats-n
 | Issue resolved.
 
 | https://jira.issues.couchbase.com/browse/MB-63286[MB-63286]
-| An issue occurred caused by a race condition in the index recovery code path, which may result in item count mismatch and wrong query results. This  issue may occur during an Indexer restart.  However, as part of the file-based rebalance process introduced in `7.2.6`, a recovery of the index is performed after the index is moved, which increases the likelihood that this race condition might be reached.
+| An issue occurred caused by a race condition in the index recovery code path, which may result in item count mismatch and wrong query results. This  issue may occur during an Indexer restart.  However, as part of the file-based rebalance process introduced in `7.6.0`, a recovery of the index is performed after the index is moved, which increases the likelihood that this race condition might be reached.
```
RichardSmedley, line 37: Does it make sense to highlight this as a problem for 7.6.x, in a release note aimed at 7.2.x -
perhaps it could be encountered in mixed mode, with a cluster part-way through a swap-rebalance upgrade?
But in that case, it should be noted that it is only then that it would occur.
Or perhaps we should note that the chances of it being a problem are less likely in 7.2.x, before file-based rebalance was introduced.

iow I'm confused about this, and I wonder if it can be less confusing?

(Although feel free to put it through now, as the old version was inaccurate, and confusing is often better than inaccurate.)

RichardSmedley (approved):
Approving to unblock - as the factual error needs correcting - but as per comment, I think this is now confusing, and needs reworking.

modules/release-notes/partials/docs-server-7.2.6-release-note.adoc
RayOffiah: Yes, I can see what you're saying. I'll have a chat with Beth tomorrow about clarifying it, or scrubbing the mention of 7.6.0 altogether. But I will send it through and look again tomorrow.

bfavini: @RichardSmedley @RayOffiah Yes, this is confusing and I think we need to clarify. For no, changing 7.2.6 to 7.6.0 solves the immediate problem, but we should follow up with Ajay to clarify.

PR #3712, opened 2024-10-11
DOC-12635: infrastructure for release/8.0

simon-dew:
Updates docs infrastructure for release/8.0

### DIFF
--- a/antora.yml
+++ b/antora.yml
@@ -1,5 +1,5 @@
 name: server
-version: '7.7'
+version: '8.0'
 asciidoc:
   attributes:
     ui-name: Server Web Console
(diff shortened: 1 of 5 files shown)


osfameron (approved):
Looks good to me (as 8.0 is prerelease does it need active reviewing now, or only at GA time?)

simon-dew:
Thanks @osfameron &mdash; the reason I flagged this particular change for review is because it might impact on work that @ggray-cb and @RayOffiah were doing for release/7.7. I'll probably merge this before the end of the day, so this is just a heads up to Gary and Ray that they may need to rebase any in-flight dev branches.

PR #3722, opened 2024-11-08
[DOC-12706]: Documentation needed for changes to the Prometheus Discovery API

RayOffiah:
Added information on the new `clusterLabels` parameter to the Prometheus discovery API page.


RayOffiah:
Too many line-breaks changes; trying again with different editor settings!

PR #3729, opened 2024-11-15
DOC-12559 Error in REST Configure Rebalance Retry documentation 7-0

rao-shwe:
DOC-12559


RayOffiah (approved):
Yup, works fine. Good to go.👍🏾

PR #3732, opened 2024-11-20
[DOC-12729/release/7.6]: Feedback on Disabling Transparent Huge Pages (THP) 

RayOffiah:
Added extra details concerning the RSS and OOM kill.
Tidied up some of the text.

### DIFF
--- a/modules/install/pages/thp-disable.adoc
+++ b/modules/install/pages/thp-disable.adoc
@@ -3,18 +3,26 @@
 :tabs:
 
 [abstract]
+--
 {description}
-THP must be disabled in order for Couchbase Server to function correctly on Linux.
+
+THP must be disabled in order for Couchbase Server to function correctly on Linux, as having THP enabled can worsen performance and possibly lead to an OOM kill. 
+--
 
 In Linux operating systems, _huge pages_ is a feature that provides a way for the CPU and OS to create pre-allocated contiguous memory space, and which is designed to improve application performance.
-_Transparent huge pages (THP)_ is a Linux OS feature that automates the creation of contiguous memory space, and conceals much of the complexity of using actual huge pages on systems with large amounts of memory.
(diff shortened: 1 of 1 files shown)


modules/install/pages/thp-disable.adoc
```diff
@@ -3,18 +3,24 @@
 :tabs:
 
 [abstract]
+--
 {description}
```
AlphaDaze, line 7: I believe the current description can remain the same, especially keeping the exsiting point of "THP must be disabled in order for Couchbase Server to function correctly on Linux.".

We could add to the abstract that having THP enabled can worsen performance and possibly lead to an OOM kill. However, these points I believe should remain concise within the abstract.

modules/install/pages/thp-disable.adoc
```diff
@@ -3,18 +3,24 @@
...
+As a result, the data stored which we keep track of may be smaller while RSS can be significantly higher,
+leading to possible OOM kill.
+--
 
 In Linux operating systems, _huge pages_ is a feature that provides a way for the CPU and OS to create pre-allocated contiguous memory space, and which is designed to improve application performance.
-_Transparent huge pages (THP)_ is a Linux OS feature that automates the creation of contiguous memory space, and conceals much of the complexity of using actual huge pages on systems with large amounts of memory.
+_Transparent huge pages (THP)_ is a Linux OS feature that automates the creation of contiguous memory space and conceals much of the complexity of using actual huge pages on systems with large amounts of memory.
 
 THP is enabled by default in most Linux operating systems, and functions very well for most applications and processes.
 However, THP is detrimental to Couchbase's performance (as it is for nearly all databases that tend to have sparse rather than contiguous memory access patterns).
```
AlphaDaze, line 16: We can add your new points from above here instead to explain the reasoning in detail.

AlphaDaze (approved):
LGTM

PR #3750, opened 2024-12-09
DOC-12777: Update Eventing REST API with sync-gateway-aware options

simon-dew:
Docs issue: DOC-12777

Adds Sync Gateway-aware Eventing REST API options to What's New in Couchbase Server 7.6.4. I haven't linked to the originating ticket because it was a CBSE.

Also updates links to the Eventing REST API. This is not strictly necessary because they were caught and handled by page redirects, but it's nice to point to the correct Antora co-ordinates.


ElliotFrancisHunter (approved):
LGTM 👍🏼

PR #3759, opened 2024-12-19
Fix references to non-existent 7.2.3 CE edition

ggray-cb:
There was no 7.2.3 version of CE, so we need to say 7.2.2.


malarky (approved):
:shipit:

PR #3761, opened 2024-12-26
Update do-a-quick-install.adoc

mmyounes:
Removed the part referring to the sandbox container since it is not used after that.
Also removed the image version "7.2.0" to allow users to download the latest image and explore new features and functionalities

### DIFF
--- a/modules/getting-started/pages/do-a-quick-install.adoc
+++ b/modules/getting-started/pages/do-a-quick-install.adoc
@@ -7,7 +7,10 @@
 {description}
 Once you install Docker, you can use a single command to download and install a pre-configured version of Couchbase Server Enterprise Edition on your computer.
 
-
+NOTE: The Docker container image that is used in this topic is meant to set up a 'sandbox' version of Couchbase Server.
+It's based on the Enterprise Edition and contains scripts that automatically configure several Couchbase components during installation.
+Although this particular image works great for a demo, it is not meant for production.
+For information about installing Couchbase Server in production, see xref:install:get-started.adoc[].
 
 [#initialize-cluster-web-console]
 == Install Couchbase Server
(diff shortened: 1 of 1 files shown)


modules/getting-started/pages/do-a-quick-install.adoc
```diff
@@ -7,10 +7,7 @@
 {description}
 Once you install Docker, you can use a single command to download and install a pre-configured version of Couchbase Server Enterprise Edition on your computer.
 
-NOTE: The Docker container image that is used in this topic is meant to set up a 'sandbox' version of Couchbase Server.
-It's based on the Enterprise Edition and contains scripts that automatically configure several Couchbase components during installation.
-Although this particular image works great for a demo, it is not meant for production.
-For information about using containers to run Couchbase Server in production, see the xref:cloud:couchbase-cloud-deployment.adoc[Cloud and Container Deployment Overview].
+
```
simon-dew: I think this is actually a useful caveat. We need to emphasize that a Docker install is not suitable for production. But the link is hyper-specific and not useful. Reinstated the note but updated the link to point to the Install section.

```suggestion
NOTE: The Docker container image that is used in this topic is meant to set up a 'sandbox' version of Couchbase Server.
It's based on the Enterprise Edition and contains scripts that automatically configure several Couchbase components during installation.
Although this particular image works great for a demo, it is not meant for production.
For information about installing Couchbase Server in production, see xref:install:get-started.adoc[].
```

modules/getting-started/pages/do-a-quick-install.adoc
```diff
@@ -25,7 +22,7 @@ Open a console window on your computer and enter the following command:
 
 [source,console]
 ----
-docker run -t --name db -p 8091-8096:8091-8096 -p 11210-11211:11210-11211 couchbase/server:enterprise-7.2.0
+docker run -t --name db -p 8091-8096:8091-8096 -p 11210-11211:11210-11211 couchbase/server:enterprise
```
simon-dew: Checked and it works. Suggest adding a line break because it looks bad having the command overflow the text box.

```suggestion
docker run -t --name db -p 8091-8096:8091-8096 -p 11210-11211:11210-11211 \
  couchbase/server:enterprise
```

simon-dew (changes requested):
Thanks for the contribution!

simon-dew (approved):

PR #3762, opened 2025-01-02
DOC-12886: correcting bucket memory quotas info

hyunjuV:
Doc-12886 -- correcting bucket memory quotas info -- the corrected info is consistent with the information here --https://docs.couchbase.com/server/current/learn/buckets-memory-and-storage/storage-engines.html#couchstore-and-magma-at-a-glance


shivaniguptasf (approved):
Looks good

ggray-cb (approved):

PR #3768, opened 2025-01-15
DOC-12699: Update views-intro.adoc

hyunjuV:
Update views deprecation notice to include that customers should use "Instead of views, use indexes and queries using the Index Service (GSI) and the Query Service (SQL++)".


shivaniguptasf (approved):
Looks good.

ggray-cb (approved):

PR #3780, opened 2025-02-12
DOC-12887: Update cbstats-reset.adoc

hyunjuV:
For cbstats reset, nothing that the command is for dev/test use only.


trondn (approved):
LGTM

PR #3789, opened 2025-02-26
[DOC-12953/release/7.0]: Docs incorrectly state that cbbackupmgr is EE only

RayOffiah:
Updated the documentation to correctly state that `cbbackupmgr` is available in both Enterprise and Community Editions. Improved clarity regarding version compatibility and packaging information.


rao-shwe (approved):
LGTM!

PR #3804, opened 2025-05-09
[DOC-12746]: We need to add info to the current documentation to let customers know how audit log records events

RayOffiah:
- Added notes to the documentation to inform users how audit logs record events.


ggray-cb (approved):
LGTM. Thanks Ray!

PR #3808, opened 2025-05-16
DOC-13223: Added guardrails details to server docs

TimLFletcher:
Capella docs have the guardrails discussion that also applies to server... so I'm bringing it in here as a result of a CBSE.

Linked issue in docs-devex DOC-13223-new

### DIFF
--- a/modules/install/pages/sizing-general.adoc
+++ b/modules/install/pages/sizing-general.adoc
@@ -137,7 +137,7 @@ These thresholds are as follows:
 |Free Disk Space
 |\<= 4%
 
-|# of Buckets
+|Number of Buckets
 |\<= 0.2 cores per bucket
 
 |===
--- a/modules/install/pages/sizing-general.adoc
+++ b/modules/install/pages/sizing-general.adoc
@@ -108,6 +108,40 @@ A reliable high-speed network for intra-cluster and inter-cluster communications
 +
 Most deployments can achieve optimal performance with 1 Gbps interconnects, but some may need 10 Gbps.
 
+
+[#guardrails]
+==== Couchbase Server Guardrails
+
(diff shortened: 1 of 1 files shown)


modules/install/pages/sizing-general.adoc
```diff
@@ -108,6 +108,40 @@ A reliable high-speed network for intra-cluster and inter-cluster communications
...
+|Data Service Resident Ratio (Couchstore)
+|\<= 1%
+
+|Data Service Resident Ratio (Magma)
+|\<= 0.1%
+
+|Free Disk Space
+|\<= 4%
+
+|# of Buckets
```
RichardSmedley: Is it better to spell out as "Number of Buckets",
as use of the octothorpe is not entirely universal beyond American English?
(_Close_ to universal, but not as unambiguous as writing 'number')

modules/install/pages/sizing-general.adoc
```diff
@@ -108,6 +108,40 @@ A reliable high-speed network for intra-cluster and inter-cluster communications
...
+
+Server-level enforcement thresholds are designed to ensure the optimal and reliable usage of available system resources by various workloads.
+These thresholds are as follows:
+
+[Attributes]
+|===
+|Cluster Metric |Threshold
+
+|Data Service Resident Ratio (Couchstore)
+|\<= 1%
```
osfameron, line 132: I think this is fine, but I did wonder if it needs the backslash escape?
and whether `<=` is clear enough without explanation (but should be fine, is used elsewhere in docs-server in a text context, once, and I don't see any counter-indications in the style guide.

modules/install/pages/sizing-general.adoc
TimLFletcher: I think that's a fair suggestion. Changed.

PR #3809, opened 2025-05-16
DOC-12393 native encryption at rest

ggray-cb:
This PR adds documentation for the native encryption at rest feature ([MB-16143](https://jira.issues.couchbase.com/browse/MB-16143)). 

[Here's the doc plan for this feature.](https://docs.google.com/document/d/1BlsE6EjWUMaHLWJCFP5_HFbZsQ2UaqU2UXNybVg-Vhg/edit?tab=t.0)

Pages added/changed for this feature (links lead to perview site):

- [What's New](https://preview.docs-test.couchbase.com/docs-server-DOC-12393_native_encryption_at_rest/server/current/introduction/whats-new.html#security) entry.
- [Encryption](https://preview.docs-test.couchbase.com/docs-server-DOC-12393_native_encryption_at_rest/server/current/learn/security/encryption-overview.html) edited for standards, also added in mention of native encryption at rest.
- [Native Encryption at Rest Overview](https://preview.docs-test.couchbase.com/docs-server-DOC-12393_native_encryption_at_rest/server/current/learn/security/native-encryption-at-rest-overview.html)
- [Manage Native Encryption at Rest](https://preview.docs-test.couchbase.com/docs-server-DOC-12393_native_encryption_at_rest/server/current/manage/manage-security/manage-native-encryption-at-rest.html)
- [Encryption-at-Rest API](https://preview.docs-test.couchbase.com/docs-server-DOC-12393_native_encryption_at_rest/server/current/rest-api/security/encryption-at-rest/encryption-at-rest.html)
- [Manage Encryption-at-Rest Keys](https://preview.docs-test.couchbase.com/docs-server-DOC-12393_native_encryption_at_rest/server/current/rest-api/security/encryption-at-rest/manage-encryption-keys.html)
- [Manage Audit, Config, and Log Encryption at Rest](https://preview.docs-test.couchbase.com/docs-server-DOC-12393_native_encryption_at_rest/server/current/rest-api/security/encryption-at-rest/manage-system-encryption-at-rest.html)
- [Rotate Data Encryption Keys](https://preview.docs-test.couchbase.com/docs-server-DOC-12393_native_encryption_at_rest/server/current/rest-api/security/encryption-at-rest/rotate-encryption-at-rest-key.html)
- [Creating and Editing Buckets](https://preview.docs-test.couchbase.com/docs-server-DOC-12393_native_encryption_at_rest/server/current/rest-api/rest-bucket-create.html) - added the new encryption at rest settings.
-  [Getting Bucket Information](https://preview.docs-test.couchbase.com/docs-server-DOC-12393_native_encryption_at_rest/server/current/rest-api/rest-buckets-summary.html) updated example output to include updated fields. Also changed example a bit to remove hand-editing of JSON for potential future automated doc testing.

[MB-16143]: https://couchbasecloud.atlassian.net/browse/MB-16143?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[See here for the passwords for the preview site.](https://confluence.issues.couchbase.com/wiki/spaces/DOCS/pages/1971224949/Docs+Team+demo+site+passwords)


modules/learn/pages/security/native-encryption-at-rest-overview.adoc
```diff
@@ -0,0 +1,175 @@
...
+Even if one or more nodes in your cluster suffer a security breach, the encryption keys remain secure within the AWS KMS.
+One downside of using the AWS KMS is that the cluster relies on an external source for its encryption keys.
+Disruptions in AWS or the network could result in errors because the cluster cannot retrieve encryption keys to encrypt or decrypt data.
+
++
+[#aws-kms-caution]
+[CAUTION]
+====
+Do not use encryption keys managed by AWS KMS to directly encrypt data.
+While retrieving individual encryption keys from AWS can take less than a second, cluster startup could result in many synchronous key retrievals.
```
timofey-barmin: > While retrieving individual encryption keys from AWS can take less than a second

We don't retrieve keys from AWS KMS. We use AWS KMS for encryption and decryption on their side.
Something like "While each encryption and decryption with AWS key can take up to a second..." would be more correct.

modules/learn/pages/security/native-encryption-at-rest-overview.adoc
```diff
@@ -0,0 +1,175 @@
...
+These key retrievals can cause delays during cluster start.
+
+Only use the encryption keys you store in the AWS KMS as Key Encryption Keys (KEKs).
```
timofey-barmin: > Only use the encryption keys you store in the AWS KMS as Key Encryption Keys (KEKs)

This phrase doesn't sound fully correct to me because any encryption key described on this page is always used as a KEK.
For example, when you are configuring AWS Key to encrypt configuration data, it really means that "AWS Key" encrypts "Configuration DEKs", while "Configuration DEKs" encrypt "Configuration data". So as you see the AWS key plays a KEK role in this case as well.
What I think we should say here is something like:

> We recommend using AWS Key to encrypt Couchbase Server Managed Keys instead of assigning them to encrypt data directly.

In other words instead of: AWS Key -> DEKs -> Data
we recommend using: AWS Key -> CB Managed Key -> DEKs -> Data

modules/learn/pages/security/native-encryption-at-rest-overview.adoc
```diff
@@ -0,0 +1,175 @@
...
+Use these keys to encrypt Couchbase Server managed encryption keys. 
+This method limits the number of retrievals from AWS while maintaining the security advantage of having keys managed by a remote KMS.
```
timofey-barmin: The same as above: We don't retrieve keys from AWS KMS. Encryption and decryption happens on AWS KMS side. Basically we ask AWS to encrypt or decrypt stuff for us with provided key ARN.
> This method limits the number of encryption operations made by AWS while maintaining the security advantage of having keys managed by a remote KMS.

modules/learn/pages/security/native-encryption-at-rest-overview.adoc
```diff
@@ -0,0 +1,175 @@
...
+AWS Key Management Service::
+You can use the https://docs.aws.amazon.com/kms/latest/developerguide/overview.html[AWS Key Management Service] (AWS KMS) to manage keys for you.
+Using this method requires some configuration.
+However, this method is more secure because AWS keeps the encryption keys secure internally.
+Even if one or more nodes in your cluster suffer a security breach, the encryption keys remain secure within the AWS KMS.
+One downside of using the AWS KMS is that the cluster relies on an external source for its encryption keys.
+Disruptions in AWS or the network could result in errors because the cluster cannot retrieve encryption keys to encrypt or decrypt data.
+
++
+[#aws-kms-caution]
```
timofey-barmin: We probably should have the same caution for KMIP as well.
If you decide to use one caution for both of them there is one caveat though:
In comments below I am saying that we are not retrieving keys from AWS.
Well, for KMIP we can retrieve keys from the KMIP server (it is optional) so we need to be careful with phrasing if you decide to write a single caution. Maybe we can use more generic word like "operation".

modules/learn/pages/security/native-encryption-at-rest-overview.adoc
```diff
@@ -0,0 +1,175 @@
...
+[#kms]
+== Encryption Key Management Services
+
+A key management service generates and stores the encryption keys Couchbase Server uses to encrypt its DEKs.
+When you create a key, you choose which key management service you want to manage it.
+Couchbase Server works with several key management services:
+
+Couchbase Server Secret Management::
+You can have Couchbase Server act as the key management system.
+This method does not require any additional configuration.
```
timofey-barmin: > This method does not require any additional configuration.
> However, using this method means that Couchbase Server stores your encryption keys locally, leaving them more vulnerable if there is a security breach.

This is not fully correct. It can be configured to be secure. 
Maybe something like:

> Couchbase Server manages the encryption key material (it generates, rotates, removes it with respect to the encryption key configuration).
This type of keys can be encrypted either with another encryption key (e.g. KMIP key), or with node master password (this method requires master password to be properly configured on each Couchbase node, note that by default master password is not configured).

modules/learn/pages/security/native-encryption-at-rest-overview.adoc
```diff
@@ -0,0 +1,175 @@
...
+
+* Bootstrap information
+* Node and internal client certificates
+* Prometheus configuration, metric data, and tokens used to gather metrics
+====
+
+[#kms]
+== Encryption Key Management Services
+
+A key management service generates and stores the encryption keys Couchbase Server uses to encrypt its DEKs.
```
timofey-barmin: > A key management service generates and stores the encryption keys Couchbase Server uses to encrypt its DEKs.

We don't always generate and store keys that encrypt DEKs.
We only do it in case of "Couchbase Server Managed Key".
In case of AWS and KMIP, keys are generated and stored externally. Couchbase basically only uses them for encryption and decryption but we don't store them on disk (in some cases we can keep them in memory though).

modules/learn/pages/security/native-encryption-at-rest-overview.adoc
```diff
@@ -0,0 +1,175 @@
...
+Also, depending on the KMS implementation and its location, decrypting data could introduce performance issues due to latency.
+
+
+
+[#kms-and-keys]
+== Using Multiple KMSs and Multiple Keys
+
+In a basic setup, you may choose to use a single encryption key managed by Couchbase Server to encrypt all data.
+This configuration is easy to manage, but does not offer much flexibility.
+Also, because Couchbase Server stores the encryption key locally, it does not provide a high level of security.
```
timofey-barmin: > Also, because Couchbase Server stores the encryption key locally, it does not provide a high level of security.

I find this phase confusing. I suggest we remove it.

modules/learn/pages/security/native-encryption-at-rest-overview.adoc
```diff
@@ -0,0 +1,175 @@
...
+It helps limit the exposure of data if a data breach compromises an encryption key.
+
+You can choose to have Couchbase Server rotate DEKs automatically.
+You can also have it automatically rotate encryption keys that it manages.
+Rotation of an externally-managed encryption key is handled by the KMS that manage it.
+By default, Couchbase Server automatically rotates DEKs but not the encryption keys it manages. 
+You choose how frequently Couchbase Server rotates DEKs and (if you enable it) its encryption keys.
+
+When Couchbase Server generates a new encryption key or DEK during rotation, it does not immediately delete the expired key.
+It keeps the expired keys so it can decrypt the data encrypted with the older key.
```
timofey-barmin: > When Couchbase Server generates a new encryption key or DEK during rotation, it does not immediately delete the expired key.
It keeps the expired keys so it can decrypt the data encrypted with the older key.

I sounds confusing to me when we say expired about historical keys.

I think a key expires when its lifetime elapses. Until then, it is either active (when we use for encryption and decryption), or inactive = historical (when we use it for decryption only).


```
                   DEK lifetime
        /--------------------------------------------\

    ----|-----------------|-----------------|--------x--------|----------x------
       DEK1             DEK2              DEK3      DEK1     DEK4       DEK2
    generated         generated         generated   expires  generated  expires

        \----------------/
           DEK rotation
             interval
```

modules/manage/pages/manage-security/manage-native-encryption-at-rest.adoc
```diff
@@ -0,0 +1,382 @@
...
+
+For an overview of native encryption at rest, see xref:learn:security/native-encryption-at-rest-overview.adoc[].
+
+## Enabling Encryption
+
+Enabling encryption at rest is a two step process:
+
+. Create at least one encryption key.
+You must use an encryption-at-rest key to encrypt bucket data.
+You can choose to use the master database password instead of an encryption-at-rest key to encrypt audit, configuration, and log data.
```
timofey-barmin: We probably want to put a link to the master password description here.

E.g. [this doc](https://docs.couchbase.com/server/current/manage/manage-security/manage-system-secrets.html#setting-the-master-password)

modules/manage/pages/manage-security/manage-native-encryption-at-rest.adoc
```diff
@@ -0,0 +1,382 @@
...
+For each key you create, you must choose a Key Management Service (KMS) that maintains the key for you.
+You have three options to choose from:
+
+* Amazon's AWS KMS 
+* Any KMS that implements the Key Management Interoperability Protocol (KMIP) 
+* Couchbase Server
+
+See xref:learn:security/native-encryption-at-rest-overview.adoc#kms[Encryption Key Management Services] for more information.
+
+You must decide what your encryption can encrypt.
```
timofey-barmin: Did you mean "You must decide what your encryption key can encrypt"?

modules/manage/pages/manage-security/manage-native-encryption-at-rest.adoc
```diff
@@ -0,0 +1,382 @@
...
+
+See xref:learn:security/native-encryption-at-rest-overview.adoc#kms[Encryption Key Management Services] for more information.
+
+You must decide what your encryption can encrypt. 
+You can create an encryption key that can encrypt any data and other encryption keys.
+You can also choose to limit what data the key can be use to encrypt, or configure it to only encrypt other encryption keys. 
+See learn:security/native-encryption-at-rest-overview.adoc#keys[Encryption-at-Rest Keys] for more information.
+
+You must have the proper privileges to create encryption keys.
+Only users with the xref:learn:security/roles.adoc#full-admin[Full Admin] or xref:learn:security/roles.adoc#security-admin[Security Admin] roles can create them.
```
timofey-barmin: (this comment describes roles that have write access for encryption keys)

It is a little bit more complex that this.
Required roles depend on the key usages.

If key is meant to encrypt logs, config, or audit (usages log-encryption, config-encryption, audit-encryption), the full admin or security admin roles are required.
For bucket encryption it is a bit more relaxed.
Full admin and security admins will still work, but also the following roles will work:

Keys that have "any bucket encryption" usage can be created or edited also by Cluster Admin, Eventing Full Admin, Backup Full Admin.

Keys that have "specific bucket encryption" usage can be created or edited also by Bucket Admin (e.g. bucket admin for bucket "travel-sample" can create a key that encrypts bucket "travel-sample" only).

modules/manage/pages/manage-security/manage-native-encryption-at-rest.adoc
```diff
@@ -0,0 +1,382 @@
...
+
++
+image::manage-security/add-encryption-key-uses.png[,300]
+
++
+If you want your key to only be able encrypt specific buckets, deselect *Data* and then select the buckets.
+
++
+IMPORTANT: If you're creating an encryption key managed by the AWS KMS, only leave *Key Encryption Key (KEK)* selected.
+Do not use AWS KMS to directly encrypt any type of data.
```
timofey-barmin: For KMIP keys we should recommend the same.

modules/manage/pages/manage-security/manage-native-encryption-at-rest.adoc
```diff
@@ -0,0 +1,382 @@
...
+[start=8]
+. Enter the host and port number for the KMS server and choose a timeout for network connections.
+. Choose which certificates to use when verifying the identity of the KMS.
+You can choose to not verify the KMS's identity, however this is insecure. 
+. Enter the details for the client certificate Couchbase Server uses to authenticate with the KMS.
+This information includes how to encrypt the certificate passphrase.
+. In the *KMIP Encryption/Decryption Approach* field, choose how Couchbase Server interacts with the KMS:
+
++
+* Select *Use KMIP Get & encrypt locally*, if you want Couchbase Server to retrieve the encryption key from the KMS and use it locally to decrypt keys and data.
```
timofey-barmin, line 131: ... and use it locally to encrypt or decrypt keys and data.

modules/manage/pages/manage-security/manage-native-encryption-at-rest.adoc
```diff
@@ -0,0 +1,382 @@
...
+* Select *Use KMIP native Encrypt/Decrypt operation* to have Couchbase Server send the encrypted DEKs to the KMS so it can decrypt them. 
+This method is more secure, because the encryption key does not leave the KMS.
+However, this does result in more round trips to the KMS.
+Depending on the KMS configuration, network latency, and other factors, these addtional KMS requests may affect performance.
```
timofey-barmin: > However, this does result in more round trips to the KMS.
Depending on the KMS configuration, network latency, and other factors, these addtional KMS requests may affect performance.

While it would logical for the system to work like this, it is not what we do right now so I suggest we remove it for now. We will need to add it back when we implement it.

modules/manage/pages/manage-security/manage-native-encryption-at-rest.adoc
```diff
@@ -0,0 +1,382 @@
...
++
+--
+This option has Couchbase Server manage the key.
+To complete creating the key:
+
+[start=8]
+. Choose whether you want to use the cluster's master password or another encryption key to encrypt your new key.
+If you want to use another encryption key, it must be configured as a Key Encryption Key (KEK).
+. Decide whether you want Couchbase Server to cache the key.
+This setting lets Couchbase Server keep the key unencrypted in memory so it does not have to decrypt it for each read or write.
```
timofey-barmin: maybe "...  so it does not have to read and decrypt it for each encryption or decryption"?

modules/manage/pages/manage-security/manage-native-encryption-at-rest.adoc
```diff
@@ -0,0 +1,382 @@
...
+. Decide whether you want to have the encryption-at-rest key auto rotate. 
+If you choose to rotate it, enter how often to rotate, a date and time for the first rotation.
+See xref:learn:secrutiy/native-encryption-at-rest-overview.adoc#rotation-expiration[Encryption Key Rotation and Expiration] for more information about key rotation.
+--
+
+====
+
+### Create an Encryption Key Using the REST API
+
+The REST API's `/settings/encryptionKey` endpoint lets you create and manage encryption keys.
```
timofey-barmin: `/settings/encryptionKeys`

modules/manage/pages/manage-security/manage-native-encryption-at-rest.adoc
```diff
@@ -0,0 +1,382 @@
...
+The output of running the  previous example looks like this:
+
+include::rest-api:example$encryption-at-rest/create-auto-generated-key-response.adoc[]
+
+See xref:rest-api:security/encryption-at-rest/manage-encryption-keys.adoc#create-key[Create or Update an Encryption-at-Rest Key] for details on how to create keys using the REST API.
+
+## Encrypt Data at Rest
+
+Once you have created an encrytion-at-rest key, you can use it to encrypt bucket data.
+For audit, configuration, and log data, you can choose to use the master database password instead of an encryption-at-rest key.
```
timofey-barmin: Here and maybe in other places where we say that we can use master password for encryption-at-rest it is probably worth adding a caution like the following:

```
Master password must be configured on each node prior to using it for encryption at rest. By default master password is empty.
```

modules/manage/pages/manage-security/manage-native-encryption-at-rest.adoc
```diff
@@ -0,0 +1,382 @@
...
+
+The following sections explain how to enable encryption for each type of data.
+
+### Encrypt Bucket Data 
+
+To encrypt a bucket, you must have at least one encrytion-at-rest key configured to encrypt all buckets or the specific bucket you want to encrypt.
```
timofey-barmin, line 197: We probably should explain what happens when encryption is enabled and disabled.
Specifically we should clarify the following:
After encryption gets enabled we don't immediately encrypt all the bucket data. We start encrypting new data that is being written. The data that already exists in the bucket is not immediately encrypted.
Eventually the data will be encrypted, as data is written and compacted.
If we want to force encryption for all existing data (make sure everything is encrypted), we should either press "re-encrypt button" in UI, or use `POST /controller/dropEncryptionAtRestDeks/bucket/<bucket>`

The same for disabling encryption.
When encryption for bucket gets disabled, we don't immediately decrypt all the data. We start writing new data unencrypted, but old data remains encrypted.
Bucket data can be fully decrypted using the same approach as for encryption: "re-encrypt button" (needs to be renamed) or by means of endpoint `POST /controller/dropEncryptionAtRestDeks/bucket/<bucket>`

Main point here is that "encryption enabled" is not the same as "bucket data is encrypted".
As well "encryption disabled" is not the same as "bucket data is decrypted"

modules/manage/pages/manage-security/manage-native-encryption-at-rest.adoc
```diff
@@ -0,0 +1,382 @@
...
+include::rest-api:example$encryption-at-rest/bucket-encryption-examples.adoc[tag=alter-bucket]
+
+If the bucket is already encrypted, Couchbase Server re-encrypts the bucket using the new key. 
+If the bucket is not encrypted, Couchbase Server encrypts it.
+
+See xref:rest-api:rest-bucket-create.adoc[] for more inform,ation about creating and updating buckets using the REST API.
+
+### Encrypt Audit, Configuration, and Log Data
+
+You can encrypt audit, configuration, and log data using the master database password or an encryption-at-rest key.
```
timofey-barmin: Is is probably worth it to explain what happens when encryption is getting enabled or disabled.
For logs, audit, and configuration the behavior is very similar to bucket data.
Enablement and disablement means that only new data will be encrypted/decrypted. Historical data is not encrypted/decrypted automatically, but it can be forced by using re-encrypt button in UI or by using API: `POST /controller/dropEncryptionAtRestDeks/<data-type>`
where data-type is log, audit, or config.

modules/manage/pages/manage-security/manage-native-encryption-at-rest.adoc
```diff
@@ -0,0 +1,382 @@
...
+. Click the menu:Encryption at Rest[] tab.
+. Click the btn:[Rotate] button next to the key you want to rotate.
+
+To manually rotate DEKs for a bucket using the Couchbase Server Web Console:
+
+. Click menu:Buckets[] on the main menu.
+. Click the name of the bucket whose DEKs you want to rotate.
+. Click the btn:[Re-encrypt] button.
+. In the *Confirm Rotate DEKs & Re-encrypt Data* dialog, click btn:[Rotate DEKs & Re-encrypt] to confirm you want to rotate the DEKs and re-encrypt the data.
+
```
timofey-barmin, line 506: Probably worth adding description for manual DEKs rotation for configuration, logs, and audit encryption which can be done from the encryption-at-rest UI page (click "rotate & re-encrypt now")

modules/manage/pages/manage-security/manage-native-encryption-at-rest.adoc
```diff
@@ -0,0 +1,382 @@
...
+    "oldestDekCreationDatetime": "2025-05-14T15:57:20Z"
+  },
+  "encryptionAtRestKeyId": 18
+}
+----
+
+To view the encryption status of audit, configuration, and log data using the REST API, send a GET request to the `/settings/security/encryptionAtRest` endpoint.
+See xref:rest-api:security/encryption-at-rest/manage-system-encryption-at-rest.adoc#get-settings[Get Audit, Config, and Log Encryption-at-Rest Settings] for more information.
+
+## Manually Rotating Keys
```
timofey-barmin: This paragraph seems a bit confusing.
It mixes two processes that are pretty different. Probably we should describe them separately.
1. Rotation of encryption keys (like Couchbase Server Managed key): this process is light weight, we basically need to re-encrypt corresponding DEKs with a new key. I think we should describe rotation for each type of the key actually as it works a bit different for each of them. See my comment below.
2. "Manual rotation of DEKs" which is not only a rotation actually but also drop of all of the DEKs and full re-encryption of data (we generate new DEK, then re-encrypt all data with that new DEK, so in the end there is no old DEKs in use). This process can be very heavy for bucket data. For audits, logs and configuration it is less heavy. It depends on the how much data we have on disk. 

Regarding manual rotation for DEKs: we don't have manual rotation currently (without re-encryption), it can only be done automatically with a configured interval

modules/manage/pages/manage-security/manage-native-encryption-at-rest.adoc
```diff
@@ -0,0 +1,382 @@
...
+
+You can manually rotate encryption-at-rest keys that Couchbase Server manages.
```
timofey-barmin, line 363: We should describe rotation for each key type actually:
1. Couchbase Server managed keys:  when rotation happens couchbase server generates a new key material, and re-encrypts all the DEKs that are configured to be encrypted with this encryption key. The process of re-encryption is asynchronous, as not all DEKs can be available at the moment when re-encryption happens (e.g. some nodes can be disconnected). Historical keys material is not removed automatically as there can hypothetically exist backups that were encrypted by those key materials. They can be remove manually later.
2. AWS keys: The AWS key material is supposed to be rotated manually or automatically externally (at the AWS KMS). After the key is rotated at the AWS KMS, administrator needs to press rotate button in CB UI. This will trigger re-encryption of all DEKs that are configured to be encrypted with this encryption key. The important part here is that Couchbase Server doesn't rotate the AWS key it self. It should be triggered externally.
3. KMIP keys: In order to rotate KMIP keys administrator needs to create new KMIP key at the KMIP server and get new keys ID (Couchbase server has nothing to do with that). Then they should click edit on the KMIP key they want to rotate and enter the new KMIP key ID in the corresponding field. Then click next. The system should start re-encryption of corresponding DEKs right after that. Note that old (historical) KMIP key ID can still be used for some time until the server re-encrypts all of the DEKs, so those historical KMIP key should not get removed at the KMIP server immediately after rotation. When re-encryption is finished the historical key ID should be removed at the UI first, and only then that key can be removed at the KMIP server.

modules/rest-api/pages/rest-bucket-create.adoc
```diff
@@ -66,93 +71,141 @@ curl -X POST -u <administrator>:<password>
...
+  -d allowedTimePeriod[abortOutside]=[ true | false ]
+
 ----
 
 All parameters are described in the following subsections.
 
 == Parameter Groups
 
-Parameters that support the creation and editing of buckets can be considered to form two groups; which are, respectively, _General_ and _Auto-compaction_.
+Parameters that support the creation and editing of buckets can be broken two groups: Genera and Auto-compaction.
```
timofey-barmin: > can be broken two groups: Genera and Auto-compaction

typos

modules/rest-api/pages/rest-bucket-create.adoc
```diff
@@ -830,39 +915,104 @@ curl -v -X POST http://localhost:8091/pools/default/buckets/testBucket \
...
+The following example sets the DEK rotation interval to 15 days (1,296,000 seconds):
+
+include::example$encryption-at-rest/bucket-encryption-examples.adoc[tag=set-dek-rotation]
+
+
+[[encryptionAtRestDekLifetime]]
+=== encryptionAtRestDekLifetime
+
+Sets the lifetime of the bucket's data encryption keys (DEKs) in seconds.
+Once this period passes after a DEK expires, Couchbase Server re-encrypts any data that's still encrypted the data using the active DEK.
```
timofey-barmin: > Once this period passes after a DEK expires

DEK's lifetime starts after DEK creation.

modules/rest-api/pages/rest-bucket-create.adoc
```diff
@@ -830,39 +915,104 @@ curl -v -X POST http://localhost:8091/pools/default/buckets/testBucket \
...
+It then deletes the expired DEK.
+
+This value defaults to `31536000`, which means Couchbase Server keeps expired DEKs for 365 days.
+Setting this value to 0 means Couchbase Server never deletes expired DEKs.
+
+If you set `encryptionAtRestDekRotationInterval` to a non-zero value and `encryptionAtRestDekLifetime` to 0, Couchbase Server keeps old DEKs forever.
+Depending on how often you rotate the DEKs, this can lead to a large number of DEKs being kept.
```
timofey-barmin: There is one more thing here.
The number of DEKs is limited by 50 currently. If we reach that number, new DEKs will not be generated (rotation will not work)

modules/rest-api/pages/security/encryption-at-rest/encryption-at-rest.adoc
```diff
@@ -0,0 +1,58 @@
...
+See xref:learn:security/native-encryption-at-rest-overview.adoc[] for more information.
+
+== APIs in this Section
+
+[cols="76,215,249"]
+|===
+| HTTP Method | URI | Documented at
+
+| `GET`
+| `/settings/encryptionKey/`
```
timofey-barmin: Here and below:

 `/settings/encryptionKey` ->  `/settings/encryptionKeys`

modules/rest-api/pages/security/encryption-at-rest/encryption-at-rest.adoc
```diff
@@ -0,0 +1,58 @@
...
+
+| `GET`
+| `/settings/security/encryptionAtRest`
+| xref:rest-api:security/encryption-at-rest/manage-system-encryption-at-rest.adoc#get-settings[Get Audit, Config, and Log Encryption-at-Rest Settings]
+
+| `POST`
+| `/settings/security/encryptionAtRest`
+| xref:rest-api:security/encryption-at-rest/manage-system-encryption-at-rest.adoc#change-settings[Change Audit, Config, and Log Data Encryption-at-Rest Settings]
+
+// No?
```
timofey-barmin: We need dropEncryptionAtRestKeys and also please add

`POST /controller/forceEncryptionAtRest/bucket/<bucket-name>`

`POST /controller/forceEncryptionAtRest/<type>`

This endpoint forces encryption for corresponding data.
Basically when we enable encryption say for bucket or for logs, we will start encrypting new bucket data (or new logs). Old bucket data (old logs) may remain unencrypted. These ^^ endpoints can be called to trigger encryption for all historical data that is unencrypted.

The difference between `forceEncryptionAtRest` and `dropEncryptionAtRestKeys` is that 'dropEncryptionAtRestKeys' forces re-encryption of all data (and DEK rotation), but  `forceEncryptionAtRest` only encrypts what is not encrypted yet (and doesn't trigger DEK rotation).

modules/rest-api/pages/security/encryption-at-rest/encryption-at-rest.adoc
```diff
@@ -0,0 +1,58 @@
...
+// | `POST`
+// | `/controller/dropEncryptionAtRestKeys/<bucket>`
```
timofey-barmin: `/controller/dropEncryptionAtRestKeys/bucket/<bucket-name>`

modules/rest-api/pages/security/encryption-at-rest/manage-encryption-keys.adoc
```diff
@@ -0,0 +1,819 @@
...
+.Path Parameters
+:priv-link: get-privs
+include::partial$user-pw-host-port-params.adoc[]
+
+`KEY_ID` (optional)::
+The id of the single encryption-at-rest key whose details you want to retrieve.
+
+
+[[get-privs]]
+=== Required Privileges
```
timofey-barmin, line 63: (this comment describes roles that have read access for encryption keys)

It is a bit more complex:
All keys are readable by Full Admin, Read-Only Admin, and Security Admin 

Keys that have bucket encryption usages are readable by any role that has access to that bucket settings (many roles that includes Bucket Admin, Cluster Admin, etc...). 
Here is the full list if you want to put all of them:
"Full Admin"
"Read-Only Admin"
"Security Admin"
"Local User Admin"
"External User Admin"
"Cluster Admin"
"Eventing Full Admin"
"Backup Full Admin"
"Bucket Admin"
"Application Access"
"Views Admin"
"XDCR Admin"
"Data Reader"
"Data Writer"
"Data DCP Reader"
"Data Backup & Restore"
"Data Monitor"
"Search Admin"
"Search Reader"
"Query Select"
"Query Update"
"Query Insert"
"Query Delete"
"Query Manage Index"
"Query List Index"
"Query System Catalog"
"Query CURL Access"
"XDCR Inbound"
"Sync Gateway"

modules/rest-api/pages/security/encryption-at-rest/manage-encryption-keys.adoc
```diff
@@ -0,0 +1,819 @@
...
+[source,ini]
+----
+[profile profile-name]
+region = us-east-1
+output = json
+role_arn = arn:aws:iam::123456789012:role/RoleName
+source_profile = base
+----
+
+* `profile`: Path to a file containing the name of the profile defined in the `credentials-file` to use when authenticating with AWS KMS.
```
timofey-barmin: That is not a path but just a profile name to use

modules/rest-api/pages/security/encryption-at-rest/manage-encryption-keys.adoc
```diff
@@ -0,0 +1,819 @@
...
+
+* `port`: Integer port number for the KMS server. 
+Most KMS servers use port 5696.
+* `host`: The URL for the KMS.
+* `reqTimeoutMs` (integer, optional): The timeout for network communication with the KMS in milliseconds.
+Defaults to 1000.
+* `keyPath`: Absolute path on each server to the private key Couchbase Server uses to authenticate with the KMS.
+This key file must be in PEM format.
+* `keyPassphrase` (optional): The private key's passphrase, if it has one.
+* `encryptWith`: Controls which system performs the decryption of keys.
```
timofey-barmin: `encryptWith` -> `encryptionApproach`

modules/rest-api/pages/security/encryption-at-rest/manage-encryption-keys.adoc
```diff
@@ -0,0 +1,819 @@
...
+** `"useSysCa"`: Use the certificates in the operating system's trust store.
+** `"useCbCa"`: Use the certificates in Couchbase Server's trust store.
+** `"skipServerCertVerification"`: Skip verification of the KMS.
++
+CAUTION: Not verifying the identity of the KMS is insecure. 
+
+* `encryptWith` (optional): controls how the passphrase for the private key is encrypted for local storage. 
+The two options are:
+
+** `"nodeSecretManager"` (default): Couchbase Server encrypts the passphrase using the database's master password.
```
timofey-barmin: `database's master password.` -> `node master password.`

modules/rest-api/pages/security/encryption-at-rest/manage-encryption-keys.adoc
```diff
@@ -0,0 +1,819 @@
...
+.Fields
+
+* `autoRotation` (Boolean, optional):  Controls whether Couchbase Server automatically rotates the key.
+See xref:learn:security/native-encryption-at-rest-overview.adoc#rotation-expiration[Encryption Key Rotation and Expiration] for more information about key rotation.
+Defaults to `true`, which means that Couchbase Server automatically rotates  the key. 
+In this case, you must supply both the `nextRotationTime` and `rotationIntervalInDays` fields as well.
+* `encryptWith` (optional): How Couchbase Server should encrypt this key for storage.
+The two options are:
+
+** `"nodeSecretManager"`: (default): Couchbase Server encrypts the key using the database's master password.
```
timofey-barmin: "database's master password." -> "node master password".

modules/rest-api/pages/security/encryption-at-rest/manage-encryption-keys.adoc
```diff
@@ -0,0 +1,819 @@
...
+** `"encryptionKey"`: Couchbase Server uses a KEK-enabled encryption key to encrypt the key.
+If you choose this option, you must also supply the `encryptWithKeyId` field.
+
+* `encryptWithKeyId` (integer): The `id` attribute of an existing encryption-at-rest key to use to encrypt this key. 
+The key you select must have `KEK-encryption` in its `usage` list.
+* `canBeCached` (Boolean, optional): Determines if Couchbase Server is allowed to cache the decrypted key in memory. 
+Setting this to `true` (the default) makes using the key more efficient because Couchbase Server does not have to decrypt it each time it needs to use it. 
+Disabling caching by setting this value to `false` slightly increases the overhead of using encryption at rest but helps reduce the chance of in-memory key exposure attacks.
```
timofey-barmin: > Disabling caching by setting this value to `false` slightly increases the overhead of using encryption

Disabling caching by setting this value to `false` can significantly increase node start time, as well as time taken by any operation that involves re-encryption of DEKs (such as key rotation, or encryption settings change)

modules/rest-api/pages/security/encryption-at-rest/manage-encryption-keys.adoc
```diff
@@ -0,0 +1,819 @@
...
+Disabling caching by setting this value to `false` slightly increases the overhead of using encryption at rest but helps reduce the chance of in-memory key exposure attacks. 
+* `nextRotationTime`: A date and time string in https://en.wikipedia.org/wiki/ISO_8601[ISO 8601 format^] that sets when Couchbase Server must rotate this key.   
+This value is required if you set `autoRotation` to `true`.
+* `rotationIntervalInDays` (integer): How often, in days, Couchbase Server rotates this key. 
+This field is required if you set `autoRotation` to `true`.
+--
+====
+
+[#create-privs]
+=== Required Privileges
```
timofey-barmin, line 324: It is a bit more complicated. See my comment about about write permissions for keys.

modules/rest-api/pages/security/encryption-at-rest/encryption-at-rest.adoc
```diff
@@ -0,0 +1,58 @@
...
+| `GET`
+| `/settings/encryptionKey/`
+| xref:rest-api:security/encryption-at-rest/manage-encryption-keys.adoc#list-keys[List Encryption-at-Rest Keys]
+
+| `GET`
+| `/settings/encryptionKey/{KEY_ID}`
+| xref:rest-api:security/encryption-at-rest/manage-encryption-keys.adoc#list-keys[List Single Encryption-at-Rest Key]
+
+| `POST`
+| `/settings/encryptionKey`
```
timofey-barmin: there is also `POST /settings/encryptionKey/test`
which tests the key (tries to use it for encryption and decryption)
Parameters are identical `POST /settings/encryptionKey`.

modules/rest-api/pages/security/encryption-at-rest/encryption-at-rest.adoc
```diff
@@ -0,0 +1,58 @@
...
+| xref:rest-api:security/encryption-at-rest/manage-encryption-keys.adoc#create-key[Create an Encryption-at-Rest Key]
+
+| `PUT`
+| `/settings/encryptionKey/{KEY_ID}`
```
timofey-barmin: there is also `PUT /settings/encryptionKey/{KEY_ID}/test`
which tests the key (tries to use it for encryption and decryption)
Parameters are identical to `PUT /settings/encryptionKey/{KEY_ID}`.

modules/rest-api/pages/security/encryption-at-rest/manage-encryption-keys.adoc
```diff
@@ -0,0 +1,819 @@
...
+include::partial$user-pw-host-port-params.adoc[]
+
+`KEY_ID` (integer)::
+ The `id` attribute of the key you want to delete.
+
+
+[[del-privs]]
+=== Required Privileges
+
+You must have at least on one of the following roles:
```
timofey-barmin: Same as in the "(this comment describes roles that have write access for encryption keys)" comment.

modules/rest-api/pages/security/encryption-at-rest/manage-system-encryption-at-rest.adoc
```diff
@@ -0,0 +1,434 @@
...
+
+.Path Parameters
+:priv-link: get-privs
+include::partial$user-pw-host-port-params.adoc[]
+
+
+[[get-privs]]
+=== Required Privileges
+
+You must have at least on one of the following roles:
```
timofey-barmin, line 51: Correction:

Correct roles for `GET /settings/security/encryptionAtRest` are

"Full Admin"
"Read-Only Admin"
"Security Admin"

modules/rest-api/pages/security/encryption-at-rest/manage-system-encryption-at-rest.adoc
```diff
@@ -0,0 +1,434 @@
...
+
+Each type of system data that you can configure to be encrypted has its own object in the output (`audit`, `config`, `log`).
+Some important fields in each of these objects are:
+
+* The `info.dataStatus` field shows whether the data is being encrypted.
+* If Couchbase Server is encrypting the data, `encryptionMethod` indicates what it's using to encrypt it.  
+This value can be:
+** `disabled`: not being encrypted
+** `encryptionKey`: encrypted using an encryption-at-rest key.
+** `nodeSecretManager`: Couchbase Server uses the master database password to encrypt the data.
```
timofey-barmin: "master database password" -> "node master password"

https://docs.couchbase.com/server/current/manage/manage-security/manage-system-secrets.html#setting-the-master-password

modules/rest-api/pages/security/encryption-at-rest/encryption-at-rest.adoc
ggray-cb: I can't seem to get the dropEncryptionAtRestKeys keys mentioned in the key management document to work. I keep getting 404's back from Couchbase when I try to POST to http://localhost:8091/controller/dropEncryptionAtRestKeys/bucket/travel-sample. Has the endpoint been renamed?

ggray-cb: Also a question about this method: after calling this endpoint, is all the data [in the bucket | of the type specified] encrypted with the same DEK (the now active DEK)? Or do we create a new DEK to replace each inactive DEK, re-encrypt the data that was encrypted with the inactive DEK, and then make the new DEK inactive?

modules/manage/pages/manage-security/manage-native-encryption-at-rest.adoc
ggray-cb: So, is there a bug in the Web Console UX? Currently, I see "Rotate and re-encrypt now" as the link text below the Config, Audit, and log entries on the encryption at rest tab. If we want to make a distinction between dropping & rotating, it should be reflected in the UI.

modules/rest-api/pages/security/encryption-at-rest/encryption-at-rest.adoc
ggray-cb: > I can't seem to get the dropEncryptionAtRestKeys keys mentioned in the key management document to work. I keep getting 404's back from Couchbase when I try to POST to http://localhost:8091/controller/dropEncryptionAtRestKeys/bucket/travel-sample. Has the endpoint been renamed?

Nevermind. Found it.

ggray-cb: > Also a question about this method: after calling this endpoint, is all the data [in the bucket | of the type specified] encrypted with the same DEK (the now active DEK)? Or do we create a new DEK to replace each inactive DEK, re-encrypt the data that was encrypted with the inactive DEK, and then make the new DEK inactive?

Looking at other comments, this looks like it's the case? That's what I've updated the descriptions to in the docs.

modules/learn/pages/security/native-encryption-at-rest-overview.adoc
```diff
@@ -1,139 +1,177 @@
...
+You can have Couchbase Server manage encryption keys or have a third-party Key Management Service (KMS) manage them. 
 See <<#kms>> for more information about choosing a KMS.
 
 * What data Couchbase Server can use the encryption key to encrypt. 
 You can restrict the key to encrypting one or more types of data:
 
 +
-** **Other encryption keys** 
+** **Other encryption keys:** 
 Enabling a key to encrypt other encryption-at-rest keys makes it a Key Encryption Key (KEK).
```
timofey-barmin: > Enabling a key to encrypt other encryption-at-rest keys makes it a Key Encryption Key (KEK).

This is a bit confusing.
All types of CB encryption keys are used as KEKs (because they always encrypt other keys - be it DEKs, or other KEKs).
"Other encryption keys" allows encryption of other encryption keys. It doesn't make this key a KEK.

> KEKs are useful when you want to use multiple layers of encryption keys.
> For example, you can use a KEK managed by a third-party KMS as a primary encryption key that encrypts keys managed by Couchbase Server.

This is 100% correct

modules/learn/pages/security/native-encryption-at-rest-overview.adoc
```diff
@@ -1,139 +1,177 @@
...
 
 == Encrypting Audit, Logs, and Configuration Data
 
 In addition to data, you can encrypt audit, logs, and most configuration data. 
-You enable encrypting each type of information separately.
-You can use either an encryption-at-rest key or the database's master password to encrypt data.
-By default, Couchbase Server uses the database's master password to configuration data.
+You enable the encryption of each of these types of data separately.
+You can use either an encryption-at-rest key or the master password to encrypt data.
+By default, Couchbase Server uses the master password to configuration data.
```
timofey-barmin: Typo?

`By default, Couchbase Server uses the master password to encrypt configuration data.`

modules/learn/pages/security/native-encryption-at-rest-overview.adoc
```diff
@@ -1,139 +1,177 @@
...
+
+You can use the REST API to force Couchbase Server to encrypt existing unencrypted data. 
+See xref:rest-api:security/encryption-at-rest/force-encryption-at-rest.adoc[] for more information.
+
+Similarly to enabling encryption, when you turn off encryption at rest, Couchbase Server does not immediately decrypt all encrypted data.
+It stops encrypting the new data it writes but does not decrypt and rewrite the existing encrypted data.
+It still decrypts encrypted data as it reads it.
+It only rewrites data in an unencrypted format when updating data.
+You can use the same REST API call to force Couchbase Server to decrypt and rewrite the all of the data. 
+See xref:rest-api:security/encryption-at-rest/force-encryption-at-rest.adoc[] for more information.
```
timofey-barmin: > You can use the same REST API call to force Couchbase Server to decrypt and rewrite the all of the data. 
> See xref:rest-api:security/encryption-at-rest/force-encryption-at-rest.adoc[] for more information.

This is incorrect. To force decryption xref:rest-api:security/encryption-at-rest/drop-encryption-deks.adoc should be used. 

xref:rest-api:security/encryption-at-rest/force-encryption-at-rest.adoc[] will be a no-op when encryption is disabled.

modules/learn/pages/security/native-encryption-at-rest-overview.adoc
```diff
@@ -1,139 +1,177 @@
...
 [#kms-and-keys]
 == Using Multiple KMSs and Multiple Keys
 
 In a basic setup, you may choose to use a single encryption key managed by Couchbase Server to encrypt all data.
 This configuration is easy to manage, but does not offer much flexibility.
-Also, because Couchbase Server stores the encryption key locally, it does not provide a high level of security.
 
 Couchbase Server does not limit you to a single KMS.
-You can choose any KMS for each encryption key.
+You can choose any KMS onfigured for each encryption key.
```
timofey-barmin: Typo?

modules/manage/pages/manage-security/manage-native-encryption-at-rest.adoc
```diff
@@ -0,0 +1,544 @@
...
+[{tabs}]
+====
+AWS::
++
+--
+[start=8]
+. Enter the Amazon Resource Name (ARN) for the encryption key and the AWS Region in which the KMS is located.
+. Choose whether to use the AWS Instance Metadata Service.
+Enable this option if your Couchbase Server cluster runs on AWS EC2 instances to allow it to connect to other AWS services. 
+. Enter the paths on your cluster where you have stored the AWS credential, configuration, and profile files.
```
timofey-barmin: > Enter the paths on your cluster where you have stored the AWS credential, configuration, and profile files.

Profile here is not a file path, but just a name of the profile.
Due to a bug in UI it was called "profile path" in UI. It should be fixed by now.
AWS key screenshot also needs to be updated.

modules/rest-api/pages/rest-bucket-create.adoc
```diff
@@ -93,11 +93,11 @@ All parameters are described in the following subsections.
 
 == Parameter Groups
 
-Parameters that support the creation and editing of buckets can be broken two groups: Genera and Auto-compaction.
+Parameters that support the creation and editing of buckets can be broken into two groups: Genera and Auto-compaction.
```
timofey-barmin, line 96: > Genera

typo?

modules/rest-api/pages/security/encryption-at-rest/drop-encryption-deks.adoc
```diff
@@ -0,0 +1,200 @@
...
+:tabs:
+:page-toclevels: 3
+
+[abstract]
+{description}
+
+== Description
+
+You may want to drop data encryption keys (DEKs) if you believe they have been compromised.
+When you have Couchbase Server drop DEKs, it creates new a DEK to replace the active and all historical DEKs for the bucket or data type.
```
timofey-barmin: > new a DEK 

typo?

modules/rest-api/pages/security/encryption-at-rest/drop-encryption-deks.adoc
```diff
@@ -0,0 +1,200 @@
...
+This re-encryption includes any unencrypted data that was written before you enabled encryption at rest.
+
+NOTE: If you just want to encrypt any unencrypted data, you can do so without dropping the DEKs. 
+See xref:rest-api:security/encryption-at-rest/force-encryption-at-rest.adoc[] for more information.
+
+Dropping the DEKs is not the same as rotating them.
+When it rotates a DEK, Couchbase Server just creates a new DEK and makes it active.
+It does not re-encrypt data, and it keeps the inactive DEKs for a period of time.
+You cannot manually rotate DEKs.
+When you drop DEKs, Couchbase Server creates new a new DEK, re-encrypts all data with it, and deletes the old DEKs.
```
timofey-barmin, line 28: > new a new DEK

typo?

modules/rest-api/pages/security/encryption-at-rest/drop-encryption-deks.adoc
```diff
@@ -0,0 +1,200 @@
...
+== Description
+
+You may want to drop data encryption keys (DEKs) if you believe they have been compromised.
+When you have Couchbase Server drop DEKs, it creates new a DEK to replace the active and all historical DEKs for the bucket or data type.
+It then re-encrypts all data using the new DEK and deletes the old DEKs.
+This re-encryption includes any unencrypted data that was written before you enabled encryption at rest.
+
+NOTE: If you just want to encrypt any unencrypted data, you can do so without dropping the DEKs. 
+See xref:rest-api:security/encryption-at-rest/force-encryption-at-rest.adoc[] for more information.
+
```
timofey-barmin, line 23: Probably makes sense to add a note about the behavior of this endpoint when encryption is disabled.
It drops all current DEKs but since the encryption is disabled, it just decrypts everything that is currently still encrypted.
In other words it makes sure all data is decrypted after encryption disablement.

modules/rest-api/pages/security/encryption-at-rest/manage-encryption-keys.adoc
```diff
@@ -108,13 +142,13 @@ An example of running the previous command:
           "keyMaterial": "******"
         }
       ],
-      "encryptWith": "nodeSecretManager",
+      "encryptionApproach": "nodeSecretManager",
```
timofey-barmin, line 145: Should be `encryptWith`

modules/rest-api/pages/security/encryption-at-rest/manage-encryption-keys.adoc
```diff
@@ -131,14 +165,14 @@ An example of running the previous command:
       "encryptionApproach": "useGet",
       "certPath": "/scripts/certs/cb_cert.pem",
       "caSelection": "useSysAndCbCa",
-      "encryptWith": "nodeSecretManager",
+      "encryptionApproach": "nodeSecretManager",
```
timofey-barmin, line 168: Should be `encryptWith`

modules/manage/pages/manage-security/manage-native-encryption-at-rest.adoc
```diff
@@ -0,0 +1,544 @@
...
+       --data-binary @- <<EOF | jq
+{
+  "name": "Example KMIP Key",
+  "type": "kmip-aes-key-256",
+  "usage": [
+    "KEK-encryption"
+  ],
+"data": {
+      "port": 5696,
+      "host": "https://kms.example.com",
```
timofey-barmin: No need in `https://`

modules/manage/pages/manage-security/manage-native-encryption-at-rest.adoc
```diff
@@ -0,0 +1,544 @@
...
+----
+
+The result of running this command looks like this:
+
+[source,json]
+----
+{
+  "data": {
+    "port": 5696,
+    "host": "https://kms.example.com",
```
timofey-barmin: should be `"host": "kms.example.com"`

modules/rest-api/pages/security/encryption-at-rest/manage-encryption-keys.adoc
```diff
@@ -0,0 +1,842 @@
...
+    "usage": [
+      "bucket-encryption-beer-sample",
+      "bucket-encryption-test"
+    ],
+    "creationDateTime": "2025-04-23T16:22:39Z"
+  },
+  {
+    "data": {
+      "port": 5696,
+      "host": "https://kms.example.com",
```
timofey-barmin, line 161: should be just `"host": "kms.example.com"`

modules/rest-api/pages/security/encryption-at-rest/manage-encryption-keys.adoc
```diff
@@ -0,0 +1,842 @@
...
+       --data-binary @- <<EOF | jq
+{
+  "name": "Example KMIP Key",
+  "type": "kmip-aes-key-256",
+  "usage": [
+    "KEK-encryption"
+  ],
+"data": {
+      "port": 5696,
+      "host": "https://kms.example.com",
```
timofey-barmin, line 452: Should be just `"host": "kms.example.com"`

modules/rest-api/pages/security/encryption-at-rest/manage-encryption-keys.adoc
```diff
@@ -0,0 +1,842 @@
...
+----
+
+The following example shows the output from running the previous example.
+
+[source,json]
+----
+{
+  "data": {
+    "port": 5696,
+    "host": "https://kms.example.com",
```
timofey-barmin, line 475: Should be `    "host": "kms.example.com",`

modules/rest-api/pages/security/encryption-at-rest/manage-encryption-keys.adoc
```diff
@@ -131,14 +165,14 @@ An example of running the previous command:
...
-      "encryptWith": "nodeSecretManager",
+      "encryptionApproach": "nodeSecretManager",
       "historicalKeys": [],
       "activeKey": {
         "id": "0788edb1-1418-4225-903b-bc1f9f59aa4d",
         "kmipId": "550e8400-e29b-41d4-a716-446655440000",
         "creationDateTime": "2025-04-23T14:44:20Z"
       },
-      "encryptWithKeyId": -1
+      "encryptionApproachKeyId": -1
```
timofey-barmin, line 175: Should be `encryptWithKeyId`

modules/rest-api/pages/security/encryption-at-rest/manage-encryption-keys.adoc
```diff
@@ -526,15 +336,15 @@ See <<#create-examples>> for examples of the returned JSON.
 `400 Bad Request`::
 Returned when errors occur, such as leaving out a required setting or invalid JSON. 
 Also returns a descriptive JSON message.
-For example, if you set `encryptWith` to `encryptionKey` but do not set `encryptWithKeyID`, you receive this message:
+For example, if you set `encryptionApproach` to `encryptionKey` but do not set `encryptionApproachKeyID`, you receive this message:
```
timofey-barmin, line 339: > For example, if you set `encryptionApproach` to `encryptionKey` but do not set `encryptionApproachKeyID`, you receive this message:

Should be:
For example, if you set `encryptWith` to `encryptionKey` but do not set `encryptWithKeyID`, you receive this message:

modules/rest-api/pages/security/encryption-at-rest/manage-encryption-keys.adoc
```diff
@@ -526,15 +336,15 @@ See <<#create-examples>> for examples of the returned JSON.
...
 
 +
 [source,json]
 ----
 {
   "errors": {
     "data": {
-      "encryptWith": "encryptWithKeyId must be set when 'encryptionKey' is used"
+      "encryptionApproach": "encryptionApproachKeyId must be set when 'encryptionKey' is used"
```
timofey-barmin, line 347: Should be:
`      "encryptWith": "encryptWithKeyId must be set when 'encryptionKey' is used"`

modules/rest-api/pages/security/encryption-at-rest/manage-encryption-keys.adoc
```diff
@@ -545,7 +355,7 @@ Returned if you do not have the proper roles to call this API.
 See <<#create-privs>>.
 
 `404 Object Not Found`::
-Returned if you specified a `encryptWithKeyId` field or a `KEY_ID` path parameter which does not match the `id` field of an existing encryption-at-rest key.
+Returned if you specified a `encryptionApproachKeyId` field or a `KEY_ID` path parameter which does not match the `id` field of an existing encryption-at-rest key.
```
timofey-barmin, line 358: Should be:
```
Returned if you specified a `encryptWithKeyId` field or a `KEY_ID` path parameter which does not match the `id` field of an existing encryption-at-rest key.

```

modules/rest-api/pages/security/encryption-at-rest/manage-encryption-keys.adoc
```diff
@@ -726,15 +537,15 @@ The output from the previous example looks like this:
         "keyMaterial": "******"
       }
     ],
-    "encryptWith": "nodeSecretManager",
+    "encryptionApproach": "nodeSecretManager",
```
timofey-barmin, line 540: Should be:
```
    "encryptWith": "nodeSecretManager",
```

modules/rest-api/pages/security/encryption-at-rest/rotate-encryption-at-rest-key.adoc
```diff
@@ -95,13 +96,13 @@ The following example rotates the encryption-at-rest key for the named `Example
         "keyMaterial": "******"
       }
     ],
-    "encryptWith": "nodeSecretManager",
+    "encryptionApproach": "nodeSecretManager",
```
timofey-barmin, line 99: Should be:
```
    "encryptWith": "nodeSecretManager",
```

modules/rest-api/pages/security/encryption-at-rest/rotate-encryption-at-rest-key.adoc
```diff
@@ -151,14 +152,14 @@ The JSON returned by this command shows the encryption-at-rest has a new key wit
...
```
timofey-barmin, line 155: Should be:
```
    "encryptWith": "nodeSecretManager",
```

modules/rest-api/partials/encryption-at-rest-post-put-fields.adoc
```diff
@@ -0,0 +1,228 @@
...
+
+When using the AWS KMS, the `data` object has the following schema:
+
+[source,json]
+----
+  "data": {
+    "keyARN": "<arn>",
+    "region": "<region>",
+    "useIMDS": <true|false>,
+    "profile": "<profile-file-path>",
```
timofey-barmin, line 47: Should be:
```
"profile": "<profile>"
```

modules/rest-api/partials/encryption-at-rest-post-put-fields.adoc
```diff
@@ -0,0 +1,228 @@
...
+KMIP KMS::
++
+--
+When using a KMIP-compatible KMS, the `data` object has the following schema:  
+
+[source,json]
+----
+"data": {
+      "port": <kms-port>,
+      "host": "<kms-host-url>",
```
timofey-barmin, line 119: It is not a URL, just FQDN or IP address

modules/rest-api/partials/encryption-at-rest-post-put-fields.adoc
```diff
@@ -0,0 +1,228 @@
...
+      },
+      "encryptionApproachKeyId": <local-KEY_ID>
+    }
+----
+
+.Fields
+
+* `port`: Integer port number for the KMS server. 
+Most KMS servers use port 5696.
+* `host`: The URL for the KMS.
```
timofey-barmin, line 138: Not a URL, just FQDN or IP

modules/rest-api/partials/encryption-at-rest-post-put-fields.adoc
```diff
@@ -0,0 +1,228 @@
...
+Can be one of the following values:
+
+** `"useSysAndCbCa"` (default): Use the certificates in both the operating system's and Couchbase Server's trust stores.
+** `"useSysCa"`: Use the certificates in the operating system's trust store.
+** `"useCbCa"`: Use the certificates in Couchbase Server's trust store.
+** `"skipServerCertVerification"`: Skip verification of the KMS.
++
+CAUTION: Not verifying the identity of the KMS is insecure. 
+
+* `encryptionApproach` (optional): controls how the passphrase for the private key is encrypted for local storage.
```
timofey-barmin, line 163: Should be:
```
* `encryptWith` (optional): controls how the passphrase for the private key is encrypted for local storage.
```

modules/rest-api/partials/encryption-at-rest-post-put-fields.adoc
```diff
@@ -0,0 +1,228 @@
...
+** `"skipServerCertVerification"`: Skip verification of the KMS.
++
+CAUTION: Not verifying the identity of the KMS is insecure. 
+
+* `encryptionApproach` (optional): controls how the passphrase for the private key is encrypted for local storage. 
+The two options are:
+
+** `"nodeSecretManager"` (default): Couchbase Server encrypts the passphrase using the xref:manage:manage-security/manage-system-secrets.adoc#setting-the-master-password[master password].
+** `"encryptionKey"`: Use a KEK-enabled encryption key to encrypt the passphrase.
+If you choose this option, you must also supply the `encryptionApproachKeyId` parameter.
```
timofey-barmin, line 168: Should be:
```
If you choose this option, you must also supply the `encryptWithKeyId` parameter.
```

modules/rest-api/partials/encryption-at-rest-post-put-fields.adoc
```diff
@@ -0,0 +1,228 @@
...
+
+* `activeKey.kmipId`: The ID of the encryption key stored in the KMS.
+The format of this value depends on the KMS.
+It's often in the form of a UUID or a friendly name.
+
+* `encryptionApproachKeyId` (integer): The `id` attribute of the encryption key Couchbase Server uses to encrypt the private key's passphrase when storing it locally.
```
timofey-barmin: Should be:
```
* `encryptWithKeyId` (integer): The `id` attribute of the encryption key Couchbase Server uses to encrypt the private key's passphrase when storing it locally. 
```

modules/rest-api/partials/encryption-at-rest-post-put-fields.adoc
```diff
@@ -0,0 +1,228 @@
...
+** `"encryptionKey"`: Use a KEK-enabled encryption key to encrypt the passphrase.
+If you choose this option, you must also supply the `encryptionApproachKeyId` parameter.
+
+* `activeKey.kmipId`: The ID of the encryption key stored in the KMS.
+The format of this value depends on the KMS.
+It's often in the form of a UUID or a friendly name.
+
+* `encryptionApproachKeyId` (integer): The `id` attribute of the encryption key Couchbase Server uses to encrypt the private key's passphrase when storing it locally. 
+See <<#list-keys>> to learn how to get an encryption key's `id`.
+Required if you set `encryptionApproach` to `encryptionKey`.
```
timofey-barmin: Should be:
```
Required if you set `encryptWith` to `encryptionKey`. 
```

modules/manage/pages/manage-security/manage-native-encryption-at-rest.adoc
timofey-barmin: This comment is not addressed.

modules/manage/pages/manage-security/manage-native-encryption-at-rest.adoc
timofey-barmin: I meant the distinction between "dropping&rotating DEKs" vs. "rotation of encryption keys".
Not between "dropping DEKs" vs "rotating DEKs".

modules/rest-api/pages/security/encryption-at-rest/encryption-at-rest.adoc
timofey-barmin: > Also a question about this method: after calling this endpoint, is all the data [in the bucket | of the type specified] encrypted with the same DEK (the now active DEK)? Or do we create a new DEK to replace each inactive DEK, re-encrypt the data that was encrypted with the inactive DEK, and then make the new DEK inactive?

We do the following:
1. Generate new DEK and make it "active"
2. Reencrypt all data with the current active DEK
3. Remove old DEKs

Usually, in the end, all data will be encrypted with one DEK.
But it can happen that automatic DEK rotation gets triggered during the re-encryption process (as re-encryption can take a long time). In this case one more DEK will be generated, and after that that new DEK will be used for encryption.
So in the end of re-encryption we can have more than one DEK, but all of them will be generated after the `/controller/dropEncryptionAtRestKeys` call.

modules/rest-api/pages/security/encryption-at-rest/manage-encryption-keys.adoc
timofey-barmin: Seems like you renamed it everywhere but that was a mistake... encryptionApproach is only used in KMIP.
Moreover, KMIP also has `encryptWith`.
These are two different fields.

PR #3812, opened 2025-06-02
[DOC-13280]: The Whats New page looks odd

RayOffiah:
Fixed the page and rearranged some content.


osfameron (approved):
Not quite sure what the fix was, as there's also rearrangement, but ran a local `preview` and doesn't look squiffy 👍

PR #3839, opened 2025-07-26
DOC-12402 Edit doc section XDCR Compression Type Values

rao-shwe:
DOC-12402

Changed instances of the word "provided" to "attempted".

No doc preview as this is an XS sized doc.

### DIFF
--- a/modules/xdcr-reference/pages/xdcr-advanced-settings.adoc
+++ b/modules/xdcr-reference/pages/xdcr-advanced-settings.adoc
@@ -119,13 +119,13 @@ For more information, see xref:learn:clusters-and-availability/xdcr-conflict-res
 
 The _XDCR Compression Type_ setting can be either `Auto` or `None`:
 
-* If `Auto` is specified, data compression _is_ provided during replication, provided that the target cluster is running Couchbase Server 5.5 or later.
-If the target cluster is _not_ running Couchbase Server 5.5 or later, data compression is _not_ provided during replication.
+* If `Auto` is specified, data compression _is_ attempted during replication, given that the target cluster is running Couchbase Server 5.5 or later.
+If the target cluster is _not_ running Couchbase Server 5.5 or later, data compression is _not_ attempted during replication.
 (Note that the consequences of specifying `Auto` as the value of `compressionType` are thereby determined irrespective of whether the data resides in the source bucket in compressed or uncompressed form.)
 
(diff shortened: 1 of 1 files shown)


modules/xdcr-reference/pages/xdcr-advanced-settings.adoc
```diff
@@ -119,13 +119,13 @@ For more information, see xref:learn:clusters-and-availability/xdcr-conflict-res
 
 The _XDCR Compression Type_ setting can be either `Auto` or `None`:
 
-* If `Auto` is specified, data compression _is_ provided during replication, provided that the target cluster is running Couchbase Server 5.5 or later.
-If the target cluster is _not_ running Couchbase Server 5.5 or later, data compression is _not_ provided during replication.
+* If `Auto` is specified, data compression _is_ attempted during replication, given that the target cluster is running Couchbase Server 5.5 or later.
```
Aditya-Sood, line 122: I think we can italicise `attempted` also here? like:
"...data compression _is attempted_ during replication..."

otherwise the emphasis on "attempt" (in the sense that it is not guaranteed) could be lost

modules/xdcr-reference/pages/xdcr-advanced-settings.adoc
```diff
@@ -119,13 +119,13 @@ For more information, see xref:learn:clusters-and-availability/xdcr-conflict-res
...
 
-* If `None` is specified, data compression is not provided during replication if the target cluster is not running Couchbase Server 5.5 or later.
+* If `None` is specified, data compression is not attempted during replication if the target cluster is not running Couchbase Server 5.5 or later.
 If the target cluster _is_ running Couchbase Server 5.5 or later:
 
-** If the data resides in the source bucket in uncompressed form, data compression is _not_ provided.
+** If the data resides in the source bucket in uncompressed form, data compression is _not_ attempted.
 
-** If the data resides in the source bucket in _compressed_ form, data compression _is_ provided.
+** If the data resides in the source bucket in _compressed_ form, data compression _is_ attempted.
```
Aditya-Sood, line 131: 1. same here, can we italicise `attempted` as well

2. there is also another line in the doc ticket clarifying the use of the wording:
> _(By saying "compression is attempted" we mean that the data will be snappy-compressed only if the encoded size is lesser than the raw size.)_

Can we add that at the bottom as well?

PR #3874, opened 2025-09-03
Fixing broken What's New page and misc.

ggray-cb:
* Fixed broken What's New page.
* Moved the Native Encryption at Rest entry under Security and Authentication.
* Also removed "Overview" from the Native Encryption at Rest title.


RayOffiah (commented):
Yup, looks fine to me. 👍🏾

PR #3882, opened 2025-09-17
DOC-11938 RBAC-API-for-creating-external-user-is-wrong-in-the-lookup-page

supritha-kumar:
'/settings/rbac/users/external/<new-username>' needs to be updated at https://docs.couchbase.com/server/current/rest-api/rbac.html#create-an-external-user-and-assign-roles

### DIFF
--- a/modules/rest-api/pages/rbac.adoc
+++ b/modules/rest-api/pages/rbac.adoc
@@ -431,7 +431,7 @@ If successful, the call creates local user `sdavis` and adds them to the `Cluste
 [#create-an-external-user-and-assign-roles]
 === Create an External User, and Assign Roles
 
-To create an external user, and assign them one or more roles, use the `PUT` method with the `/settings/rbac/users/local/<new-username>` URI.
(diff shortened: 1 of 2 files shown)


modules/rest-api/pages/rbac.adoc
```diff
@@ -431,7 +431,7 @@ If successful, the call creates local user `sdavis` and adds them to the `Cluste
 [#create-an-external-user-and-assign-roles]
 === Create an External User, and Assign Roles
 
-To create an external user, and assign them one or more roles, use the `PUT` method with the `/settings/rbac/users/external/<new-username>` URI.
+To create an external user, and assign them one or more roles, use the `PUT` method with the `/settings/rbac/users/local/<new-username>` URI.
```
ggray-cb: I believe this is wrong. This is discussing creating an external user. The /external/ should be part of this path. I believe all you need is the other change in the PR.

ggray-cb (changes requested):

modules/rest-api/pages/rbac.adoc
supritha-kumar: Have reverted the change.

ggray-cb (approved):

PR #3907, opened 2025-10-03
DOC-13599]: Feedback on Creating or Editing a Reference | Couchbase Docs

RayOffiah:
Changed menu heading to make it clearer that the references can be edited as well as being created from scratch.


RichardSmedley (approved):
:+1:

PR #3951, opened 2025-10-29
Doc 13590 mention FLE on encryption page

ggray-cb:
Adds references to the SDK's field-level encryption to the encryption at rest docs.

- In the [Encryption](https://preview.docs-test.couchbase.com/docs-server-DOC-13590_mention_FLE_on_encryption_page/server/current/learn/security/encryption-overview.html) page, moved the standalone section of field-level encryption to the encryption at rest discussion.
- In the [Native Encryption at Rest](https://preview.docs-test.couchbase.com/docs-server-DOC-13590_mention_FLE_on_encryption_page/server/current/learn/security/native-encryption-at-rest-overview.html) page, added note about field-level encryption as a possible alternative.

Note: couldn't manage to use a partial for both of these, even though the text is almost identical. Antora's fussy formatting strikes again.

### DIFF
--- a/modules/learn/pages/security/native-encryption-at-rest-overview.adoc
+++ b/modules/learn/pages/security/native-encryption-at-rest-overview.adoc
@@ -9,6 +9,19 @@ This feature is transparent to the database's users.
 Couchbase Server automatically decrypts data when reading it from disk and encrypts it when writing it to disk.
 For steps to take when managing this feature, see xref:manage:manage-security/manage-native-encryption-at-rest.adoc[].
 
+[NOTE]
+====
+Applications can use the SDK to encrypt specific fields.
+Depending on your application's requirements, field-level encryption may be more appropriate than encrypting the entire bucket.
+See the SDK documentation for your development language for more information.
+For example:
+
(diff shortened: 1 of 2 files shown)


modules/learn/pages/security/native-encryption-at-rest-overview.adoc
```diff
@@ -9,6 +9,20 @@ This feature is transparent to the database's users.
...
 
+[NOTE]
+.Field-Level Encryption in Applications
+====
+Applications can use the SDK to encrypt specific fields.
+Depending on your application's requirements, field-level encryption may be more appropriate than encrypting the entire bucket.
+See the SDK documentation for your development language for more information.
+For example:
+
+* Go SDK: xref:go-sdk:howtos:encrypting-using-sdk.adoc[]
```
RichardSmedley, line 20: I was going to suggest a concise way of linking to *all* of the SDKs,
but FLE is, it turns out, currently only implemented in exactly half of them. :facepalm:

RichardSmedley (approved):
:+1:

PR #3972, opened 2025-11-19
Suggest using Disk I/O optimized mode for magma

TimLFletcher:
https://jira.issues.couchbase.com/browse/DOC-9868


rao-shwe (approved):
Use ventilated prose (use line breaks for each sentence).

The rest LGTM! 👍

PR #3984, opened 2025-11-25
DOC-12481 Updated max threads

TimLFletcher:
DOC-12481

Tiny change to update number of max threads that changed from 7.0 onward


rao-shwe (approved):
All good! 👍

PR #3987, opened 2025-11-25
DOC-12563 fixed mcstat info for 7.6

TimLFletcher:
DOC-12563

The doc was a bit malformed, stating the description twice. The description needed an update as well as info on the help command, which was pretty much just wrong. I've reworded the dev's advice.

This is for 7.6


rao-shwe (approved):
LGTM! 👍

PR #4001, opened 2025-11-28
[DOC-12426]: Feedback on Facets | Couchbase Docs

RayOffiah:
Added link to an explanation of index partitioning.

### DIFF
--- a/modules/fts/pages/fts-search-response-facets.adoc
+++ b/modules/fts/pages/fts-search-response-facets.adoc
@@ -8,8 +8,8 @@ In Facets, you already have a search in mind, and you want to collect additional
 
 Facet-query results may not equal the total number of documents across all buckets if: 
 
-1. There is more than one pindex. 
-2. Facets_size is less than the possible values for the field.
+1. There is more than one search index partition (`pindex`).
+2. `Facets_size` is less than the possible values for the field.
 
 == Facets Results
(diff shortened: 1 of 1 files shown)


sarahlwelton (commented):
I'm questioning why we're making this change because this page isn't even included in the side nav for the docs. This might require more thought than just doing this small change. 

Nav is here: https://github.com/couchbaselabs/docs-devex/blob/0d992dd9e653f0b39e51f018b360d77faa352378/modules/search/partials/nav.adoc

If you want to give this to me, I can make the more appropriate updates so the info on this page isn't lost, but it ends up in the right place in the rewritten docs for Search. The intention was not to keep limping this page along.

RayOffiah:
> I'm questioning why we're making this change because this page isn't even included in the side nav for the docs. This might require more thought than just doing this small change.
> 
> Nav is here: https://github.com/couchbaselabs/docs-devex/blob/0d992dd9e653f0b39e51f018b360d77faa352378/modules/search/partials/nav.adoc
> 
> If you want to give this to me, I can make the more appropriate updates so the info on this page isn't lost, but it ends up in the right place in the rewritten docs for Search. The intention was not to keep limping this page along.

Yup, makes sense 👍🏾

I'll close this PR and reassign the ticket to you.

PR #4006, opened 2025-12-01
[DOC-12324]: Configure Client Certificates page | Java Client Access

RayOffiah:
Added missing section: create the `client.ext` file.

### DIFF
--- a/modules/manage/pages/manage-security/configure-client-certificates.adoc
+++ b/modules/manage/pages/manage-security/configure-client-certificates.adoc
@@ -349,6 +349,11 @@ keytool -certreq -alias selfsigned -keyalg RSA -file my.csr \
 ----
 +
 This creates the signing-request file, `my.csr`.
++
+Note that in this example, although only the `Common Name` is being used to establish the identity of the user seeking authorization, one or more `Subject Alternative Names` could also be added.
+For example, by adding `-ext "san=email:john.smith@mail.com"` to the certificate signing-request used in the current step, the email-address `john.smith@mail.com` could be established as the basis for an alternative username to be submitted for authentication.
+See xref:learn:security/certificates.adoc#identity-encoding-in-client-certificates[Specifying Usernames for Client-Certificate Authentication], for more information.
+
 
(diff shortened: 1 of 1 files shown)


modules/manage/pages/manage-security/configure-client-certificates.adoc
```diff
@@ -349,6 +349,17 @@ keytool -certreq -alias selfsigned -keyalg RSA -file my.csr \
...
+(`extendedKeyUsage = clientAuth` means that the certificate will be used for client authentication):
++
+[source,bash]
+----
+cat > client.ext <<EOF
+extendedKeyUsage = clientAuth
+keyUsage = digitalSignature,keyEncipherment
+EOF
+----
 +
```
RichardSmedley, line 368: Doesn't the note apply to the previous step - creating the csr?

RayOffiah: Yes!

modules/manage/pages/manage-security/configure-client-certificates.adoc
```diff
@@ -354,6 +354,18 @@ Note that in this example, although only the `Common Name` is being used to esta
 For example, by adding `-ext "san=email:john.smith@mail.com"` to the certificate signing-request used in the current step, the email-address `john.smith@mail.com` could be established as the basis for an alternative username to be submitted for authentication.
 See xref:learn:security/certificates.adoc#identity-encoding-in-client-certificates[Specifying Usernames for Client-Certificate Authentication], for more information.
 
+
+. Create an  extensions file.
+(`extendedKeyUsage = clientAuth` means that the certificate will be used for client authentication):
++
+[source,bash]
```
RichardSmedley, line 361: We don't have source specified elsewhere on this page, but in any case it should be console, as these are commands given at the terminal;
bash is used for marking .sh scripts in source.

Not a blocker - the whole page should be revisited some time (and others across the Server docs and elsewhere.)

RichardSmedley (approved):
:+1:

PR #4023, opened 2025-12-05
DOC-13786 Remove unwanted screenshots from the backup and restore pages

rao-shwe:
DOC-13786


rao-shwe:
This task can be taken up some other time with a new PR.

PR #4028, opened 2025-12-08
[DOC-12771]: Feedback on Filter a Replication | Couchbase Docs

RayOffiah:
Clarifying that the `Binary Documents` option is checked for binary documents to be replicated.

Please check to make sure that the other options in this section still hold. If they don't, I'll fix 'em.


hyunjuV (commented):
The documentation is correct.
It should not be changed.
The tooltip in the UI is incorrect.
Please see my comment here -- https://jira.issues.couchbase.com/browse/DOC-12771?focusedCommentId=1310876

RayOffiah:
Not required. The problem is the UI, not the docs.

PR #4032, opened 2025-12-12
[DOC-13127]: Docs currently incorrectly state that the backup plan task array can be empty

RayOffiah:
The text now makes it clear that the plan must have at least one task when it's created.


rao-shwe (approved):
LGTM! 👍

PR #4034, opened 2025-12-12
[DOC-13831]: Increase port range on `docker run` command to cover the backup REST API

RayOffiah:
Changed the port configurations to cover port 8097, which is used for the backup REST APIs


rao-shwe (approved):
LGTM! 👍

PR #4036, opened 2025-12-16
Remove tools-ver parameter from install pages

rao-shwe:
DOC-13681

`:{tools-ver}: 8.0.0` was added in the pages' head matter.

`{tools-ver}` was added in the body to call the parameter for `8.0.0` version number automatically. But in the output docs, instead of `8.0.0`, the content was displayed as `{tools-ver}`.


rao-shwe:
This PR won't need an editorial review as this was a suggested change during an editorial review and the suggestion didn't work.

PR #4041, opened 2025-12-18
[DOC-12408]: Feedback on Index Service Metrics | Couchbase Docs

RayOffiah:
Additional details for index cache hits and misses.

Minor fix for the template.

Preview URL:
https://preview.docs-test.couchbase.com/docs-server-DOC-12408/release/8.0/Feedback-on-Index-Service-Metrics-/server/current/metrics-reference/index-service-metrics.html

### DIFF
--- a/modules/metrics-reference/attachments/index_metrics_metadata.json
+++ b/modules/metrics-reference/attachments/index_metrics_metadata.json
@@ -54,7 +54,7 @@
     "unit": "Nanoseconds"
   },
   "index_cache_hits": {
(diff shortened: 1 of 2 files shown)


osfameron (approved):

osfameron (changes requested):
er wait, we shouldn't edit that file directly

osfameron:
e.g. the feedback needs to go to the devs, to add to the Index service metrics on their end.
When they build, we can pick up the latest version from the build tarball.

(If we update this file directly, then the next time the devs update the metrics and we update the tarball, we'll just lose it again.)

RayOffiah:
Ah, right okay. 

I'll have a chat with the devs. 👍🏾

PR #4051, opened 2025-12-24
DOC-11359 In Ports description, ports for Analytics service seems to be confused

supritha-kumar:
Updated Analytics port info as per

https://jira.issues.couchbase.com/browse/DOC-11359


ggray-cb (approved):
LGTM

PR #4053, opened 2025-12-26
DOC-13101 Feedback on Set GSI Settings

supritha-kumar:
https://jira.issues.couchbase.com/browse/DOC-13101

Removed the line indicating that bloom filters are disabled by default.

### DIFF
--- a/modules/rest-api/pages/post-settings-indexes.adoc
+++ b/modules/rest-api/pages/post-settings-indexes.adoc
@@ -92,7 +92,6 @@ For further details, refer to {rebalancing-the-index-service}[Rebalancing the In
 
 | `enablePageBloomFilter`
 | Whether Bloom filters are enabled for memory management.
-The default is that they are _disabled_.
 See xref:indexes:storage-modes.adoc#per-page-bloom-filters[Per Page Bloom Filters].
 | boolean
 


modules/rest-api/pages/post-settings-indexes.adoc
```diff
@@ -92,7 +92,6 @@ For further details, refer to {rebalancing-the-index-service}[Rebalancing the In
 
 | `enablePageBloomFilter`
 | Whether Bloom filters are enabled for memory management.
-The default is that they are _disabled_.
```
ggray-cb, line 95: Do we want to say they are enabled by default here?

ggray-cb (changes requested):
I have a question/suggestion about adding the default setting.

modules/rest-api/pages/post-settings-indexes.adoc
supritha-kumar: @ggray-cb I specifically did not call that out here as the very next line redirects to https://docs.couchbase.com/server/current/indexes/storage-modes.html#per-page-bloom-filters which has the following note.
Leaving it as is so that we don't have to update in case things change again.

From release 7.2.1 onward, Bloom filters for plasma back indexes will be enabled by default. During an upgrade, mixed mode clusters with nodes that support bloom filters will enable it for back indexes, even if they was disabled in the past. Users must explicitly disable it again after a cluster setup, or a new node is added.
Once bloom filters are disabled in mixed mode, adding a new 7.2.1+ node will not re-enable them.

ggray-cb (approved):

PR #4058, opened 2026-01-07
DOC-13790 Server no rolling upgrade possible

supritha-kumar:
Added a note about pausing Eventing functions before rolling an upgrade
https://jira.issues.couchbase.com/browse/DOC-13790

### DIFF
--- a/modules/install/pages/upgrade-cluster-online-full-capacity.adoc
+++ b/modules/install/pages/upgrade-cluster-online-full-capacity.adoc
@@ -30,6 +30,9 @@ For information on _swap rebalance_, see xref:install:upgrade-procedure-selectio
 
 * Nodes will be upgraded _one at a time_.
 
+NOTE: In Couchbase Server versions 8.0 or earlier, running Eventing functions while rolling an upgrade increases rebalance duration.
+Pause all your Eventing functions before starting the upgrade to reduce rebalance time, and resume them after the upgrade completes.
+
 The overall procedure is described in seven stages, below.
 
 [#prepare-a-spare-node]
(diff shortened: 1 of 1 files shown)


modules/install/pages/upgrade-cluster-online-full-capacity.adoc
```diff
@@ -30,6 +30,9 @@ For information on _swap rebalance_, see xref:install:upgrade-procedure-selectio
 
 * Nodes will be upgraded _one at a time_.
 
+NOTE: In Couchbase Server versions 8.0 or earlier, running Eventing functions while rolling an upgrade increases rebalance duration.
+Pause all your Eventing functions before starting the upgrade to reduce rebalance time, and resume them after the upgrade completes.
```
simon-dew: I'm not sure why we need to mention the version. We usually only mention the version when contrasting new behavior with existing behavior. Do we know if this behavior will change in 8.1? If so, we should mention it in the 8.1 docs rather than here.


```suggestion
NOTE: Running Eventing functions during a rolling upgrade increases rebalance duration.
To reduce rebalance time, pause all your Eventing functions before starting the upgrade, and resume them after the upgrade completes.
```

simon-dew (changes requested):
I have one question about the version number, and reordered the wording to emphasise the reason for the admonition.

supritha-kumar:
> I have one question about the version number, and reordered the wording to emphasise the reason for the admonition.

Thank you Simon. Have committed your changes.

simon-dew (approved):
LGTM 👍

PR #4063, opened 2026-01-09
[DOC-13414]: Change alog_sleep_time example value

RayOffiah:
Added note explaining the scanner always scans the entire key table, so increasing the frequency of the scan will not decrease the amount of work it is doing.

Changed example `alog_sleep_time` to 2880 minutes, which is a more realistic example.


rao-shwe (approved):
LGTM! 👍

PR #4078, opened 2026-01-26
[DOC-9267] Guidance for Search Service Sizing

sarahlwelton:
The long-awaited guidance for sizing the Search Service in a Couchbase Server deployment. 

Preview URL:
https://preview.docs-test.couchbase.com/docs-server-DOC-9267-fts-sizing/server/current/install/sizing-general.html#sizing-search-service-nodes


modules/install/pages/sizing-general.adoc
```diff
@@ -534,6 +534,276 @@ NOTE: The storage engine used in the sizing calculation corresponds to the stora
...
+|258.05
+|87,000
+|0.10
+|2 (1 replica + 1)
+
+|====
+
+Based on these variables, the required vCPUs could be either: 
+
+* stem:[290], using a value of 300 in the vCPU calculation.
```
TusharMadaan04: We should have two cases for 150 & 200 respectively instead of 200 & 300 as 150 & 200 are the values of QPS/vCPU that we have documented earlier

modules/install/pages/sizing-general.adoc
```diff
@@ -534,6 +534,276 @@ NOTE: The storage engine used in the sizing calculation corresponds to the stora
...
+
+|500,000,000
+|344.86 (more faceting, sorting, and more complex queries)
+|12,000
+|0.10
+|2 (1 replica + 1)
+
+|====
+
+Based on these variables, the required vCPUs would be stem:[40], based on the more complex queries needing a higher QPS per vCPU and using a value of 300 in the calculation.
```
TusharMadaan04: Same here, the value should be 200

modules/install/pages/sizing-general.adoc
```diff
@@ -534,6 +534,280 @@ NOTE: The storage engine used in the sizing calculation corresponds to the stora
 | Nitro
 |===
 
+== Sizing Search Service Nodes
+
+Search Service nodes manage Search indexes and serve your Search queries. 
+
+Basic Search indexes are lists of all the unique terms that appear in the documents on your cluster.
+For each term, the Search index also contains a list of the documents where that term appears.
```
abhinavdangeti: It'd be good to use the nomenclature - `inverted index` here .. which is the primary data structure within the search index that indicates the list of the documents where the term appears.

modules/install/pages/sizing-general.adoc
```diff
@@ -534,6 +534,280 @@ NOTE: The storage engine used in the sizing calculation corresponds to the stora
...
+These lists inside a Search index can cause the Search index to be larger than your original dataset.
```
abhinavdangeti: This is a possibility, it depends very much on the data though .. in many a situation it could be lesser too because we de-dup terms/prefixes/sub-strings/suffixes in the index and the inverted-index/postings-lists are compressed bitmaps alongside a map for doc keys.

Will defer to your judgement on how to frame all of this :)

modules/install/pages/sizing-general.adoc
```diff
@@ -534,6 +534,280 @@ NOTE: The storage engine used in the sizing calculation corresponds to the stora
...
+For more information about what options can increase index size and storage requirements, see xref:search:child-field-options-reference.adoc[].
+
+In general, when sizing nodes for a deployment that uses the Search Service, you need to determine the number of vCPUs and the amount of RAM that will support your workload.
+
+=== Calculating Node Requirements
+
+To size the Search Service nodes in your cluster, you need the following information: 
+
+* The number of documents you need to include in your Search index or indexes.
+* The average size of the documents that need to be included in your Search index, in KB.
```
abhinavdangeti, line 556: Not accurate enough - it's not the average size of the docs that we need, but the number of the fields and their sizes that'll drive footprint.

modules/install/pages/sizing-general.adoc
```diff
@@ -534,6 +534,280 @@ NOTE: The storage engine used in the sizing calculation corresponds to the stora
...
+=== Calculating Node Requirements
+
+To size the Search Service nodes in your cluster, you need the following information: 
+
+* The number of documents you need to include in your Search index or indexes.
+* The average size of the documents that need to be included in your Search index, in KB. 
+* A sample document or documents that show the structure of your data. 
+* The specific queries per second (QPS) target you need from the Search Service.
+
+You should also consider your replication and recovery needs.
```
abhinavdangeti: recovery/high-availability* needs

abhinavdangeti (commented):
@sarahlwelton I've some comments around search index nomenclature and requirements for sizing right.

I think the estimation itself @TusharMadaan04 's done nice work on and you've captured it quite well here.

modules/install/pages/sizing-general.adoc
sarahlwelton: Perhaps this framing works better, just to keep things accurate?

modules/install/pages/sizing-general.adoc
sarahlwelton: Sure, that's easy enough to add

modules/install/pages/sizing-general.adoc
sarahlwelton: But we do use this "average size of doc" value in the formula further down on the page. For just a rough estimate, I'd think that's "good enough"?

abhinavdangeti: I suppose so :)

preview/DOC-9267-fts-sizing.yml
simon-dew, line 1: reminder to delete before merging

modules/install/pages/sizing-general.adoc
```diff
@@ -534,6 +534,312 @@ NOTE: The storage engine used in the sizing calculation corresponds to the stora
...
+
+[#search-vcpus]
+==== vCPUS
+
+A heavy QPS workload requires more vCPUs. 
+If your workload requires a high QPS, this is the most important part of your sizing for the Search Service.
+
+For example, if your target QPS is 30,000 and your queries are less complex, divide your total QPS target by 200 to get your required vCPUs: 
+
+[stem]
```
simon-dew, line 576: It takes a couple of seconds for the LaTeX to be rendered when you first load the page. As it's rendering, the stem formulas are displayed with large whitespace margins, which makes the page pretty hard to read until rendering is done. Would asciimath any faster?

simon-dew (approved):
LGTM 👍 

One observation about the rendering of the STEM formulas. This probably falls outside the scope of this PR, and may be a question for @osfameron. Not a blocker.

modules/install/pages/sizing-general.adoc
sarahlwelton: I can change these to AsciiMath but I also can't say I had the same problem - they've already rendered for me by the time the page loads up. I wouldn't be able to test this to confirm or not.

preview/DOC-9267-fts-sizing.yml
sarahlwelton: Was keeping around in case there was any review comments - which there was :)

modules/install/pages/sizing-general.adoc
simon-dew: Fair enough — in that case I would say merge this as is, unless there are any other SME comments. The STEM rendering is a slippery problem and not an issue to be fixed in this PR

PR #4088, opened 2026-02-13
DOC-13717 rest api for app telemetry

ggray-cb:
This PR adds documentation for the Server `/settings/appTelemetry` REST API  endpoint.

Here's a list of the changes in this PR, with links to a preview. You will need the [Docs Team credentials](https://confluence.issues.couchbase.com/wiki/spaces/DOCS/pages/1971224949/Docs+Team+demo+site+passwords) on Confluence to access the preview.

* Added new [Application Telemetry](https://preview.docs-test.couchbase.com/docs-server-DOC-13717_REST_API_for_appTelemetry/server/current/rest-api/application-telemetry.html) page that explains getting and configuring telemetry settings. 
* Added a short section to [Configure Prometheus to Collect Couchbase Metrics](https://preview.docs-test.couchbase.com/docs-server-DOC-13717_REST_API_for_appTelemetry/server/current/manage/monitor/set-up-prometheus-for-monitoring.html#app-telemetry) explaining that you can add application metrics to the Prometheus endpoint by enabling application telemetry. It seemed like the right place to mention that.

**Note:** I've made some assumptions about the `maxScrapeClientsPerNode` and `scrapeIntervalSeconds` settings, so be sure to pay special attention to their descriptions.

### DIFF
--- a/modules/manage/pages/monitor/set-up-prometheus-for-monitoring.adoc
+++ b/modules/manage/pages/monitor/set-up-prometheus-for-monitoring.adoc
@@ -145,7 +145,12 @@ See xref:rest-api:rest-discovery-api.adoc[Prometheus Discovery API].
 
 You can enable application telemetry to have Couchbase Server collect metrics from your applications that use the Couchbase SDKs. 
 When you enable application telemetry, Couchbase Server collects telemetry data from your applications. 
-It the reports the collected data as metrics through the same Prometheus endpoint that it uses to report its own metrics. 
+
+Couchbase Server reports the collected data as metrics through the same Prometheus endpoint that it uses to report its own metrics. 
 Enabling this feature lets you use Prometheus to monitor the health of both your Couchbase Server cluster and your applications that use the Couchbase SDKs.
+The application telemetry metrics can help you troubleshoot client issues such as poor performance or timeouts.
+
+Application telemetry is disabled by default in Couchbase Server 8.0.
+A future version of Couchbase Server may enable it by default.
 
 See xref:rest-api:application-telemetry.adoc[] to learn how to enable application telemetry.
--- a/modules/manage/pages/monitor/set-up-prometheus-for-monitoring.adoc
+++ b/modules/manage/pages/monitor/set-up-prometheus-for-monitoring.adoc
@@ -147,6 +147,8 @@ You can enable application telemetry to have Couchbase Server collect metrics fr
 When you enable application telemetry, Couchbase Server collects telemetry data from your applications. 
 
 Couchbase Server reports the collected data as metrics through the same Prometheus endpoint that it uses to report its own metrics. 
+The application telemetry metrics are aggregated across all clients instead of being reported separately for each client.
+
 Enabling this feature lets you use Prometheus to monitor the health of both your Couchbase Server cluster and your applications that use the Couchbase SDKs.
 The application telemetry metrics can help you troubleshoot client issues such as poor performance or timeouts.
 
--- a/modules/manage/pages/monitor/set-up-prometheus-for-monitoring.adoc
+++ b/modules/manage/pages/monitor/set-up-prometheus-for-monitoring.adoc
@@ -139,3 +139,12 @@ The `http_sd_configs` section contains its own copy of the `basic_auth` and `tls
 
 You can change the list that the discovery API returns by adding query parameters to the URL in the `http_sd_configs` section. 
 See xref:rest-api:rest-discovery-api.adoc[Prometheus Discovery API].
+
+== Include Application Telemetry in Prometheus Metrics
+
+You can enable application telemetry to have Couchbase Server collect metrics from your applications that use the Couchbase SDKs. 
+When you enable application telemetry, Couchbase Server collects telemetry data from your applications. 
+It the reports the collected data as metrics through the same Prometheus endpoint that it uses to report its own metrics. 
+Enabling this feature lets you use Prometheus to monitor the health of both your Couchbase Server cluster and your applications that use the Couchbase SDKs.
+
+See xref:rest-api:application-telemetry.adoc[] to learn how to enable application telemetry.
--- a/modules/manage/pages/monitor/set-up-prometheus-for-monitoring.adoc
+++ b/modules/manage/pages/monitor/set-up-prometheus-for-monitoring.adoc
@@ -84,7 +84,7 @@ curl  --get -u prometheus:password \
 Add a new job to collect metrics from your Couchbase Database. 
 In this job, add a `file_sd_configs` key entry. Set its value to the path to the file containing the list of nodes you created earlier. 
 +
-The following example defines a new job to collect metrics from a Couchbase Server, where the certificate  and node list files are in the `/etc/prometheus` directory.
+The following example defines a new job to collect metrics from a Couchbase Server, where the certificate and node list files are in the `/etc/prometheus` directory.
 +
 [source, yaml]
 ----
@@ -109,7 +109,7 @@ However, it's easier to just configure Prometheus to call the discovery API itse
 == Configure HTTP Service Discovery
 
 You can configure Prometheus to call the Couchbase Server discovery API itself. 
-With this method, has the benefit that Prometheus defaults to periodically calling the endpoint so it automatically learns of any added or removed nodes. 
+This method has the benefit that Prometheus defaults to periodically calling the endpoint so it automatically learns of any added or removed nodes. 
 
 To use the HTTP service discovery method, add a job to your Prometheus hosts's `prometheus.yml` configuration file to collect data from your Couchbase Database. 
 In this job, add an `http_sd_configs` section that tells Prometheus to call your database's discovery API endpoint. 
--- a/modules/manage/pages/monitor/set-up-prometheus-for-monitoring.adoc
+++ b/modules/manage/pages/monitor/set-up-prometheus-for-monitoring.adoc
@@ -140,6 +140,7 @@ The `http_sd_configs` section contains its own copy of the `basic_auth` and `tls
 You can change the list that the discovery API returns by adding query parameters to the URL in the `http_sd_configs` section. 
 See xref:rest-api:rest-discovery-api.adoc[Prometheus Discovery API].
 
+[#app-telemetry]
 == Include Application Telemetry in Prometheus Metrics
 
 You can enable application telemetry to have Couchbase Server collect metrics from your applications that use the Couchbase SDKs. 
--- a/modules/rest-api/pages/application-telemetry.adoc
+++ b/modules/rest-api/pages/application-telemetry.adoc
@@ -98,7 +98,7 @@ curl -sS -u $USER:$PASSWORD \
      -X GET 'http[s]://{host}:{port}/settings/appTelemetry' 
 ----
 
-.Path and curl Parameters
+==== Path and curl Parameters
 :priv-link: get-privs
 include::partial$user-pw-host-port-params.adoc[]
 
@@ -177,12 +177,12 @@ curl -sS -u $USER:$PASSWORD \
   [-d scrapeIntervalSeconds=<integer>]
 ----
 
-.Path and curl Parameters
+==== Path and curl Parameters
 :priv-link: config-privs
 include::partial$user-pw-host-port-params.adoc[]
 
 
-.REST Parameters
+==== REST Parameters
 `enabled` (boolean, optional)::
 Set to `true` to enable application telemetry or `false` to turn it off.
 
--- a/modules/rest-api/pages/application-telemetry.adoc
+++ b/modules/rest-api/pages/application-telemetry.adoc
@@ -45,7 +45,7 @@ GET /settings/appTelemetry
 [source,bash]
 ----
 curl -sS -u $USER:$PASSWORD \
-     -X GET 'http[s]://<hostname>:<port>/settings/appTelemetry' 
+     -X GET 'http[s]://{host}:{port}/settings/appTelemetry' 
 ----
 
 .Path and curl Parameters
@@ -56,7 +56,7 @@ include::partial$user-pw-host-port-params.adoc[]
 [#get-privs]
 === Required Privileges
 
-Your user account must have at least one of the following roles to get the application telemetry status:
+Your user account must have at least 1 of the following roles to get the application telemetry status:
 
 * xref:learn:security/roles.adoc#full-admin[Full Admin]
 * xref:learn:security/roles.adoc#cluster-admin[Cluster Admin]
@@ -107,7 +107,7 @@ Running the previous command returns a JSON object similar to the following:
 
 By sending a POST request to the `/settings/appTelemetry` endpoint, you can:
 
-* Enable or disable application telemetry.
+* Enable or turn off application telemetry.
 * Set the limit on the number of clients a single node can scrape telemetry data from at the same time.
 * Set how often the nodes scrape telemetry data from clients.
 
--- a/modules/rest-api/pages/application-telemetry.adoc
+++ b/modules/rest-api/pages/application-telemetry.adoc
@@ -12,10 +12,10 @@ Having Couchbase Server collect telemetry information from your applications tha
 This telemetry data is useful to diagnose issues such as poor performance or timeouts.
 
 When you enable application telemetry, Couchbase Server advertises to SDK clients that it can collect telemetry data.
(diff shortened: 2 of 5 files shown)


modules/rest-api/pages/application-telemetry.adoc
```diff
@@ -0,0 +1,222 @@
...
+
+`maxScrapeClientsPerNode` (integer, optional)::
+Set the maximum number of clients from which a single node can scrape telemetry data at the same time.
+If the number of client telemetry connections reaches this threshold, the node rejects new telemetry connections until the number of connect clients drops.
+
++
+Valid values are from `1` to `1024`. 
+
++
+The default value is `1024`.
```
ingenthr, line 204: Are there any sentinel values like `0` which indicate unlimited?

How would a user know when the limit is being hit?  

@avsej or @timofey-barmin may know this off the top of their heads.

modules/rest-api/pages/application-telemetry.adoc
```diff
@@ -0,0 +1,222 @@
...
+  "maxScrapeClientsPerNode": 512,
+  "scrapeIntervalSeconds": 90
+}
+----
+
+== See Also
+
+* xref:manage:monitor/set-up-prometheus-for-monitoring.adoc[]
+
+* See the SDK Telemetry from the Server section of the Collecting Information and Logging page in the documentation for the SDK you use.
```
ingenthr, line 294: Somewhere in here we should mention the SDK-API level required to get AppTelemetry from the SDK.  @RichardSmedley might have a good way to communicate this. 

To get details, most users will need to update both their SDK, and the cluster, and enable it.  As currently written, the SDK side requirement is possibly missed. 

Another requirement which we may want to mention is that this feature requires the management port to be exposed from cluster to client (typically port 8091/18091, but can be different with alt-address).

ingenthr (changes requested):
Nice work on this @ggray-cb .  We're getting a number of questions about this functionality, so it'll be great to see it in the docs.  

I had a few small comments and @shivaniguptacb may want to review.  Also, it looks like copilot caught a few important things (like the kotlin/python ambiguity).

shivaniguptasf:
Nice work! 
One small thing - I didn't see a mention that in 8.0 it is disabled by default. Might want to add that both to the Server Monitoring page, and to the REST API reference page that you added.

modules/rest-api/pages/application-telemetry.adoc
ggray-cb: According to the [dev PR](https://review.couchbase.org/c/ns_server/+/221739) the lower limit is 1.

modules/rest-api/pages/application-telemetry.adoc
```diff
@@ -0,0 +1,222 @@
...
+----
+GET /settings/appTelemetry
+----
+
+=== curl Syntax
+
+[source,bash]
+----
+curl -sS -u $USER:$PASSWORD \
+     -X GET 'http[s]://{host}:{port}/settings/appTelemetry'
```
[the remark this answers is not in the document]
ggray-cb: For what it's worth, we're making {host} and {port} the standard for these placeholdsers going forward.

modules/rest-api/pages/application-telemetry.adoc
ggray-cb: Also, attempting to set the value to 0 results in an error sating the range is 1 to 1024.

modules/rest-api/pages/application-telemetry.adoc
ggray-cb: I've added a section on prerequisites. I added the SDK API lvel requirements and linked to the SDK page that lists which versions of the SDK implement that API version. I also added the port availability requirement.

ggray-cb:
@shivaniguptasf Added default value for the `enabled` setting, plus added notes stating that it defaults to off but that may change in the future.

ggray-cb:
Adding @RichardSmedley because he expressed interest in reviewing the current draft. Also added @Peter-Searby in hopes he could review the techical details about what happens if a node hits its limit for connected nodes. @ingenthr asked if there was a way for users to tell this has happened.

modules/rest-api/pages/application-telemetry.adoc
```diff
@@ -8,20 +8,34 @@
...
+
+* A Couchbase Server cluster only supports application telemetry when all of its nodes are running version 8.0 or later. 
 Earlier versions of Couchbase Server do not support application telemetry.
-You cannot have the cluster collect application telemetry if the cluster is running in mixed mode with some nodes running a version earlier than 8.0.
-If you enable application telemetry on a cluster running in mixed mode with pre-8.0 nodes, the cluster does not advertise its ability to collect telemetry to clients.
+Your cluster cannot collect application telemetry if it's running in mixed mode where some nodes are running a pre-8.0 version.
+If you enable application telemetry on a mixed-mode cluster with pre-8.0 nodes, it does not advertise to clients that it can collect telemetry.
+
+* Your applications must use an SDK that implements version 3.8 or later of the SDK API.
+See the table in xref:java-sdk:project-docs:compatibility.adoc#api-version[API Version] to determine which version of the SDK your application uses implements version 3.8 of the SDK API.
```
ingenthr: Do note that this is a link to Java, so the user will have to switch to their platform.  I don't know if you want to turn this into something like "see the table for your SDK.  For example, in java…"

@RichardSmedley may have a better way to approach this.

ingenthr (changes requested):
Nearly there, thanks!  There's one bit of ambiguity I think we should try to address if we can.  See comments.

modules/manage/pages/monitor/set-up-prometheus-for-monitoring.adoc
```diff
@@ -139,3 +139,18 @@ The `http_sd_configs` section contains its own copy of the `basic_auth` and `tls
...
 You can change the list that the discovery API returns by adding query parameters to the URL in the `http_sd_configs` section. 
 See xref:rest-api:rest-discovery-api.adoc[Prometheus Discovery API].
+
+[#app-telemetry]
+== Include Application Telemetry in Prometheus Metrics
+
+You can enable application telemetry to have Couchbase Server collect metrics from your applications that use the Couchbase SDKs. 
+When you enable application telemetry, Couchbase Server collects telemetry data from your applications. 
+
+Couchbase Server reports the collected data as metrics through the same Prometheus endpoint that it uses to report its own metrics.
```
Peter-Searby, line 149: Might be worth mentioning that the metrics are aggregated across clients. Currently it reads in a way that suggests you'd be able to see per-client metrics, which is not the case.

modules/rest-api/pages/application-telemetry.adoc
```diff
@@ -0,0 +1,238 @@
...
+
+NOTE: Application telemetry is off by default in Couchbase Server 8.0.
+Future versions of Couchbase Server may enable it by default.
+
+Couchbase Server reports the telemetry data it collects through the same Prometheus metrics endpoint it uses to publish its own metrics.
+See xref:manage:monitor/set-up-prometheus-for-monitoring.adoc[] to learn how to set up Prometheus to collect metrics from your Couchbase Server cluster.
+
+=== Prerequisites
+
+You Couchbase Server cluster and your clients must meet the following requirements to use application telemetry:
```
Peter-Searby:
```suggestion
Your Couchbase Server cluster and your clients must meet the following requirements to use application telemetry:
```

modules/rest-api/pages/application-telemetry.adoc
```diff
@@ -0,0 +1,238 @@
...
+
+* A Couchbase Server cluster only supports application telemetry when all of its nodes are running version 8.0 or later. 
+Earlier versions of Couchbase Server do not support application telemetry.
+Your cluster cannot collect application telemetry if it's running in mixed mode where some nodes are running a pre-8.0 version.
+If you enable application telemetry on a mixed-mode cluster with pre-8.0 nodes, it does not advertise to clients that it can collect telemetry.
```
Peter-Searby: What is this referring to? You can't configure app telemetry on a mixed-mode cluster

modules/rest-api/pages/application-telemetry.adoc
```diff
@@ -0,0 +1,238 @@
...
+This API endpoint supports the following methods:
+
+* <<#get-status>>
+* <<#configure-telemetry>>
+
+
+[#get-status]
+== Get Application Telemetry Status
+
+The following method gets the current state of application telemetry for the cluster.
```
Peter-Searby: Do we normally refer to settings as "state"? Feels like it suggests it would be a status, rather than the configuration

modules/rest-api/pages/application-telemetry.adoc
```diff
@@ -0,0 +1,238 @@
...
+Sets the maximum number of clients a single node can scrape telemetry data from at the same time.
+If the number of client telemetry connections reaches this threshold, the node rejects new telemetry connections until the number of connected clients drops.
+
++
+Valid values are from `1` to `1024`. 
+
++
+The default value is `1024`.
+You can set `maxScrapeClientsPerNode` to a lower value to reduce potential overhead on your nodes from collecting telemetry data.
+However, lowering it too far could result in clients being unable to find a node for their telemetry connection.
```
Peter-Searby: > lower value to reduce potential overhead

> lowering it too far could result in clients being unable to find a node

Feels like basically saying "lower value reduces number of clients, but lowering it too far could result in reducing the number of clients".  Like, the whole point is to reduce the number of clients, so the two statements don't really feel consistent with each other.

```suggestion
You can set `maxScrapeClientsPerNode` to a lower value to reduce potential overhead on your nodes from collecting telemetry data by preventing too many clients connecting to each node.
However, if the limit is reached on all nodes, this will result in any additional clients being unable to find a node for their telemetry connection. 
```

modules/rest-api/pages/application-telemetry.adoc
```diff
@@ -0,0 +1,238 @@
...
+`scrapeIntervalSeconds` (integer, optional)::
+Sets how often the nodes scrape telemetry data from clients in seconds.
+
++
+Valid values are `60` to `600`.
+
++
+The default value is `60`.
+You can increase this value to reduce the overhead of collecting telemetry data on your nodes.
+However, setting it too high could result in a loss of telemetry data when clients close their connections.
```
Peter-Searby: What does this mean?

Peter-Searby (changes requested):

Peter-Searby:
> Also added @Peter-Searby in hopes he could review the techical details about what happens if a node hits its limit for connected nodes. @ingenthr asked if there was a way for users to tell this has happened.

That's basically up to the SDKs to handle. When they attempt to connect to a node that is full, they'll try other nodes, but I can't remember the exact details. Users would presumably see this from the SDK logging, but perhaps @DemetrisChr can confirm?

The admins for the cluster would also be able to detect this happening by following the `cm_app_telemetry_curr_connections` metric. Once this reaches the max scrape clients configured for that node, further connections are rejected.

modules/rest-api/pages/application-telemetry.adoc
RichardSmedley: It's not easy, but I think that we do need to avoid telling customers about the SDK API version, which will only cause confusion.
Really, the only way to do this is linking every SDK, much like the rendered-on-a-single-line links at https://docs.couchbase.com/server/current/guides/creating-data.html#related-links
(in this case going straight to https://docs.couchbase.com/java-sdk/3.9/howtos/collecting-information-and-logging.html#sdk-telemetry-from-the-server etc., with an "and all later versions" at the end of the list, perhaps.)

Probably not the _only_ solution, so happy to discuss further.

RichardSmedley: tangential point: we want people using the latest version of the SDK,
so could just link to `current`, and say it's in "recent" versions.

modules/rest-api/pages/application-telemetry.adoc
```diff
@@ -0,0 +1,238 @@
...
+----
+
+== See Also
+
+* xref:manage:monitor/set-up-prometheus-for-monitoring.adoc[]
+
+* See the SDK Telemetry from the Server section of the Collecting Information and Logging page in the documentation for the SDK you use.
+For example:
+
+** xref:cxx-sdk:howtos:collecting-information-and-logging.adoc#sdk-telemetry-from-the-server[C++ SDK]
```
RichardSmedley: See line 34 comments (and note the use of a single line for 12 SDK links in the linked docs page).

modules/rest-api/pages/application-telemetry.adoc
ggray-cb: OK, so the REST API returns an error if you attempt to enable it in mixed mode?

Peter-Searby: That's correct. On the 8.0 node(s) it returns error 400 "This http API endpoint isn't supported in mixed version clusters". On the pre-8.0 node(s) it returns error 404

modules/rest-api/pages/application-telemetry.adoc
ggray-cb: I assume that if you set `scrapeIntervalSeconds` to a larger value (say the maximum of 10 minutes), you  increase the chances that the client will disconnect before the node can scrape it. I'm not sure if that makes a difference or not. Could there be cases where a client gives up and disconnects if it gets repeated timeouts? If so, a longer scrape interval increases the chances those timeouts would not be scraped from the client before it disconnects?

modules/rest-api/pages/application-telemetry.adoc
ggray-cb: I prefer telling the reader exactly what the minimum version of the SDKs that they need  vs. simply saying "use the latest," I've extracted the SDK version numbers from the table I linked and listed them out with links to the SDKs.

modules/rest-api/pages/application-telemetry.adoc
ggray-cb: Here I'm just linking to the SDKs where I have found the Collecting Information and Logging topic which discusses application telemetry. It looks like nt all SDKs offer this topic.

modules/rest-api/pages/application-telemetry.adoc
ggray-cb: I'v reworded a bit differently and also added in a mention of the metric admins can use to monitor the number of connections.

modules/rest-api/pages/application-telemetry.adoc
Peter-Searby: Ah, I see what you mean. Short-lived connections will be less likely to get scraped, certainly.

I can't speak to the behaviour of SDKs when they get timeouts, but I would guess that it depends on the application configuration.

Short-lived connections will generally not get well represented in the telemetry data, and if the interval is increased then they will be even less well represented.

modules/rest-api/pages/application-telemetry.adoc
Peter-Searby: I think you misread my comment here. I meant that "state" sounds like a status, not that that is a better way of referring to it. Settings are definitely not a "status" or a "state" in my mind

modules/rest-api/pages/application-telemetry.adoc
ggray-cb: I've cleaned up the explanation a bit and also gave an example of why Server would lose more telemetry when the value is set high.

modules/rest-api/pages/application-telemetry.adoc
ggray-cb: Thanks. I've added those responses and stated you cannot enable it in mixed mode with pre-8.0 nodes.

modules/rest-api/pages/application-telemetry.adoc
```diff
@@ -0,0 +1,300 @@
...
+This API endpoint supports the following methods:
+
+* <<#get-status>>
+* <<#configure-telemetry>>
+
+
+[#get-status]
+== Get Application Telemetry Status
+
+The following method gets the current status of application telemetry for the cluster.
```
Peter-Searby: > I think you misread my comment here. I meant that "state" sounds like a status, not that that is a better way of referring to it. Settings are definitely not a "status" or a "state" in my mind

Reposting comment because I couldn't un-resolve the original comment, so this might have got missed

ggray-cb: OK, rewored to "The following method gets the current application telemetry settings the cluster."

Peter-Searby (approved):

PR #4095, opened 2026-03-16
Update to metrics for Server 7.6 and change to Handlebars template

ggray-cb:
* Updated Metrics for 7.6.10 from build #8026.
* Updated the metrics Handlebars template to add an anchor to each entry. The anchor named the same as the entry, including underscores. This covers [DOC-12159](https://jira.issues.couchbase.com/browse/DOC-12159). Also solves a pain point that will let the server docs link to individual metrics.

[DOC-12159]: https://couchbasecloud.atlassian.net/browse/DOC-12159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/metrics-reference/attachments/cbas_metrics_metadata.json
+++ b/modules/metrics-reference/attachments/cbas_metrics_metadata.json
@@ -2,7 +2,20 @@
   "cbas_active_links": {
     "added": "7.2.4",
     "help": "Number of active links.",
-    "type": "gauge"
+    "type": "gauge",
+    "unit": "number"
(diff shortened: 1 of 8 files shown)


osfameron (approved):
So that's just adding a `[[key]]` bookmark.
That's not used in the template, but I assume you want to refer to these from other pages? Sounds sensible 👍

PR #4105, opened 2026-04-15
DOC-13858 Server File-Based Rebalance for Data Service

rao-shwe:
This is my reformatting/editing of Schewtha's draft of the FBR docs.

~~**Important Note:** This draft does not contain documentation for the Web Console's FBR settings. The UI wasn't ready when this draft was written and revised. This draft is mainly for review before inclusion before the early Totoro release. Other features are taking prioority over getting the GUI documentation done.~~ The draft contains the UI docs now for the new File-based Rebalance field.

Cribbing CoPilot's summary of these changes:

Adds documentation for the new Data Service File-Based Rebalance (FBR) feature in Couchbase Server Totoro, including a new REST API reference page, navigation entry, conceptual coverage in the rebalance learn page, a new bucket-level dataServiceRebalanceType parameter, and updates to general settings, node management, and the 8.1 new-features page. Also performs cosmetic cleanup (removing italic emphasis) in the rebalance learn page and updates the shared user/pwd/host/port REST parameter partial.

# Changes (with links to the preview):

You will need the [Docs Team credentials](https://confluence.issues.couchbase.com/wiki/spaces/DOCS/pages/1971224949/Docs+Team+demo+site+passwords) on Confluence to view the preview.

* [New REST reference page file-based-data-rebalance.adoc](https://preview.docs-test.couchbase.com/docs-server-DOC-13858-fbr-for-data-service/server/current/rest-api/file-based-data-rebalance.html) documenting GET/POST /internalSettings usage for FBR, plus nav and rebalance-table entries.
* New dataServiceRebalanceType bucket parameter documentation added to [rest-bucket-create.adoc](https://preview.docs-test.couchbase.com/docs-server-DOC-13858-fbr-for-data-service/server/current/rest-api/rest-bucket-create.html), with cross-references from general settings, add-node-and-rebalance, and the rebalance learn page.
* [Conceptual FBR section added to rebalance.adoc](https://preview.docs-test.couchbase.com/docs-server-DOC-13858-fbr-for-data-service/server/current/learn/clusters-and-availability/rebalance.html#file-based-rebalance) and a [Totoro new-features entry](https://preview.docs-test.couchbase.com/docs-server-DOC-13858-fbr-for-data-service/server/current/introduction/whats-new.html#data-service-changes), plus formatting cleanup of italic markers across rebalance.adoc and rest-rebalance-overview.adoc.


modules/ROOT/nav.adoc
```diff
@@ -358,6 +358,7 @@ include::cli:partial$cbcli/nav.adoc[]
     **** xref:rest-api:rest-cluster-removenode.adoc[Removing Nodes from Clusters]
 
    *** xref:rest-api:rest-rebalance-overview.adoc[Rebalance]
+    **** xref:rest-api:file-based-data-rebalance.adoc[]
```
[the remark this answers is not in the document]
ggray-cb: CoPilot diesn't seem to understand the reference in the rest-rebalance-table.adocis a link to a heading within the page, not the page title itseld.

modules/introduction/partials/new-features-81.adoc
```diff
@@ -13,24 +13,48 @@ TBD
...
+
+The following changes apply to the Data Service rebalance behavior in {totoro}:
+
+* *File-Based Rebalance*: FBR transfers vBucket data files directly from the source node to the destination node during the backfill phase of a vBucket move, bypassing the full DCP backfill mechanism used in prior releases.
+
+* *Enabled by default*: FBR is enabled by default for Enterprise Edition, both for self-managed deployments and Couchbase Capella.
+No configuration is required to activate it.
+
+* *Automatic rebalance type selection*: The server automatically determines whether FBR or DCP is more efficient for each vBucket move.
+When FBR is not applicable or not expected to be faster, the server falls back to DCP automatically.
```
hyunjuV: I do not believe that this is true the way it's written (where the server automatically determines which method is more efficient **for each** vBucket move).

When FBR is enabled (which it is, by default) the backfill is done using the same method for all vBuckets.  Generally, it's always done using file-based rebalance unless there is storage migration or eviction policy change pending.  So, the "Automatic rebalance type selection" does apply when there are scenarios where DCP rebalance is required for all phases -- like when storage migration or eviction policy change is pending effect.  If DCP rebalance is required for all phases, then, that's what the Server will do (even with FBR enabled).

cc @BenHuddleston

modules/introduction/partials/new-features-81.adoc
```diff
@@ -13,24 +13,48 @@ TBD
...
+
+* *New bucket-level rebalance type setting*: A new per-bucket setting, `dataServiceRebalanceType`, allows operators to control rebalance behavior at the bucket level, overriding the cluster-level FBR setting.
+
+* *Separate vBucket move concurrency for FBR*: A new setting, `dataServiceFileBasedRebalanceMovesPerNode`, controls the maximum number of concurrent file-based vBucket moves per node.
+This is independent of the existing `rebalanceMovesPerNode` setting, which applies to DCP rebalance.
+
```
hyunjuV, line 36: There is also a new File-Based Rebalance throttle rate setting (`snapshot_download_throttle_bytes`) which sets the max rate at which file-based rebalance snapshots will be transferred between nodes.  Very high transfer rates means that rebalance will proceed very quickly but may have a negative impact on KV operation latencies during rebalance.  A value of 0 means that snapshot transfer is unthrottled.

Additional details:
1.   snapshot_download_throttle_bytes is an option in GET, POST /pools/default/settings/memcached/global

2. In the UI (under Advanced Rebalance Settings for the Data Service), the setting of 150 MiB/s translates to snapshot_download_throttle_bytes=157286400.  By default, the value is 0 (unthrottled).

3. The rate of FBR snapshots transfer can be seen with a rate function applied to the metric `kv_ep_snapshot_read_bytes`.

modules/learn/pages/clusters-and-availability/rebalance.adoc
```diff
@@ -65,17 +67,17 @@ The phases through which rebalance moves a replica vBucket are shown by the foll
 
 image::clusters-and-availability/replicaVbucketMove.png[,640,align=left]
 
-The move has two principal phases. Phase 1 is _Backfill_. Phase 2 is _Book-keeping_.
+The move has two principal phases. Phase 1 is Backfill. Phase 2 is Book-keeping.
```
hyunjuV: The picture and the description is slightly different for File based vs DCP rebalance. 
The existing pictures are for DCP rebalance.

FYI -- this Google [doc](https://docs.google.com/document/d/1Fxahxvvn1KPP9XzSpkKHA1sbVTHtXT53HiYXTKB2rw4/edit?tab=t.0) has info on how the picture would look different for file-based rebalance.

modules/learn/pages/clusters-and-availability/rebalance.adoc
```diff
@@ -85,23 +87,23 @@ The phases in which rebalance moves an active vBucket are shown by the following
 image::clusters-and-availability/activeVbucketMove.png[,780,align=left]
 
 The move has four principal phases.
-Phase 1, _Backfill_, and Phase 2, _Book-keeping_, are identical to those required for replica vBuckets; except that the _Book-keeping_ phase includes additional _Persistence Time_.
+Phase 1, Backfill, and Phase 2, Book-keeping, are identical to those required for replica vBuckets; except that the Book-keeping phase includes additional Persistence Time.
```
hyunjuV, line 95: Same in this section as in the "Rebalance Phases for Replica vBuckets".  
The picture and the description is slightly different for File based vs DCP rebalance. 
The existing pictures are for DCP rebalance.

FYI -- this Google [doc](https://docs.google.com/document/d/1Fxahxvvn1KPP9XzSpkKHA1sbVTHtXT53HiYXTKB2rw4/edit?tab=t.0) has info on how the picture would look different for file-based rebalance.

modules/learn/pages/clusters-and-availability/rebalance.adoc
```diff
@@ -85,23 +87,23 @@ The phases in which rebalance moves an active vBucket are shown by the following
...
 
-The total time for the move is termed _Move Time_.
+The total time for the move is termed Move Time.
 
 [#limiting-concurrent-vbucket-moves]
 === Limiting Concurrent vBucket Moves
 
-Since vBucket moves are highly resource-intensive, Couchbase Server allows the concurrency of such moves to be _limited_: a setting is provided that determines the maximum number of concurrent vBucket moves permitted on any node.
+Since vBucket moves are highly resource-intensive, Couchbase Server allows the concurrency of such moves to be limited: a setting is provided that determines the maximum number of concurrent vBucket moves permitted on any node.
 The minimum value for the setting is `1`, the maximum `64`, the default `4`.
```
hyunjuV: For DCP rebalance, the minimum value for the setting is `1`, the maximum `64`, the default `4`.
For FBR, the minimum value for the setting is `1`, the maximum `1024`, the default `4`.  (I see that this info is presented in the file-based rebalance section.)

modules/learn/pages/clusters-and-availability/rebalance.adoc
```diff
@@ -113,10 +115,105 @@ Conversely, a lower setting may degrade rebalance performance, while freeing up
...
+FBR applies only to the Backfill phase of a vBucket move, as described in xref:learn:clusters-and-availability/rebalance.adoc#data-service-rebalance-phases[Data Service Rebalance Phases].
+The Active Takeover phase always uses DCP, regardless of whether FBR was used for backfill.
+Because takeover always uses DCP, the DCP rebalance infrastructure remains fully operational in {totoro}.
+
+[#fbr-automatic-selection]
+==== Automatic Rebalance Method Selection
+
+When you enable FBR, the server automatically selects the most efficient method for each vBucket move.
+It uses FBR when it estimates the FBR would be at least 10% faster than DCP.
+The server falls back to DCP automatically in the following situations:
```
hyunjuV: This is for lines 178-180.
It may not work out that the Server automatically selects "the most efficient method for each vBucket move".  So, probably best to replace lines 178-180 with a single line (something like below):

When you enable FBR, the server still falls back to DCP automatically in the following situations:

modules/learn/pages/clusters-and-availability/rebalance.adoc
```diff
@@ -113,10 +115,105 @@ Conversely, a lower setting may degrade rebalance performance, while freeing up
...
+Migrating the storage format requires a full data reload, which is only possible through DCP.
+
+Eviction policy changes::
+Changing a bucket's eviction policy requires data to be reprocessed during rebalance, which requires DCP.
+
+Ephemeral buckets::
+Ephemeral buckets store data entirely in memory and have no persistent storage files for FBR to copy.
+
+Scenarios where DCP may be faster::
+For example, DCP can be faster when the data resident ratio is 100%.
```
hyunjuV: I'm not sure that the Server automatically does DCP rebalance (for the backfill) in this scenario -- should check with @BenHuddleston

modules/learn/pages/clusters-and-availability/rebalance.adoc
```diff
@@ -113,10 +115,105 @@ Conversely, a lower setting may degrade rebalance performance, while freeing up
...
+
+[#fbr-performance]
+==== Performance
+
+The primary goal of FBR is to deliver significant improvements to rebalance speed for large datasets.
+The target throughput is 1 TB of data movement in 30 minutes.
```
hyunjuV: I don't think that we should be specific since the actual rebalance time depends on too many variables.  Please remove line 198.

modules/learn/pages/clusters-and-availability/rebalance.adoc
```diff
@@ -113,10 +115,105 @@ Conversely, a lower setting may degrade rebalance performance, while freeing up
...
+
+[#fbr-concurrent-moves]
+==== Limiting Concurrent File-Based vBucket Moves
+
+In Couchbase Server {totoro}, FBR uses a separate concurrent-moves setting, `dataServiceFileBasedRebalanceMovesPerNode`, which is independent of the DCP `rebalanceMovesPerNode` setting.
+The default value for both settings is `4`.
+Changing one setting does not affect the other.
+
+The setting may be established by means of the xref:manage:manage-settings/general-settings.adoc#rebalance-settings[Couchbase Web Console] or the xref:manage:manage-settings/general-settings.adoc#rebalance-settings-via-rest[REST API].
+
```
hyunjuV, line 259: Need to add info about FBR throttling option -- see info below.

=== File-Based Rebalance Throttle Rate Setting

Since file-based rebalance can increase network usage, there's a way to throttle the file transfer rate, if needed.  By default, there is no throttling.
The File-Based Rebalance throttle rate setting is called snapshot_download_throttle_bytes, and it sets the max rate at which file-based rebalance snapshots will be transferred between nodes. Very high transfer rates means that rebalance will proceed very quickly but may have a negative impact on KV operation latencies during rebalance. A value of 0 means that snapshot transfer is unthrottled.

Additional details:

snapshot_download_throttle_bytes is an option in GET, POST /pools/default/settings/memcached/global

In the UI (under Advanced Rebalance Settings for the Data Service), the setting of 150 MiB/s translates to snapshot_download_throttle_bytes=157286400. By default, the value is 0 (unthrottled).

The rate of FBR snapshots transfer can be seen with a rate function applied to the metric kv_ep_snapshot_read_bytes.

modules/manage/pages/manage-nodes/add-node-and-rebalance.adoc
```diff
@@ -186,6 +186,10 @@ Note that the figure in the *Items* column for node `10.142.181.101` is `31.5 K/
 The figure for `10.142.181.102` indicates the converse.
 Therefore, replication has successfully distributed the contents of `travel-sample` across both nodes, providing a single replica vBucket for each active vBucket.
 
+NOTE: By default, Couchbase Server Enterprise Edition automatically uses File-Based Rebalance (FBR) to move data for eligible vBuckets during node addition.
+The server selects the optimal rebalance method for each vBucket move transparently.
```
hyunjuV: Line 190 should be removed -- the rebalance method is not selected for **each** vBucket move.

modules/rest-api/pages/rest-bucket-create.adoc
```diff
@@ -600,6 +602,60 @@ curl -v -X POST http://localhost:8091/pools/default/buckets/testBucket /
...
+[#dataservicerebalancetype]
+=== dataServiceRebalanceType
+[.edition]#{enterprise}#
+
+Controls the rebalance method used for this bucket's vBucket moves during Data Service rebalance.
+This setting overrides the cluster-level `dataServiceFileBasedRebalanceEnabled` setting for the bucket you set it on.
+
+The valid values are:
+
+* `auto` (default): The server automatically selects File-Based Rebalance (FBR) or DCP for each vBucket move based on which is estimated to be at least 10% faster.
```
hyunjuV: As far as I'm aware:
- Not true that file-based rebalance or DCP is chosen for **each** vBucket move.
- Also not true that file-based or DCP decision is made based on estimates of which is likely to be faster.

Should just say:
The server automatically selects File-Based Rebalance (FBR) or DCP.


cc @BenHuddleston

hyunjuV (changes requested):
1.  The info about File-Based Rebalance throttle rate setting snapshot_download_throttle_bytes should be added, along with metric kv_ep_snapshot_read_bytes that can be monitored to view the throughput.

2.  The rest of the comments were mostly about where the phrasing/text suggests that rebalance type decision is made for **each** vBucket (not true) and that rebalance type decisions are made for scenarios like when either file-based or DCP rebalance is likely to perform x percent faster than the other or if all data is 100% memory resident, etc (not true as far as I'm aware, but should double check with Ben Huddleston).

While true that for ephemeral buckets, there's no backfill phase (so, no file-based rebalance), for non-ephemeral buckets, memory residency is not considered in the decision to do file-based rebalance or DCP rebalance (I do not believe).

modules/introduction/partials/new-features-81.adoc
BenHuddleston: The server does not know whether or not FBR or DCP is more efficient for any given vBucket move. The server will use FBR if possible for any given vBucket, otherwise DCP is used. FBR can only be used for new vBucket builds (i.e. it was not on the node on which it is being build before), only if configured to do so (FBR settings are enabled), and only if no storage/eviction policy migration is currently being done.

Some of this confusion probably comes from calling this "file based **rebalance**", as the actual feature is file based backfill (one particular rebalance step).

BenHuddleston: Suggest:

> * *Automatic rebalance type selection*: The server automatically determines whether FBR can be used for a given vBucket move.
When FBR is not applicable, the server falls back to DCP automatically.

modules/learn/pages/clusters-and-availability/rebalance.adoc
BenHuddleston: Both settings can be set as high as 1024.

modules/learn/pages/clusters-and-availability/rebalance.adoc
BenHuddleston: We do not, FBR is used whenever it is possible to do so. It's quite hard to determine which type of backfill would be faster as it likely depends on disk characteristics, storage backend, fragmentation, and item sizes.

modules/manage/pages/manage-nodes/add-node-and-rebalance.adoc
BenHuddleston: See previous comment, it is selected on a per-vBucket basis.

modules/rest-api/pages/rest-bucket-create.adoc
BenHuddleston: See previous, it is selected on a per-vBucket basis and we do not consider the efficiency of the move, only the eligibility.

modules/introduction/partials/new-features-81.adoc
hyunjuV: Thanks for the clarification @BenHuddleston .

The below summary is to correct any mistaken comments I may have made in my original review comments:

- The server does automatically determine whether FBR can be used for a given vBucket move (or, as in the documentation phrasing, **for each** vBucket move).
- However, this determination is **not** made based on perf considerations (like whether file copy or DCP would be faster).
- Also, this determination is **not** made based on memory residency (like whether data is 100% memory resident).
- But, for ephemeral buckets, DCP is always used (i.e. file copy is not used since there's no persistent storage).

So, as noted by Ben in his suggestion above, best to just say that the server automatically determines whether file-based or DCP rebalance is applicable and not go into specifics.

modules/manage/pages/manage-nodes/add-node-and-rebalance.adoc
hyunjuV: Per Ben's comment, line 190 is OK.

modules/rest-api/pages/rest-bucket-create.adoc
hyunjuV: Per Ben's comment:

The server **does** automatically select File-Based rebalance (FBR) or DCP **for each vBucket move**, but it's not based on which is estimated to be faster (not based on any performance estimates).  So, best to just say that server automatically selects without going into details on how.

modules/manage/pages/manage-nodes/add-node-and-rebalance.adoc
BenHuddleston: IMO "optimal" implies "the most performant" which we do not consider at all (see other comments). I would remove line 190 or at least the word optimal.

modules/rest-api/pages/rest-bucket-create.adoc
ggray-cb: I'm confused about this setting. 

It sounds like from earlier discussions that `auto` is the standard setting. And that setting has rebalance use FBR when it can move the vBucket by copying its files. If it cannot do that, then it falls back to DCP. 

However, having `preferFileBased` and `preferDcp` implies there is more going on here. How does `preferFileBased` differ from `auto`? 

Or does `auto` actually just mean use the cluster default, and the  two `prefer` settings are overrides for that default? I.e. if you set `dataServiceFileBasedRebalanceEnabled` to false, you can use `preferFileBased` to enable FBR for the specific bucket?

BenHuddleston: `auto` and `preferFileBased` are the same. I added `preferFileBased` as it's explicit and makes it easier for us to change the meaning of `auto` in the future should we ever need to.

ggray-cb: Thanks. I've changed the description of preferFileBased is now that it's a synonym for auto.

modules/manage/pages/manage-nodes/add-node-and-rebalance.adoc
ggray-cb: Removed.

modules/learn/pages/clusters-and-availability/rebalance.adoc
ggray-cb: Actually, `rebalanceMovesPerNode` is limited to 64:

```
curl -X POST http://node1.example.com:8091/settings/rebalance -u $USERNAME:$PASSWORD -d rebalanceMovesPerNode=1024
{"errors":{"rebalanceMovesPerNode":"The value must be in range from 1 to 64 (inclusive)"}}
```

BenHuddleston: That's interesting. I had thought that we only set this via the `internalSettings` endpoint. It seems that we have two different endpoints that allow you to modify the setting, with different validation. The `internalSettings` version - https://review.couchbase.org/c/ns_server/+/97684 -  predates `settings/rebalance` which was only added for https://jira.issues.couchbase.com/browse/MB-37060. I only added support for the new setting to `internalSettings`. 

I'll do some investigating and work out what we want to do here.

ggray-cb:
OK, I think I've addressed all comments and missing pieces from the original draft. I've updated the preview site with the latest draft.

modules/learn/pages/clusters-and-availability/rebalance.adoc
BenHuddleston: I've filed https://jira.issues.couchbase.com/browse/MB-72985 and https://jira.issues.couchbase.com/browse/MB-72986 to correct the server behaviour here. We will only allow setting both of these values via `settings/rebalance` with a range of 1-64.

hyunjuV (approved):

modules/learn/pages/clusters-and-availability/rebalance.adoc
```diff
@@ -84,39 +91,176 @@ The phases in which rebalance moves an active vBucket are shown by the following
 
 image::clusters-and-availability/activeVbucketMove.png[,780,align=left]
 
-The move has four principal phases.
-Phase 1, _Backfill_, and Phase 2, _Book-keeping_, are identical to those required for replica vBuckets; except that the _Book-keeping_ phase includes additional _Persistence Time_.
+The move has 4 principal phases.
+Phase 1, Backfill, and Phase 2, Book-keeping, are identical to those required for replica vBuckets; except that the Book-keeping phase includes additional Persistence Time.
```
supritha-kumar, line 95: suggest removing the semi colon.
for reference:

Phase 1, Backfill, and Phase 2, Book-keeping, are identical to those required for replica vBuckets. The exception is that the Book-keeping phase includes additional Persistence Time.

supritha-kumar: Also, consider making this a bullet list.

modules/learn/pages/clusters-and-availability/rebalance.adoc
```diff
@@ -84,39 +91,176 @@ The phases in which rebalance moves an active vBucket are shown by the following
...
+[#fbr-throttle]
+==== FBR Backfill Network Throttle
+
+FBR copies vBucket snapshots via the network during the rebalance backfill phase.
+Directly copying vBucket files is efficient, but it can consume more network bandwidth than DCP backfills.
+Depending on available network bandwidth and the size of the vBuckets, these transfers can add latency to other database activities such as key-value retrievals.
+
+If you see additional latency during rebalances, you can monitor the amount of bandwidth FBR file transfers are consuming using the following metrics:
+
+* xref:metrics-reference:data-service-metrics.adoc#kv_ep_snapshot_read_bytes[`kv_ep_snapshot_read_bytes`]: the number of bytes a node has read from other node's files for FBR.
```
supritha-kumar, line 246: Shouldn't this be the metrics2 page? 
"xref:metrics-reference:data-service-metrics-2.adoc#kv_ep_snapshot_read_bytes[`kv_ep_snapshot_read_bytes`]"

modules/learn/pages/clusters-and-availability/rebalance.adoc
supritha-kumar, line 332: I know this is legacy content but calling it out since i noticed it. Please consider changing "that is" to "that's" .

modules/learn/pages/clusters-and-availability/rebalance.adoc
```diff
@@ -190,26 +334,26 @@ If needed, you can retry these requests on another Query node that is still in t
...
+If there are no Analytics replicas, and an Analytics node fails over, the Analytics Service stops working cluster-wide: ingestion of shadow data stops and no Analytics operations can be run.
 In this case:
 
 * If the Analytics node is recovered, the Analytics Service is resumed and ingestion of shadow data resumes from the point before the node failed over.
 
 * If the Analytics node is removed, the Analytics Service becomes active again after rebalance, but ingestion of shadow data must begin again from scratch.
 
-If there _are_ Analytics replicas, and an Analytics node fails over, the Analytics Service continues to work: one of the replicas is promoted to serve the shadow data that was stored on the failed over node.
+If there are Analytics replicas, and an Analytics node fails over, the Analytics Service continues to work: 1 of the replicas is promoted to serve the shadow data that was stored on the failed over node.
 The Analytics Service only needs to rebuild any shadow data that isn't already ingested from the Data Service, depending on the state of the promoted replica.
```
supritha-kumar, line 357: I know this is legacy content but calling it out since i noticed it. Please consider changing "isn't" to "is not" .

modules/manage/pages/manage-settings/general-settings.adoc
```diff
@@ -174,24 +174,54 @@ See xref:learn:data/durability.adoc[Durability], for information.
...
+File-Based Rebalance Throttle Rate::
+Limits the rate of data transfer for File-Based Rebalance (FBR) on a per node basis.
+This limit applies to the total of incoming and outgoing data per second.
+The default value of `0` lets the node use unlimited bandwidth when transferring data during rebalance backfills.
+For more information, see xref:learn:clusters-and-availability/rebalance.adoc#fbr-throttle[FBR Backfill Network Throttle].
+
+
+Max moves per node during rebalance::
+Sets the maximum number of concurrent vBucket moves permitted on each node during rebalance backfills.
+The *For DCP-based rebalance* field sets the limit for Data Change Protocol (DCP) backfills.
```
supritha-kumar, line 211: in rebalance.adoc, DCP concurrency is 64 and FBR, 1024. This file indicates 64 for both.
Assuming this needs to be updated. Please take a look.

modules/manage/pages/manage-settings/general-settings.adoc
```diff
@@ -910,45 +945,150 @@ If successful, the command returns the following object:
...
+
 ----
 
-To _set_ a new value for the parameter use the `POST` method with the same URI, and with the `rebalanceMovesPerNode` parameter.
-Note that the minimum value is `1`, and the maximum `64`.
+`rebalanceMovesPerNode` is the concurrency limit for DCP backfills and `dataServiceFileBasedRebalanceMovesPerNode` is the limit for FBR.
+Both are set to the default value of `4`.
+
+To set a new values, send a `POST` request to the same endpoint setting the parameters to a new values.
+The valid range for these settings are from `1` to `64`.
```
supritha-kumar, line 1026: in rebalance.adoc, DCP concurrency is 64 and FBR, 1024. This file indicates 64 for both.
Assuming this needs to be updated. Please take a look.

hyunjuV: @supritha-kumar 
Yes, the values for `dataServiceFileBasedRebalanceMovesPerNode` was changed later.
The valid values are 1-64, default 4.
It's an option in the  REST APIs`GET /settings/rebalance` and `POST /settings/rebalance` (earlier, it had been an option in /internalSettings).

PR #4110, opened 2026-04-22
[DOC-13921]: DOC-13921--extras

RayOffiah:
Implemented text change suggestions.
Fixed problem with MathJax formula on Firefox for Linux. (Replaced with plain text)

Preview:

https://preview.docs-test.couchbase.com/docs-server-DOC-13921--extras/server/current/xdcr-reference/xdcr-lowering-memory-footprint.html


ggray-cb (commented):
Looks good!

PR #4114, opened 2026-05-01
DOC-11561: Analytics JDK requirements

simon-dew:
Docs issue: DOC-11561

Updates additional installation requirements for Analytics Service. Does a driveby fix of a couple of examples in the client certificate documentation to suggest using a supported version of JRE also.

No doc preview, minimal changes.


simon-dew:
Will merge by close of play today

PR #4122, opened 2026-05-27
[DOC-14129]: Magma Compaction rate limiter documentation

RayOffiah:
Added page detailing REST call for Magma limit rate for compaction.

preview
----

Preview URL:
https://preview.docs-test.couchbase.com/docs-server-DOC-14129--8.1--Magma-Compaction-rate-limiter-documentation/server/current/rest-api/rest-magma-compaction-rate-limiting.html

### DIFF
--- a/modules/rest-api/pages/rest-magma-compaction-rate-limiting.adoc
+++ b/modules/rest-api/pages/rest-magma-compaction-rate-limiting.adoc
@@ -77,6 +77,25 @@ Running this command returns the global settings:
 -d magma_compaction_rate_limit=500 | jq
 ----
 
+The call executes and returns a JSON object containing the new setting.
+
+[source, json]
+----
+{
+  "max_connections": 20000,
+  "system_connections": 5000,
+  "magma_compaction_rate_limit": 500,
+  "magma_enable_compaction_dataonly_ratelimiting": false
+}
+----
+
+.Setting Compaction for data only rate limiting
+[source,shell]
+----
+curl -u Administrator:password -X POST localhost:8091/pools/default/settings/memcached/global \
+-d magma_enable_compaction_dataonly_ratelimiting=true | jq
+----
+
 The call executes and returns a string containing the new setting.
 
 [source, json]
@@ -86,24 +105,5 @@ The call executes and returns a string containing the new setting.
   "system_connections": 5000,
   "magma_compaction_rate_limit": 500,
   "magma_enable_compaction_dataonly_ratelimiting": true
-}
-----
-
-.Setting Compaction for data only rate limiting
-[source,shell]
-----
-curl -u Administrator:password -X POST localhost:8091/pools/default/settings/memcached/global \
--d magma_enable_compaction_dataonly_ratelimiting=false | jq
-----
-
-The call executes and returns a string containing the new setting.
-
-[source, json]
-----
-{
-  "max_connections": 20000,
-  "system_connections": 5000,
-  "magma_compaction_rate_limit": 500,
-  "magma_enable_compaction_dataonly_ratelimiting": false
 }
 ----
--- a/modules/rest-api/pages/rest-magma-compaction-rate-limiting.adoc
+++ b/modules/rest-api/pages/rest-magma-compaction-rate-limiting.adoc
@@ -6,8 +6,14 @@
 
 == Description
 
-Use these API calls to limit the disk I/O bandwidth (in bytes per second) that Magma Storage Engine Compaction may consume. 
+Use these API calls to limit the disk I/O bandwidth (in *bytes per second*) that the Magma Storage Engine Compaction may consume. 
 The limit is global to the Data Service and shared across all buckets on the node
+
+NOTE: The limit setting is global, 
+even though the application of the setting is on a per-node basis. 
+This allows compactions on each node to consume a maximum of `compaction-rate-limit` bytes per second.
+
+Compaction is a background maintenance activity that reclaims space from stale and deleted data and keeps data files efficient to read. It operates on the same data files that front-end reads and writes use, so large bursts of compaction can consume enough disk I/O to slow front-end operations. Rate limiting smooths those bursts into a steady background rate, keeping front-end performance predictable — at the cost of slower compaction.
 
 Magma supports 2 types of compaction:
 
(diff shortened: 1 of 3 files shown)


modules/rest-api/pages/rest-magma-compaction-rate-limiting.adoc
```diff
@@ -0,0 +1,109 @@
...
+:stem: asciimath
+[abstract]
+{description}
+
+== Description
+
+Use these API calls to set the rate limit (in bytes per second). 
+This setting is a global rate limit for the Memcached process shared across buckets. 
+The strategy for this rate limiting is to set the block waiting time on the compaction threads to maintain the specified limits. 
+
```
ggray-cb, line 11: Is there any guidance we can give for customers as to when they should enable rate limiting? It seems off by default, so I assume they would only enable it if they wre seeing some sort of issue?

ggray-cb (dismissed):
Have one question/suggestion on adding info about when these limits should be enabled. Otherwise, looks good.

modules/rest-api/pages/rest-magma-compaction-rate-limiting.adoc
RayOffiah: That's a good point. I'll check with the devs.

anuthan:
@t3rm1n4l Could you help with when the users might want to use this setting, and when they shouldn't?

t3rm1n4l:
I would recommend removing the magma_enable_compaction_dataonly_ratelimiting setting from documentation. It's confusing to the customer and it's an advanced setting that we are still testing. We can also remove the detailed description about compaction types. 

Some additional info and rephrase suggestions..  

-----
## Description

Use these API calls to limit the disk I/O bandwidth (in **bytes per second**) that Magma Storage Engine  Compaction may consume. The limit is global to the Data Service and shared across all buckets on the node

Compaction is a background maintenance activity that reclaims space from stale and deleted data and keeps data files efficient to read. It operates on the same data files that front-end reads and writes use, so large bursts of compaction can consume enough disk I/O to slow front-end operations. Rate limiting smooths those bursts into a steady background rate, keeping front-end performance predictable — at the cost of slower compaction. 

## When to use it

Rate limiting helps wherever compaction can surge and contend with front-end I/O, for example:

- **Bursty write heavy workloads**, where spikes in writes drive spikes in compaction.
- **Large collection drops**, which mark a large amount of data for removal at once.
- **Sudden large number document expirations**, such as a batch of items with closer TTL range expiring together.

The trade-off is that compaction runs more slowly, so disk usage may stay higher for longer. Set the limit low enough to protect front-end write bursts, but high enough that compaction keeps pace with the write rate. 

## Recommended value

A practical starting point is **100 MB/s** (`104857600` bytes/s), or about **10× your incoming write rate** — estimated as writes per second per node × average document size. For example, 10,000 writes/s of 1 KB documents is roughly 10 MB/s incoming, giving a ~100 MB/s limit.

t3rm1n4l:
@shivaniguptacb Please review

modules/rest-api/pages/rest-magma-compaction-rate-limiting.adoc
```diff
@@ -0,0 +1,102 @@
...
+----
+
+Running this command returns the global settings:
+
+[source, json]
+----
+{
+  "max_connections": 20000,
+  "system_connections": 5000,
+  "magma_enable_compaction_dataonly_ratelimiting": false
```
anirudhRowjee: If we don't have data-only rate limiting exposed, looking for this name instead of `magma_compaction_rate_limit` will confuse the customer - it would help if we used this key in both the set and get example (line 100)

modules/rest-api/pages/rest-magma-compaction-rate-limiting.adoc
```diff
@@ -0,0 +1,102 @@
...
+[source, shell]
+----
+GET /pools/default/settings/memcached/global 
+POST /pools/default/settings/memcached/global 
+
+----
+
+== Examples
+
+.Retrieve the compaction limit settings
```
anirudhRowjee, line 67: I'd prefer if we changed this to "Compaction Rate Limit Settings" (and on line 85)

modules/rest-api/pages/rest-magma-compaction-rate-limiting.adoc
```diff
@@ -0,0 +1,102 @@
...
+== Settings
+
+[cols="1,2,1"]
+|===
+| Setting | Description| Range
+
+| `magma_compaction_rate_limit`
+| Controls the I/O bandwidth
+(given in bytes per second)
+that compactions across all shards consume.
```
anirudhRowjee: Can we say "compactions on all shards **on one node** consume"?

modules/rest-api/pages/rest-magma-compaction-rate-limiting.adoc
```diff
@@ -0,0 +1,102 @@
+= Magma Compaction Rate Limiting
+:description: This API is used to set the rate limit for Magma compaction threads. 
+:stem: asciimath
+[abstract]
+{description}
+
+== Description
+
+Use these API calls to limit the disk I/O bandwidth (in bytes per second) that Magma Storage Engine Compaction may consume. 
+The limit is global to the Data Service and shared across all buckets on the node
```
anirudhRowjee, line 10: Would it be worth adding that "though the setting is global, the application is on a per-node basis, allowing compactions on each node to consume a maximum of (compaction rate limit) bytes per second"

anirudhRowjee (changes requested):

modules/rest-api/pages/rest-magma-compaction-rate-limiting.adoc
shivaniguptasf, line 1: Two general comments:
- I think the description of the two types of compactions (LSM tree compaction and data log compaction) is unnecessary, especially since it is not referred to in any way later the text.
- What is the default setting? This should be clearly mentioned. Majority of users I assume will not tinker with the default setting. Provide strict guidance on when the default should be tinkered with.

RayOffiah:
Ready for another review. 👍🏾

shivaniguptasf (approved):
looks good now. thanks.

PR #4127, opened 2026-06-09
DOC-14243-Cluster-Management

supritha-kumar:
Sub branch for documenting Cluster Management


modules/ROOT/nav.adoc
```diff
@@ -104,6 +104,15 @@ include::third-party:partial$nav.adoc[]
...
  *** xref:learn:security/native-encryption-at-rest-overview.adoc[]
+ **** xref:learn:security/encryption-at-rest-keys.adoc[Encryption Keys]
+ **** xref:learn:security/per-node-per-bucket.adoc[Per-Node, Per-Bucket, and Per-Data-Type Encryption]
+ **** xref:learn:security/transparent-encryption-decryption.adoc[Transparent Encryption and Decryption]
+ **** xref:learn:security/master-password-encryption-at-rest.adoc[Master Password Encryption at Rest]
+ **** xref:learn:/security/encryption-key-management-services.adoc[Encryption Key Management Services]
+ **** xref:learn:security/multiple-KMS-multiple-keys.adoc[Multiple KMSs and Multiple Keys]
+ **** xref:learn:security/encryption-key-rotation-and-expiration.adoc[Encryption Key Rotation and Expiration]
+ **** xref:learn:security/when-CBServer-encrypts-data.adoc[When Couchbase Server Encrypts Data]
+ **** xref:learn:security/key-health-monitoring-alerts.adoc[Key Health Monitoring and Alerts]
```
sarahlwelton, line 115: Do you want all of these to be at the same level as "Encryption Keys" in the nav?

modules/learn/pages/security/encryption-at-rest-keys.adoc
```diff
@@ -0,0 +1,395 @@
+[#keys]
```
sarahlwelton: I don't think you really need to set a direct link on the title of a page? Why is it an H2 heading anyway?

modules/learn/pages/security/encryption-at-rest-keys.adoc
```diff
@@ -0,0 +1,395 @@
...
+
+* *Configuration:* By default, Couchbase Server uses the master password to encrypt configuration data.
+You can choose to use an encryption-at-rest key instead.
+
+* *Logs:* You can choose to use an encryption-at-rest key to encrypt logs.
+You can also use the master password instead.
+
+* *Audits:* As with configuration and log data, you can choose to use an encryption-at-rest key or the master password to encrypt audit data.
+
+* *Other:* You can enable a key to encrypt service-generated data, such as Query Service temporary spill files.
```
sarahlwelton: Can we call this "service-generated data" instead of "other," then?

modules/learn/pages/security/encryption-at-rest-keys.adoc
```diff
@@ -0,0 +1,395 @@
...
+
+[#two-tier-encryption-model]
+== Two-Tier Encryption Model (KEK to DEK)
+
+Couchbase Server uses a two-tier key hierarchy to encrypt data at rest:
+
+=== Key Encryption Keys (KEKs)
+You create and manage KEKs through the Couchbase Server REST API or UI.
+A KEK encrypts and decrypts 1 or more Data Encryption Keys.
+You can store a KEK locally (managed by Couchbase Server) or in an external KMS such as AWS KMS, Azure Key Vault, GCP KMS, HashiCorp Vault, or a KMIP-compatible server.
```
sarahlwelton, line 43: Can we link to more info about each of these key vaults?

modules/learn/pages/security/encryption-at-rest-keys.adoc
```diff
@@ -0,0 +1,395 @@
...
+For example, you can use 1 KEK for bucket data and a different KEK for configuration data.
+In multi-tenant configurations, each tenant's bucket can use its own KEK.
+See <<using-multiple-kmss-and-multiple-keys>> for more information.
+
+IMPORTANT: Use an external KMS key only as a KEK for locally managed keys.
+Do not assign an external KMS key directly to encrypt data.
+Doing so causes Couchbase Server to call the external KMS for every DEK decryption, which increases latency and creates an availability dependency.
+See <<remote-kms-best-practice>> for more information.
+
+== Encrypting Bucket Data
```
sarahlwelton, line 73: Can we link to this from where we talk about bucket encryption further up the page?

modules/learn/pages/security/encryption-at-rest-keys.adoc
```diff
@@ -0,0 +1,395 @@
...
+Encrypting just sensitive data can help reduce the overhead of encrypting and decrypting data on your cluster.
+
+Each bucket can have its own encryption key.
+This configuration is useful in multi-tenancy configurations where each customer may want to manage their own encryption key.
+
+== Encrypting Audit, Logs, and Configuration Data
+
+In addition to data, you can encrypt audit, logs, and most configuration data. 
+You enable the encryption of each of these types of data separately.
+You can use either an encryption-at-rest key or the master password to encrypt data.
```
sarahlwelton, line 88: This feels a little repetitive, especially since we don't link to it from anywhere else.

modules/learn/pages/security/encryption-at-rest-keys.adoc
```diff
@@ -0,0 +1,395 @@
...
+
+=== Configuration Data
+
+Configuration data includes cluster topology, security settings, passwords, certificates, and service configuration files.
+Now, it also includes the Prometheus monitoring configuration and indexing MetaKV entries.
```
sarahlwelton: Now? Is this new in this update?

modules/learn/pages/security/encryption-at-rest-keys.adoc
```diff
@@ -0,0 +1,395 @@
...
+
+By default, Couchbase Server uses the master password to encrypt configuration data.
+Couchbase Server does not set the master password by default.
+You must set it on each node in the cluster.
+See xref:manage:manage-security/manage-system-secrets.adoc#setting-the-master-password[Setting the Master Password] for more information.
```
sarahlwelton: Format for introducing links is usually: 

```suggestion
For more information, see xref:manage:manage-security/manage-system-secrets.adoc#setting-the-master-password[Setting the Master Password].
```

modules/learn/pages/security/encryption-at-rest-keys.adoc
```diff
@@ -0,0 +1,395 @@
...
+You can also assign an encryption-at-rest key to encrypt configuration data instead of the master password.
+
+[NOTE]
+====
+* Some configuration data cannot be encrypted.
+This includes:
+
+** Bootstrap information.
+** Node and internal client certificates.
+** Prometheus configuration, metric data, and tokens used to gather metrics.
```
sarahlwelton, line 113: But we said further up that "now, this includes Prometheus configuration"? Can this be encrypted or no?

modules/learn/pages/security/encryption-at-rest-keys.adoc
```diff
@@ -0,0 +1,395 @@
...
+Audit data includes the audit event files that Couchbase Server generates.
+You can assign an encryption-at-rest key or use the master password to encrypt audit data.
+
+[IMPORTANT]
+====
+When you enable encryption for system data, only new data is encrypted.
+Existing system data (such as log files written before encryption was enabled) remains unencrypted.
+====
+
+== Service-Specific Details
```
sarahlwelton, line 143: I'm honestly wondering if all of this info about what each Service encrypts should be in a separate topic? This is a lot of info and its deviating away from being about encryption keys.

modules/learn/pages/security/encryption-at-rest-keys.adoc
```diff
@@ -0,0 +1,395 @@
...
+
+==== Observability
+
+You can check the encryption status of every index using a REST API endpoint.
+// TODO: :info: confirm endpoint details with the Index team — a REST endpoint shows the encryption status of every bucket, vBucket, and index.
+
+For more information about the Index Service, see xref:learn:services-and-indexes/indexes/global-secondary-indexes.adoc[Global Secondary Indexes].
+
+
+=== Search Service (FTS)
```
sarahlwelton:
```suggestion
=== Search Service
```

modules/learn/pages/security/encryption-at-rest-keys.adoc
```diff
@@ -0,0 +1,395 @@
...
+When you turn off encryption for the bucket, the Search Service stops encrypting new data.
+The Search Service continues to decrypt data that was encrypted before encryption was disabled.
+
+==== Performance Considerations
+
+Encrypting Search index data adds overhead to both write and read operations.
+You may observe increased latency and reduced throughput for Search queries, especially for vector search workloads.
+Test your workload with encryption enabled to understand the performance impact in your environment.
+
+For more information about the Search Service, see xref:fts:fts-introduction.adoc[Full Text Search: Fundamentals].
```
sarahlwelton: Not a good link. 

```suggestion
For more information about the Search Service, see xref:search:search.adoc[].
```

modules/learn/pages/security/encryption-at-rest-keys.adoc
```diff
@@ -0,0 +1,395 @@
...
+Several services generate diagnostic files that contain potentially sensitive information.
+When you enable encryption at rest, these files are encrypted as they are written.
+
+Use the `cbcat` CLI tool to decrypt any encrypted service-generated file.
+Run `cbcat` with the `--with-gosecrets` flag and the path to `gosecrets.cfg`.
+See xref:cli:cbcat.adoc[cbcat] for usage information.
+
+IMPORTANT: Encryption at rest does not retroactively encrypt service-generated files that were written before you enabled encryption.
+Only files written after enablement are encrypted.
+
```
sarahlwelton, line 377: You're missing a "See Also" at the bottom, here.

modules/learn/pages/security/encryption-key-management-services.adoc
```diff
@@ -0,0 +1,563 @@
+[#kms]
+= Encryption Key Management Services
+:description: A key management service (KMS) creates, stores, and manages the lifecycle of encryption keys. Couchbase Server uses a KMS to manage its encryption-at-rest keys.It uses these keys to encrypt and decrypt data encryption keys (DEKs) which are the keys nodes use to encrypt and decrypt the data they store on disk.
```
sarahlwelton:
```suggestion
:description: A key management service (KMS) creates, stores, and manages the lifecycle of encryption keys. Couchbase Server uses a KMS to manage its encryption-at-rest keys. It uses these keys to encrypt and decrypt data encryption keys (DEKs) which are the keys nodes use to encrypt and decrypt the data they store on disk.
```

modules/learn/pages/security/encryption-key-management-services.adoc
```diff
@@ -0,0 +1,563 @@
...
+See xref:manage:manage-security/manage-system-secrets.adoc#setting-the-master-password[Setting the Master Password] for more information about the master password.
+This password is not set by default--you must configure it for each node. 
+See xref:manage:manage-security/manage-system-secrets.adoc[] for more information about the master password.
+
++
+You can also use an encryption key managed by another KMS as a key encryption key (KEK) to encrypt the keys that Couchbase Server manages.
+In this configuration, Couchbase Server relies on the remote KMS to decrypt its encryption keys.
+This is the recommended configuration when using an external KMS.
+It limits the number of encryption and decryption requests Couchbase Server has to send to the external KMS.
+After Couchbase Server has the KMS decrypt its own encryption keys, it can use them to encrypt and decrypt DEKs locally.
```
sarahlwelton: But here, we're talking about Couchbase Server being the KMS - do we really need to repeat this in this section?

modules/learn/pages/security/encryption-key-management-services.adoc
```diff
@@ -0,0 +1,563 @@
...
+
+. Create an IAM user or role with the following minimum permissions on the key:
++
+--
+* `kms:Encrypt`
+* `kms:Decrypt`
+* `kms:DescribeKey`
+--
+
+. Generate an access key ID and secret access key for the IAM user (or configure an instance profile if running on EC2).
```
sarahlwelton, line 70: Okay, but then we don't link to or explain how to set up the AWS KMS anyway?

modules/learn/pages/security/encryption-key-management-services.adoc
```diff
@@ -0,0 +1,563 @@
...
+
+=== Integration with Couchbase Server
+
+When you create an encryption-at-rest key using the AWS KMS provider, you supply the following:
```
sarahlwelton, line 74: Is this what Couchbase Server needs to set up AWS as a KMS? Shouldn't this still be prerequisites?

modules/learn/pages/security/encryption-key-management-services.adoc
```diff
@@ -0,0 +1,563 @@
...
+| `keyResourceName`
+| The full resource name of the Cloud KMS key (for example, `projects/my-project/locations/us-central1/keyRings/my-ring/cryptoKeys/my-key`).
+
+| `credentialsFile`
+| The path to the JSON service account key file on each Couchbase Server node.
+Not required if you use instance-attached service account credentials.
+|===
+
+NOTE: Place the service account key file on every node in the cluster at the same path.
+Couchbase Server reads this file to authenticate with Cloud KMS.
```
sarahlwelton, line 154: But then we don't have a considerations section here, and again, these aren't steps for configuring with this specific service and we don't link to these steps?

modules/learn/pages/security/encryption-key-management-services.adoc
```diff
@@ -0,0 +1,563 @@
...
+
+. Generate a new symmetric key in the KMIP server and record its ID.
+. Update the KMIP key ID in the corresponding encryption key in Couchbase Server.
+. Couchbase Server re-encrypts all keys encrypted by the old KMIP key with the new key.
+Previous KMIP key IDs are saved until re-encryption is complete, then removed automatically.
+
+NOTE: Unlike AWS KMS (which uses transparent rotation), KMIP requires you to manually update the key ID in Couchbase Server after generating a new key in the KMIP server.
+
+[#bring-your-own-key]
+== Bring-Your-Own-Key (BYOK)
```
sarahlwelton, line 491: This is significant enough and, I imagine, enough of a different workflow that it might be worth putting into a separate topic? 

Again, this is a huge page and we could probably make it easier to navigate with tabs.

modules/learn/pages/security/encryption-key-management-services.adoc
```diff
@@ -0,0 +1,563 @@
...
+BYOK protects against offline access to storage media.
+* Use the recommended multi-layer key hierarchy: external key → local KEK → DEK → data.
+See <<remote-kms-best-practice>> for details.
+* Key rotation in the external KMS must be coordinated with Couchbase Server.
+After rotating a key in your KMS, trigger a corresponding rotation in Couchbase Server.
+See <<encryption-key-rotation-and-expiration>> for details.
+
+
+[#aws-kms-caution]
+== Remote KMS Best Practice
```
sarahlwelton, line 535: Again, might be better as a separate page.

sarahlwelton (changes requested):
I'm not going to review all of this right now as I feel I'm addressing a lot of the same issues throughout, and I don't want to overwhelm with comments. 

I can keep going if you like, but this needs a lot of work from a style and a content architecture position. There's some weird formatting decisions being made here, too.

modules/learn/pages/security/encryption-at-rest-keys.adoc
sarahlwelton: To be honest, I'm not even sure where you got that link from because that file doesn't exist.

supritha-kumar:
> I'm not going to review all of this right now as I feel I'm addressing a lot of the same issues throughout, and I don't want to overwhelm with comments.
> 
> I can keep going if you like, but this needs a lot of work from a style and a content architecture position. There's some weird formatting decisions being made here, too.

Hey Sarah, thank you for taking the time to review. A lil unsure why you got around to it despite me clearly calling out that I haven't had the time to do a self-review and changes still need to be made.
I will do the needful before the GA. Thanks!

modules/learn/pages/security/encryption-at-rest-keys.adoc
```diff
@@ -0,0 +1,395 @@
...
+==== Audit Data
+
+Audit data includes the audit event files that Couchbase Server generates.
+You can assign an encryption-at-rest key or use the master password to encrypt audit data.
+
+[IMPORTANT]
+====
+When you enable encryption for system data, only new data is encrypted.
+Existing system data (such as log files written before encryption was enabled) remains unencrypted.
```
timofey-barmin: Do we need to mention that we can encrypt pre-existing data?
It just needs to be done explicitly/separately. See the force encryption-at-rest endpoint.

modules/learn/pages/security/encryption-key-management-services.adoc
```diff
@@ -0,0 +1,574 @@
...
+| <<hashicorp-vault,HashiCorp Vault>>
+| HashiCorp Vault server
+| Totoro
+
+| <<kmip-kms,KMSs that support KMIP>>
+| KMIP-compatible server
+| 8.0
+|===
+
+All external KMS providers follow the same integration model: key material stays in the external KMS and never resides on a Couchbase Server node.
```
timofey-barmin: That's not always correct. For example, in case of KMIP we can (optionally) download the key and encrypt/decrypt locally. IIRC we do that to support older KMIP servers.

See LocalEncrypt vs NativeEncryptDecrypt

@boparai11 Could you please comment?

modules/learn/pages/security/encryption-key-rotation-expiration.adoc
```diff
@@ -0,0 +1,236 @@
...
+| Key rotation (automatic)
+| Limits the amount of data encrypted with any single key.
+Lightweight and safe to run frequently.
+
+| Suspected key compromise
+| Drop DEKs (force re-encryption)
+| Ensures the compromised key is no longer used to protect any data.
+
+| Compliance requirement for periodic re-encryption
+| Drop DEKs (scheduled)
```
timofey-barmin: I think that the recommended operation for "Compliance requirement for periodic re-encryption" should be  "Configured DEK rotation" or something similar. Or we can recommend both those options (depends on the scenario).

If DEK rotation is enabled, all the data will get re-encrypted periodically automatically. It depends on the DEK lifetime configuration.
Also the system will on-average re-encrypt N/M of data with each DEK rotation, which limits the performance impact (where N - DEK rotation interval, M - DEK lifetime interval).
 
For example:
If DEK lifetime is set to 12 month with 1 month DEK rotation, the server will initiate re-encryption of each piece of data in one year or earlier (note: it takes some time to re-encrypt the data; re-encryption time is not included in that 12 month time interval). The server will generate new key every month, and on average only 1/12 of data will be re-encrypted every month automatically. 

Also note that actual re-encryption will only happen in this case if it is strictly necessary (if we actually have data encrypted with the expired key). The data can be re-written (and hence re-encrypted with the latest key) naturally - during compactions or other activities. If that happens, the data will not be re-encrypted during DEK rotation as it is already re-encrypted with a newer DEK.

The "Drop DEKs" can also be used for this purpose if DEK rotation is disabled for some reason. But this operation is heavier as it re-encrypts %100 of the data all the time no matter what.

modules/learn/pages/security/encryption-key-rotation-expiration.adoc
```diff
@@ -0,0 +1,236 @@
...
+| Transitioning from unencrypted to fully encrypted
+| Drop DEKs (one-time)
+| Encrypts all existing data that was written before encryption was enabled.
+|===
+
+TIP: In most environments, automatic DEK rotation is sufficient.
+Reserve the Drop DEKs operation for situations where you need to guarantee that no data remains encrypted with a specific key.
+
+[#drop-deks-force-re-encryption]
+=== Drop DEKs and Force Re-encryption
```
timofey-barmin: It seems like there is a confusion here.
There are two different endpoints to manage DEKs:
1. Drop DEKs (POST /controller/dropEncryptionAtRestDeks/*) - it drops all existing DEKs and re-encrypts all the data if encrypt is enabled.
2. Force encryption (POST /controller/forceEncryptionAtRest/*) - it forces the system to encrypt all the data (usually this is the data that pre-existed when encryption was enabled). We do not generate new keys, and we do not re-encrypt already encrypted data.

When to use Drop DEKs:
- ~~After enabling encryption for the first time~~ - this needs to be moved to "When to use force encryption". It will force encryption for all the data but it will also re-encrypt what's already encrypted. To avoid that "Force Encryption" should be used.
- After a suspected key compromise
- After disabling encryption
- ~~To meet periodic re-encryption requirements~~ - IMO this is a bad suggestion. Data will be re-encrypted automatically if DEK rotation is configured. That should be the main recommended way to "to meet periodic re-encryption requirements". Or we can say that "drop DEKs" can be used to meet periodic re-encryption requirements if DEK rotation is disabled for one reason or another.

When to use Force Encryption:
- After enabling encryption for the first time - it will only encrypt the data that is not yet encrypted

modules/learn/pages/security/encryption-key-rotation-expiration.adoc
```diff
@@ -0,0 +1,236 @@
...
+| 1 day
+| Not enabled by default
+|===
+
+NOTE: The DEK lifetime must be greater than or equal to the rotation interval.
+If you set a lifetime shorter than the rotation interval, the DEK may expire before the next rotation occurs, causing Couchbase Server to force re-encryption of all data protected by the expired DEK.
+This is by design — it ensures that expired DEKs are not kept indefinitely — but it can be expensive if it happens frequently.
+
+==== Log Retention Rationale
+
```
timofey-barmin, line 264: IMO this paragraph is confusing. I think we should say something like the following:
Each time when a new DEK is generated, log files get rotated (we start writing to a new log file).
Since the number of rotated log files that are kept on disk is limited, frequent DEK rotation may lead to limited log history.
This is also true for on-demand rotation (drop DEKs procedure): when one initiates the DEKs drop, we will rotated logs, which can lead to limit log history if called frequently.

modules/learn/pages/security/key-health-monitoring-alerts.adoc
```diff
@@ -0,0 +1,411 @@
...
+
+----
+
+[#key-test-alerts]
+=== Key Test Alerts
+
+Couchbase Server generates alerts when encryption operations fail.
+These alerts appear in the Web Console and can be configured to send notifications via email or other alert destinations.
+
+The following encryption-at-rest alerts are available:
```
timofey-barmin, line 112: All the elements below describe metrics that trigger one alert actually: "encryption-at-rest error"

If any of the described metrics increase, the "encryption-at-rest error" is triggered.

There is another encr-at-rest alert though: "encryption-at-rest key test failed"
The system periodically tests configured encryption tests. If such test fails, the "encryption-at-rest key test failed" alert gets fired.

modules/learn/pages/security/master-password-encryption-at-rest.adoc
```diff
@@ -0,0 +1,137 @@
+[#master-password-and-encryption]
+== Master Password and Encryption at Rest
+:description: The master password is the foundation of the encryption key hierarchy on each Couchbase Server node. It protects locally stored encryption keys and configuration data.
```
timofey-barmin, line 2: This paragraph makes an impression that the Master Password is something that must be configured for the encr-at-rest. That's not correct.
The Master Password can be used, but the customer doesn't have to use it. They can configure the system to not use the master password at all, and rely on the external KMS instead.

Default configuration:
1. The master password is ""
2. Configuration data is set to be encrypted with master password
3. Encryption for everything else is disabled

This configuration is obviously not secure. To make it secure we have two options:
- Configure master password on each node
- Change encryption method for "config data" to use "encryption key" which needs to be created separately - in this case they don't have to configure the master password

modules/learn/pages/security/master-password-encryption-at-rest.adoc
```diff
@@ -0,0 +1,137 @@
...
+[source,console]
+----
+curl -X POST -u <username>:<password> \
+  http://<node>:8091/node/controller/changeMasterPassword \
+  -d newPassword=<new-master-password>
+----
+
+==== Using an Environment Variable
+
+On all supported operating systems (except macOS), you can set the master password using the
```
timofey-barmin: This sounds like you can configure the password with the env var. This is not correct.
That env var can be used to provide (or pass) the password to a node when it starts.
Basically it is another way to do `couchbase-cli master-password --send-password`

modules/learn/pages/security/per-node-per-bucket.adoc
```diff
@@ -0,0 +1,87 @@
...
+Each node in the cluster generates and manages its own Data Encryption Keys (DEKs).
+DEKs are not shared between nodes.
+This design means:
+
+* Each node independently encrypts and decrypts the data it stores on its local disk.
+* If a node is compromised, only the DEKs for that node are exposed — DEKs on other nodes remain
+  unaffected.
+* When you add a new node to the cluster, Couchbase Server generates new DEKs for that node during
+  the rebalance process.
+  During file-based rebalance, the Index service preserves encryption keys as it transfers data
```
timofey-barmin: Bucket DEKs can be transferred between nodes during file based rebalance (not only by index service, KV also does it).
In this case, such transferred DEKs can be used for decryption purposes only.
DEKs are always moved encrypted.

modules/learn/pages/security/transparent-encryption-decryption.adoc
```diff
@@ -0,0 +1,66 @@
...
+
+. Couchbase Server obtains the master password (from the environment variable, a prompt, or a script).
+. It uses the master password to decrypt locally stored KEKs.
+  If a KEK is managed by an external KMS, Couchbase Server sends a decrypt request to the KMS.
+  This is the only time Couchbase Server contacts the external KMS during normal operation.
+. It uses the decrypted KEKs to decrypt the DEKs stored on disk.
+. It uses the decrypted DEKs to decrypt data as services read it from disk.
+
+After startup, DEKs and KEKs are cached in memory.
+Couchbase Server does not contact the external KMS again until the next key rotation or node restart.
```
timofey-barmin: It contacts external KMS periodically to test encryption keys (this happens automatically)

modules/manage/pages/manage-security/manually-drop-DEKs-reencrypt-data.adoc
```diff
@@ -0,0 +1,74 @@
...
+[#drop-deks]
+== Manually Drop DEKs and Re-encrypt Data
+
+You can have Couchbase Server drop the data encryption keys (DEKs) for a bucket or type of data and re-encrypt the data with new DEKs.
+When you drop the DEKs for a bucket or type of data, Couchbase Server creates a new replacement DEK and uses it to re-encrypt all of the data in the bucket or of that type.
+Then it removes the old DEKs.
+
+NOTE: Dropping DEKs is not the same as rotating DEKs. 
+When Couchbase Server automatically rotates a DEK, it makes it inactive, creates a new active DEK, and then uses that for newly written data.
+It does not re-encrypt data, nor does it automatically remove the old DEKs.
```
timofey-barmin: This point can be confusing.
In our case DEK rotation consists of two parallel processes:
1. Periodic generation of new DEKs - this process doesn't lead to any re-encryption
2. Automatic drop of old "expired" DEKs - this process can lead to data re-encryption
 
This means DEK rotation can re-encrypt data if there is still data that is encrypted with the expired DEKs

modules/learn/pages/security/encryption-key-management-services.adoc
boparai11: What @timofey-barmin says is correct, for KMIP the user can choose to do encryption/decryption via key GET and in this case we get the key from external server and encrypt/decrypt the data on the Couchbase Server side with that key. 

KMIP also supports the standard encrypt/decrypt mode, so has the option to also encrypt/decrypt externally without getting the key.

modules/learn/pages/security/encryption-at-rest-keys.adoc
supritha-kumar: Removed it

modules/learn/pages/security/encryption-key-management-services.adoc
supritha-kumar: Thank you. Have resolved this.

modules/learn/pages/security/encryption-key-rotation-expiration.adoc
supritha-kumar: Have reworded this to:

A new log file is created each time a new Data Encryption Key (DEK) is generated, triggering log file rotation. Because the system retains only a limited number of rotated log files on disk, frequent DEK rotation can result in reduced log history.

This also applies to on-demand rotation using the drop DEKs procedure. Initiating a DEK drop triggers log rotation, and if performed frequently, it can similarly reduce the available log history.

modules/learn/pages/security/encryption-key-rotation-expiration.adoc
supritha-kumar: addressed and rewritten accordingly.

modules/learn/pages/security/encryption-key-rotation-expiration.adoc
supritha-kumar: | Compliance requirement for periodic re-encryption
| Configured DEK rotation (recommended)
| Automatically re-encrypts data over time based on DEK lifetime configuration.
Only re-encrypts data that has not already been re-encrypted naturally.
If DEK rotation is not available, Drop DEKs can be used as an alternative, though it re-encrypts 100% of data on every invocation.

modules/learn/pages/security/key-health-monitoring-alerts.adoc
supritha-kumar: |===

| Alert | Metric | Description

| Encryption-at-Rest Error
| `cm_encr_at_rest_generate_dek_failures_total`
| Triggered when Couchbase Server fails to generate a new DEK.
If another DEK is present, the system continues using the old DEK (possibly expired) for encryption.
If no DEK exists (first enablement), data stays unencrypted until the issue is resolved.

| Encryption-at-Rest Error
| `cm_encryption_key_rotation_failures_total`
| Triggered when the system fails to rotate a configured encryption key (KEK).
Failed rotation does not cause immediate service disruption, but can be an early sign of an external KMS connectivity or configuration problem.

| Encryption-at-Rest Error
| `cm_encryption_service_failures_total`
| Covers a range of encryption key management failures, including problems reading or writing keys to disk, encryption and decryption failures, corrupted key files, and key integrity check failures.
Impact depends on which key is affected — can range from partial data unavailability to complete node failure.

| Encryption-at-Rest Key Test Failed
|`encr_at_rest_key_test_failed`
| Triggered when a periodic encryption key test fails. The system regularly tests configured encryption keys, and this alert fires if any such test does not pass.
|===

modules/learn/pages/security/master-password-encryption-at-rest.adoc
supritha-kumar: incorporated changes

modules/learn/pages/security/master-password-encryption-at-rest.adoc
supritha-kumar: Incorporated

modules/learn/pages/security/per-node-per-bucket.adoc
supritha-kumar: Incorporated

modules/learn/pages/security/transparent-encryption-decryption.adoc
supritha-kumar: Couchbase Server contacts the external KMS periodically to test configured encryption keys.

modules/manage/pages/manage-security/manually-drop-DEKs-reencrypt-data.adoc
supritha-kumar: NOTE: Dropping DEKs is not the same as DEK rotation. DEK rotation consists of two parallel processes:

Periodic generation of new DEKs — A new active DEK is created and used for newly written data. This process alone does not trigger re-encryption.

Automatic removal of expired DEKs — When a DEK reaches the end of its configured lifetime, the system drops it. If data still exists that is encrypted with the expired DEK, that data is re-encrypted with a current DEK before the expired key is removed.

This means DEK rotation can result in data re-encryption, but only when data encrypted with an expired DEK has not already been re-encrypted naturally (e.g., through compaction or other write activity).

modules/learn/pages/security/encryption-at-rest-keys.adoc
supritha-kumar: Removed the statement as this has now moved to Ponyo

modules/learn/pages/security/encryption-at-rest-keys.adoc
supritha-kumar: This introduces the next set of info, on encrypting configuration data, audit data and logs data.
Am going to leave this here for now.
Will keep this in mind when i look to restructure this.

modules/learn/pages/security/encryption-at-rest-keys.adoc
supritha-kumar: As mentioned above, this has moved to Ponyo and leaving this as is.

modules/learn/pages/security/when-CBServer-encrypts-data.adoc
```diff
@@ -0,0 +1,141 @@
...
+For example, if you enable encryption at rest for logs, Couchbase Server encrypts new log entries but does not encrypt existing log entries.
+Existing unencrypted data in buckets is encrypted when it's mutated or compacted.
+Now, this also applies to service-generated data (the "other" data type).
+
+When you enable "other" encryption, services encrypt new service-generated files (such as Query spill files and Index backfill files) but do not retroactively encrypt files written before enablement.
+
+You can use the REST API to force Couchbase Server to encrypt existing unencrypted data. 
+See xref:rest-api:security/encryption-at-rest/force-encryption-at-rest.adoc[] for more information.
+
+See xref:manage:manage-security/manage-encryption-at-rest.adoc#force-encryption[Force Encryption of Unencrypted Data] for more information.
```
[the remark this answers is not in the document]
supritha-kumar: removed this - duplication

modules/learn/pages/security/encryption-key-management-services.adoc
supritha-kumar: point! deleted it - thank you for the catch

modules/learn/pages/security/encryption-key-management-services.adoc
supritha-kumar: Line 51 - we've provided a link to AWS Key Managemnt Systemt. CLosing this out.

modules/learn/pages/security/encryption-key-management-services.adoc
supritha-kumar: Yes, this table lists the parameters that are passed to Couchbase Server when creating the encryption-at-rest key.

modules/learn/pages/security/encryption-key-management-services.adoc
supritha-kumar: The GCP Considerations section already exists right below, and i've changed the heading level.
On the "no link to steps" point, the === Prerequisites section right above this one already has the GCP-side steps.

I've followed the same model/structure throughout the page. If it helps, i'll rename this to "Configuration Parameters".

modules/learn/pages/security/encryption-key-management-services.adoc
supritha-kumar: This is just one section - hence, leaving it here.

PR #4129, opened 2026-06-15
DOC-14389--confusion-in-cbbackupmgr-compatibility-guide

RayOffiah:
Fixing the compatibilty table and cleaniing up the text.

https://preview.docs-test.couchbase.com/docs-server-DOC-14389--confusion-in-cbbackupmgr-compatibility-guide/server/current/backup-restore/enterprise-backup-restore.html

### DIFF
--- a/modules/backup-restore/pages/enterprise-backup-restore.adoc
+++ b/modules/backup-restore/pages/enterprise-backup-restore.adoc
@@ -31,7 +31,7 @@ Administrators are therefore recommended to define plans for both https://en.wik
 
 All backup is stored in and recovered from a [.term]#Backup Repository#.
 In turn, a [.term]#Backup Repository# is stored in a Backup Archive on the filesystem.
-Each backup job in the [.term]#Backup Repository# stores its backup in two ways:
+Each backup job in the [.term]#Backup Repository# stores its backup in 2 ways:
 
 * All bucket data is stored in a small, secondary database.
 * All bucket creation scripts and configuration files are stored on the filesystem, as files.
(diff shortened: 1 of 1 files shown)


osfameron (commented):
I don't like this slanted text:
Image

I feel like this should just be a column header for "Couchbase Server Version".

That said, if the *Server docs team* wants it that way, then sure, the inline CSS is fine for now, and I can move it into the site CSS if it's a patter we want to use.
I'll add @supritha-kumar as reviewer to give comment from Server docs perspective.

supritha-kumar:
> I don't like this slanted text: Image
> 
> I feel like this should just be a column header for "Couchbase Server Version".
> 
> That said, if the _Server docs team_ wants it that way, then sure, the inline CSS is fine for now, and I can move it into the site CSS if it's a patter we want to use. I'll add @supritha-kumar as reviewer to give comment from Server docs perspective.


It's supposed to be a column header - some change has caused it to look slanting and needs to be looked at.
i feel like the [cols="2,9*"] may be causing the issue - as opposed to the earlier [cols="5,3,3,3,3,3,3,3,3"].

On a separate note, we only support 8.0, 7.6 and 7.2 - do we still keep the older versions?

RayOffiah:
> It's supposed to be a column header - some change has caused it to look slanting and needs to be looked at.
> i feel like the [cols="2,9*"] may be causing the issue - as opposed to the earlier [cols="5,3,3,3,3,3,3,3,3"].

No, that was just me trying it out. In the end, it just looked too weird so I've removed the slant.

@matthallcb would like the old versions left on the page. I guess people might still be using old versions, even if we don't support them.

I've asked about including 8.0 on the page too.

matthallcb (approved):
Thanks Ray, looks a lot better!

